PR #67, opened 2021-11-29
feat: add support for plugins

iromli:
The changesets add support for plugins (focusing on `admin-ui`).


docker-jans-persistence-loader/scripts/utils.py
```diff
@@ -334,8 +346,13 @@ def get_injected_urls():
             for url in urls
         }
 
+    approved_issuer = [ctx["hostname"]]
+    token_server_hostname = os.environ.get("CN_TOKEN_SERVER_HOSTNAME")
```
moabu: This env is a bit obscured. Here it is defaulting to an empty string while in configAPI it is defaulting to the `hostname` which may be mis leading. Here it is assumed that the env means an additional token server ( auth server setup). While in configAPI it merely means the token server hostname. So one to one in configAPI and one to many here. In an actual setup this env will be set for both which here would result in a duplicate hostname in the likely event that the hostname is as the env. 

Maybe we should change the name to `CN_TOKEN_SERVER_BASE_HOSTNAME` while we are at it.

moabu (changes requested):

PR #162, opened 2021-12-23
fix: path ignore logic

ossdhaval:
This change is required to reinstate Sonar code quality checks which have stopped being executed as of now across multiple repos. For example, `jans-core` got impacted when I merged JanssenProject/jans-core#43. There were similar PRs for 3 or 4 other repos. Not sure why all the repos did not get similar PRs.

Previous code which uses `paths` construct which caused the issue. Instead, `paths-ignore` should be used without negative match patterns.

New code has been tested on `jans-core` repo and seems to be working fine.


moabu:
> Previous code which uses paths construct which caused the issue. Instead, paths-ignore should be used without negative match patterns.


I am approving it but we do need to use both constructs. For example we cannot use branches and branches ignore , its one or the other.

PR #388, opened 2022-01-14
fix: ce setup monorepo tweaks

mbaser:


yuriyz (approved):

yuriyz (approved):

smansoft (approved):
I've checked locally this implementing of the jans-ce-setup. It works OK.

PR #683, opened 2022-01-24
docs: restore microk8s steps

ossdhaval:


moabu:
This section should not be added. We will add a quickstart script that will create the default installation instead

PR #694, opened 2022-01-25
docs: remove sonar badges table from readme

ossdhaval:
As discussed with Mike, we will bring these stats back when they are ready.


moabu (approved):
Marking this PR to be reverted once ready.

PR #719, opened 2022-01-27
fix: linux-setup apache config file name

mbaser:


yuriyz (approved):

yuriyz:
@mbaser I think we had issues due to this bug and @smansoft fixed it on jenkins. Now with this fix I guess Sergey has to correct again is it correct ? 
@smansoft please check merged PR.

Issue #344, opened 2021-02-20
Allow return of custom authz endpoint request parameters to RP in response


martynaslawinska:
_yurem commented on Nov 15, 2018_
It's related to [Support Custom Params #647](https://github.com/GluuFederation/oxAuth/issues/647)
We need to add `boolean` option near to each custom parameter name in JSON configuration. When this option is `true` oxAuth should return custom parameter back to RP.

Today we fixed collision between Gluu IDP and passport-saml. Both of them send in state jwt with additional custom parameters. We added workaround. But this new option will help to resolve this issue.

_yurem commented on Dec 13, 2018_
Also we return AuthZ custom parameters as headers which leads to cleaning them up on redirect. I think we need to return them in response:
[oxAuth/Client/src/test/java/org/xdi/oxauth/ws/rs/AuthorizationResponseCustomHeaderTest.java](https://github.com/GluuFederation/oxAuth/blob/b33ad60927f21dad9648f453c45d82341bbc3350/Client/src/test/java/org/xdi/oxauth/ws/rs/AuthorizationResponseCustomHeaderTest.java#L81)
Line 81 in [b33ad60](https://github.com/GluuFederation/oxAuth/commit/b33ad60927f21dad9648f453c45d82341bbc3350)
`81     assertTrue(authorizationResponse.getHeaders().containsKey("CustomHeader1")); `

qbert2k:
A new boolean param was added to config. In the list of allowed custom parameter we can set returnInResponse to true to return it back to the RP. If it is set to false it will not be returned but will still be stored in session attributes.

![image](https://user-images.githubusercontent.com/2474616/152263114-dcb14b63-67f7-4908-b70e-1ed21753d94e.png)

PR #756, opened 2022-02-03
feat(jans-auth-server): allow return custom authz params to rp in response

qbert2k:


yuriyz (approved):

qbert2k:
https://github.com/JanssenProject/jans/issues/344

PR #757, opened 2022-02-03
fix: for JARM issue 310 311 and 314

HemantKMehta:
Changed the use of JWT to JWE for parsing encrypted request


jans-auth-server/server/src/main/java/io/jans/as/server/authorize/ws/rs/AuthorizeRestWebServiceImpl.java
```diff
@@ -782,16 +789,34 @@ private Response requestAuthorization(
     }
 
     @Nullable
-    private ResponseMode extractResponseMode(String request) {
-        final Jwt jwt = Jwt.parseSilently(request);
-        if (jwt == null) {
+    private Jwe extractJwe(String request, Client client) {
```
yuriyz: This is not going to work, we can not flip it from Jwt to Jwe and then from Jwe to Jwt depending on use case. It will never ends. Instead we should try both. For example: 
```
private JsonWebResponse extractResponseMode(String request) {
     String[] parts = request.split("\\.");
     if (parts.length == 5) {
         // parse jwe
         ...
         return jwe;
     }
     
     // parse jwt
     ...
     return jwt;
}
```

yuriyz (changes requested):

jans-auth-server/server/src/main/java/io/jans/as/server/authorize/ws/rs/AuthorizeRestWebServiceImpl.java
HemantKMehta: Not sure, as in case of JARM we only get encrypted request so should we include the other case too?

yuriyz: @HemantKMehta  Why do you think so?
In spec I see this:
`The JWT is either signed, or signed and encrypted. If the JWT is both signed and encrypted, the JSON document will be signed then encrypted, with the result being a Nested JWT, as defined in [RFC7519].`

HemantKMehta: but it may return null for other cases so when using it I need to add as condition to check for null value 
                Jwe jwe = extractJwe(request, client);
                 if (jwe !=null)  {
                     ... 
                     ...
                 }

HemantKMehta: Oh.. Yes its both signed and encrypted. 
I am changing it as you suggested, Thanks

yuriyz: > but it may return null for other cases so when using it I need to add as condition to check for null value

Did you really take a look to example I posted above? How can you get null if before you checked parts?

HemantKMehta: > > but it may return null for other cases so when using it I need to add as condition to check for null value
> 
> Did you really take a look to example I posted above? How can you get null if before you checked parts?
Actually this message was for my code with extractJwe method not for the example you have given. 
I am updating it as you suggested.

jans-auth-server/server/src/main/java/io/jans/as/server/authorize/ws/rs/AuthorizeRestWebServiceImpl.java
```diff
@@ -782,16 +790,37 @@ private Response requestAuthorization(
     }
 
     @Nullable
-    private ResponseMode extractResponseMode(String request) {
-        final Jwt jwt = Jwt.parseSilently(request);
-        if (jwt == null) {
+    private JsonWebResponse parseRequestToJweJwt(String request) {
```
yuriyz: parseRequestToJwr

jans-auth-server/server/src/main/java/io/jans/as/server/authorize/ws/rs/AuthorizeRestWebServiceImpl.java
```diff
@@ -782,16 +790,37 @@ private Response requestAuthorization(
...
+        String[] parts = request.split("\\.");
```
yuriyz, line 801: Before we had `Jwt.parseSilently(request);` which was nullity safe. Now without null check we will get NPE here for all not `request` test cases. Did you run tests against these changes?

jans-auth-server/server/src/main/java/io/jans/as/server/authorize/ws/rs/AuthorizeRestWebServiceImpl.java
```diff
@@ -782,16 +790,37 @@ private Response requestAuthorization(
...
+                }
+                Jwe jwe = Jwe.parse(request, privateKey, null);
+                if (jwe == null) {
+                    return null;
+                }
+                return jwe;
+            }
+            final Jwt jwt = Jwt.parseSilently(request);
+            if (jwt == null) {
+                return null;
```
yuriyz: You can remove this if and simply do `return Jwt.parseSilently(request);`, it is redundant, because if it's null we anyway return it.

jans-auth-server/server/src/main/java/io/jans/as/server/authorize/ws/rs/AuthorizeRestWebServiceImpl.java
```diff
@@ -782,16 +790,37 @@ private Response requestAuthorization(
...
+        } catch (Exception e) {
+            log.error(e.getMessage(), e);
             return null;
         }
-        JwtClaims jwtClaims = jwt.getClaims();
-        return ResponseMode.getByValue(jwtClaims.getClaimAsString("response_mode"));
     }
 
-    private void fillRedirectUriResponseforJARM(RedirectUriResponse redirectUriResponse, Jwt jwt, Client client) {
+    private void fillRedirectUriResponseforJARM(RedirectUriResponse redirectUriResponse, JsonWebResponse jwt, Client client) {
```
yuriyz: jwt -> jwr

yuriyz (changes requested):

jans-auth-server/server/src/main/java/io/jans/as/server/authorize/ws/rs/AuthorizeRestWebServiceImpl.java
HemantKMehta: I executed only the JARM tests but not with any tests where request is null. Should I check if (request!=null) before split.

jans-auth-server/server/src/main/java/io/jans/as/server/authorize/ws/rs/AuthorizeRestWebServiceImpl.java
```diff
@@ -784,19 +793,39 @@ private Response requestAuthorization(
         return builder.build();
     }
 
-    private ResponseMode extractResponseMode(String request) {
-        final Jwt jwt = Jwt.parseSilently(request);
-        if (jwt == null) {
-            return null;
+    @Nullable
+    private JsonWebResponse parseRequestToJwr(String request) {
+        if (request != null) {
```
yuriyz: Lets use early exit. Means
```
if (request == null) {
    return null;
}
```
We should avoid nested ifs when possible. You can see already if-hell in other places.

yuriyz:
Looks good now but we have to postpone it until all JARM tests are passed on jenkins.

yuriyz (approved):

PR #765, opened 2022-02-04
fix: update scripts

moabu:
Updates some  external scripts


iromli (commented):
@moabu There are changes in `*.py` files that need to be double-checked:

1. The `from ...` syntax without `... import ...` will raise issue
2. The blank `print` statement and trailing string below the blank `print` may be suppressed from stdout

Are you using some auto-formatting tool to modify `*.py` files?

iromli (approved):

PR #767, opened 2022-02-04
fix: fix RedirectUri sonarcloud refactor code-coverage unit-tests

jmunozherbas:
1.- Fixed, all sonar issues
2.- inproved, code coverage percent with unit test ( methods with logic)

### DIFF
--- a/jans-auth-server/common/pom.xml
+++ b/jans-auth-server/common/pom.xml
@@ -219,5 +219,15 @@
 			<groupId>org.testng</groupId>
 			<artifactId>testng</artifactId>
 		</dependency>
-	</dependencies>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-testng</artifactId>
+            <version>0.4.8</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
 </project>
--- a/jans-auth-server/common/pom.xml
+++ b/jans-auth-server/common/pom.xml
@@ -226,8 +226,6 @@
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-testng</artifactId>
-            <version>0.4.8</version>
-            <scope>test</scope>
         </dependency>
     </dependencies>
 </project>
--- a/jans-auth-server/common/src/main/java/io/jans/as/common/util/RedirectUri.java
+++ b/jans-auth-server/common/src/main/java/io/jans/as/common/util/RedirectUri.java
@@ -12,7 +12,9 @@ import io.jans.as.model.crypto.AbstractCryptoProvider;
 import io.jans.as.model.crypto.encryption.BlockEncryptionAlgorithm;
 import io.jans.as.model.crypto.encryption.KeyEncryptionAlgorithm;
 import io.jans.as.model.crypto.signature.SignatureAlgorithm;
+import io.jans.as.model.exception.CryptoProviderException;
 import io.jans.as.model.exception.InvalidJweException;
+import io.jans.as.model.exception.InvalidJwtException;
 import io.jans.as.model.jwe.Jwe;
 import io.jans.as.model.jwe.JweEncrypter;
 import io.jans.as.model.jwe.JweEncrypterImpl;
@@ -67,12 +69,14 @@ public class RedirectUri {
     private String nestedSharedSecret;
     private JSONObject nestedJsonWebKeys;
     private AbstractCryptoProvider cryptoProvider;
+    private Jwe jwtForEncrypt;
 
     public RedirectUri(String baseRedirectUri) {
         this.baseRedirectUri = baseRedirectUri;
         this.responseMode = ResponseMode.QUERY;
 
-        responseParameters = new HashMap<String, String>();
+        this.responseParameters = new HashMap<String, String>();
+        this.jwtForEncrypt = new Jwe();
     }
 
     public RedirectUri(String baseRedirectUri, List<ResponseType> responseTypes, ResponseMode responseMode) {
@@ -103,9 +107,21 @@ public class RedirectUri {
         }
     }
 
+    public void setJwtForEncrypt(Jwe jwtForEncrypt) {
+        this.jwtForEncrypt = jwtForEncrypt;
+    }
+
     @Nullable
     public String getResponseParameter(@NotNull String key) {
         return responseParameters.get(key);
+    }
+
+    public int getResponseParamentersSize() {
+        int size = 0;
+        if (responseParameters != null) {
+            size = responseParameters.size();
+        }
+        return size;
     }
 
     public String getIssuer() {
@@ -267,8 +283,6 @@ public class RedirectUri {
                     }
                 }
             }
-        } catch (UnsupportedEncodingException e) {
-            e.printStackTrace();
         } catch (Exception e) {
             e.printStackTrace();
         }
@@ -276,7 +290,7 @@ public class RedirectUri {
         return sb.toString();
     }
 
-    private String getJarmResponse() throws Exception {
+    private String getJarmResponse() throws InvalidJweException, InvalidJwtException, CryptoProviderException {
         if (keyEncryptionAlgorithm != null && blockEncryptionAlgorithm != null) {
             if (signatureAlgorithm != null) {
                 String jws = getJwsResponse(true);
@@ -293,7 +307,7 @@ public class RedirectUri {
         }
     }
 
-    private String getJwsResponse(boolean nested) throws Exception {
+    private String getJwsResponse(boolean nested) throws InvalidJwtException, CryptoProviderException {
         Jwt jwt = new Jwt();
 
         // Header
@@ -325,11 +339,10 @@ public class RedirectUri {
         // Signature
         String signature = cryptoProvider.sign(jwt.getSigningInput(), jwt.getHeader().getKeyId(), nested ? nestedSharedSecret : sharedSecret, signatureAlgorithm);
         jwt.setEncodedSignature(signature);
-
         return jwt.toString();
     }
 
-    private String getJweResponse(String nestedJws) throws Exception {
+    private String getJweResponse(String nestedJws) throws InvalidJweException, InvalidJwtException, CryptoProviderException {
         Jwe jwe = new Jwe();
 
         // Header
@@ -356,25 +369,20 @@ public class RedirectUri {
             jwe.setSignedJWTPayload(jwt);
         }
 
-        // Encryption
+        //Encryption
         if (keyEncryptionAlgorithm == KeyEncryptionAlgorithm.RSA_OAEP
                 || keyEncryptionAlgorithm == KeyEncryptionAlgorithm.RSA1_5) {
             PublicKey publicKey = cryptoProvider.getPublicKey(keyId, jsonWebKeys, null);
-
-            if (publicKey != null) {
-                JweEncrypter jweEncrypter = new JweEncrypterImpl(keyEncryptionAlgorithm, blockEncryptionAlgorithm, publicKey);
-                jwe = jweEncrypter.encrypt(jwe);
-            } else {
+            if (publicKey == null) {
                 throw new InvalidJweException("The public key is not valid");
             }
+
+            JweEncrypter jweEncrypter = new JweEncrypterImpl(keyEncryptionAlgorithm, blockEncryptionAlgorithm, publicKey);
+            jwe = jwtForEncrypt.encryptJwe(jwe, jweEncrypter);
         } else if (keyEncryptionAlgorithm == KeyEncryptionAlgorithm.A128KW
                 || keyEncryptionAlgorithm == KeyEncryptionAlgorithm.A256KW) {
-            try {
-                JweEncrypter jweEncrypter = new JweEncrypterImpl(keyEncryptionAlgorithm, blockEncryptionAlgorithm, sharedSymmetricKey);
-                jwe = jweEncrypter.encrypt(jwe);
-            } catch (Exception e) {
-                throw new InvalidJweException(e);
-            }
+            JweEncrypter jweEncrypter = new JweEncrypterImpl(keyEncryptionAlgorithm, blockEncryptionAlgorithm, sharedSymmetricKey);
+            jwe = jwtForEncrypt.encryptJwe(jwe, jweEncrypter);
         }
 
         return jwe.toString();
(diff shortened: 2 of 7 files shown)


jans-auth-server/common/src/main/java/io/jans/as/common/util/RedirectUri.java
```diff
@@ -103,11 +107,23 @@ public void addResponseParameter(String key, String value) {
...
     public String getResponseParameter(@NotNull String key) {
         return responseParameters.get(key);
     }
 
+    public int getResponseParamentersSize() {
+        int size = 0;
+        if (responseParameters != null) {
+            size = responseParameters.size();
+        }
+        return size;
```
Milton-Ch:
```suggestion
        return responseParameters != null ? responseParameters.size() : 0;
```

jans-auth-server/common/src/test/java/io/jans/as/common/RedirectUriTest.java
```diff
@@ -0,0 +1,342 @@
...
+
+        String queryResult = redirectUri.getQueryString();
+
+        assertNoEmptyQueryString(queryResult, RESPONSE, 1);
+
+    }
+
+    @Test()
+    public void toString_withResponseModeFormPostJwtAndNoParams_NoEmpty() {
+
```
Milton-Ch: Remove empty line?

jans-auth-server/common/src/test/java/io/jans/as/common/RedirectUriTest.java
```diff
@@ -0,0 +1,342 @@
...
+        RedirectUri redirectUri = getRedirectUriTemplateGetQueryString(null, null, null, null);
+        redirectUri.addResponseParameter(AuthorizeResponseParam.CODE, "123");
+        redirectUri.addResponseParameter(AuthorizeResponseParam.EXPIRES_IN, "3000");
+        redirectUri.addResponseParameter(AuthorizeResponseParam.ACCESS_TOKEN, "tk123");
+        String queryResult = redirectUri.getQueryString();
+
+        assertNoEmptyQueryString(queryResult, AuthorizeResponseParam.CODE, 3);
+    }
+
+    @Test()
```
Milton-Ch: You can remove parentheses, they are not required here.

jans-auth-server/common/pom.xml
```diff
@@ -219,5 +219,15 @@
...
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-testng</artifactId>
+            <version>0.4.8</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
```
Milton-Ch: Worths checking if we can include it as part of a lower level, perhaps bom or similar level.

jans-auth-server/common/src/test/java/io/jans/as/common/RedirectUriTest.java
```diff
@@ -0,0 +1,342 @@
...
+        RedirectUri redirectUri = getRedirectUriTemplateGetQueryString(null, null, null, null);
+        redirectUri.parseQueryString(null);
+        assertEquals(redirectUri.getResponseParamentersSize(), 0);
+    }
+
+    @Test
+    public void parseQueryString_oneParam_size1() throws CryptoProviderException {
+        RedirectUri redirectUri = getRedirectUriTemplateGetQueryString(null, null, null, null);
+        redirectUri.parseQueryString(AuthorizeResponseParam.EXPIRES_IN + "=");
+        assertEquals(redirectUri.getResponseParamentersSize(), 1);
```
Milton-Ch, line 72: I'd suggest to add more asserts here in order to validate right params were gotten.

jans-auth-server/common/src/test/java/io/jans/as/common/RedirectUriTest.java
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    @Test
+    public void parseQueryString_threeParams_size3() throws CryptoProviderException {
+        RedirectUri redirectUri = getRedirectUriTemplateGetQueryString(null, null, null, null);
+        redirectUri.parseQueryString(AuthorizeResponseParam.CODE + "=Pcode&" + AuthorizeResponseParam.EXPIRES_IN + "=3000&" + AuthorizeResponseParam.ACCESS_TOKEN + "=tk123");
+        assertEquals(redirectUri.getResponseParamentersSize(), 3);
```
Milton-Ch, line 86: Similar to my previous comment, length doesn't define that the whole method was processed properly.
I'd suggest to extract params and compare them with the real params.

jans-auth-server/common/src/test/java/io/jans/as/common/RedirectUriTest.java
```diff
@@ -0,0 +1,342 @@
...
+
+    @Test()
+    public void getQueryString_responseModeFormPostJWT_NoEmpty() throws CryptoProviderException {
+        RedirectUri redirectUri = getRedirectUriTemplateGetQueryString(ResponseMode.FORM_POST_JWT, null, null, null);
+        AbstractCryptoProvider cryptoProvider = mock(AbstractCryptoProvider.class);
+        when(cryptoProvider.sign(anyString(), anyString(), anyString(), Mockito.any())).thenReturn("12345");
+        redirectUri.setCryptoProvider(cryptoProvider);
+        String queryResult = redirectUri.getQueryString();
+        //No empty Result
+        assertTrue(queryResult.length() > 0);
```
Milton-Ch, line 129: I'd suggest to add `verify` sentences at the end for CryptoProvider as well. Verify that certain methods were called with specific values.

jans-auth-server/common/src/test/java/io/jans/as/common/RedirectUriTest.java
```diff
@@ -0,0 +1,342 @@
...
+        when(cryptoProvider.getPublicKey(anyString(), Mockito.any(), Mockito.any())).thenReturn(null);
+        redirectUri.setCryptoProvider(cryptoProvider);
+
+        String queryResult = redirectUri.getQueryString();
+        assertEquals(queryResult, "");
+    }
+
+    @Test()
+    public void getQueryString_withEncriptionAlgorithmRSANoSignatureAlgorithm_responseEncoded() throws UnsupportedEncodingException, CryptoProviderException, InvalidJweException {
+        RedirectUri redirectUri = getRedirectUriTemplateGetQueryString(ResponseMode.JWT, KeyEncryptionAlgorithm.RSA1_5, BlockEncryptionAlgorithm.A256GCM, null);
```
Milton-Ch: Minor: try to keep line length limit.

jans-auth-server/common/src/test/java/io/jans/as/common/RedirectUriTest.java
```diff
@@ -0,0 +1,342 @@
...
+            public byte[] getEncoded() {
+                return new byte[0];
+            }
+
+            @Override
+            public BigInteger getModulus() {
+                return null;
+            }
+        };
+    }
```
Milton-Ch: I'm guessing that you can replace this by using `mock` of RSAPublicKey instead.

jans-auth-server/common/src/main/java/io/jans/as/common/util/RedirectUri.java
```diff
@@ -356,25 +369,20 @@ private String getJweResponse(String nestedJws) throws Exception {
...
-            if (publicKey != null) {
-                JweEncrypter jweEncrypter = new JweEncrypterImpl(keyEncryptionAlgorithm, blockEncryptionAlgorithm, publicKey);
-                jwe = jweEncrypter.encrypt(jwe);
-            } else {
+            if (publicKey == null) {
                 throw new InvalidJweException("The public key is not valid");
             }
+
+            JweEncrypter jweEncrypter = new JweEncrypterImpl(keyEncryptionAlgorithm, blockEncryptionAlgorithm, publicKey);
+            jwe = jwtForEncrypt.encryptJwe(jwe, jweEncrypter);
```
Milton-Ch: Why are you using jwtForEncrypt? do you really require to extract it outside? I mean, why cannot be jwtEncrypter?

jans-auth-server/common/pom.xml
yuriyz: I do agree, version should go to at least parent pom.

jans-auth-server/common/src/main/java/io/jans/as/common/util/RedirectUri.java
```diff
@@ -67,12 +69,14 @@
     private String nestedSharedSecret;
     private JSONObject nestedJsonWebKeys;
     private AbstractCryptoProvider cryptoProvider;
+    private Jwe jwtForEncrypt;
```
yuriyz, line 63: We have to strive to keep variables as narrow in scope as possible. This one went to class level while it has to be local to method without obvious reason. In short, lets drop it.

jans-auth-server/common/src/main/java/io/jans/as/common/util/RedirectUri.java
```diff
@@ -399,52 +407,57 @@ private void appendFragmentSymbol(StringBuilder sb) {
...
+                appendQuerySymbol(sb);
             }
+            sb.append(getQueryString());
+        } else {
+            appendDeafultToString(sb);
         }
         return sb.toString();
     }
+
+    private void appendDeafultToString(StringBuilder sb) {
```
yuriyz, line 455: Typo: Deafult -> Default

jans-auth-server/common/src/test/java/io/jans/as/common/RedirectUriTest.java
```diff
@@ -0,0 +1,342 @@
...
+
+    @Test()
+    public void toString_withResponseModeFormPostJwtAndNoParams_NoEmpty() {
+
+        RedirectUri redirectUri = getRedirectUriTemplateToString();
+        redirectUri.setResponseMode(ResponseMode.FORM_POST_JWT);
+        redirectUri.setCryptoProvider(cryptoProvider);
+
+        assertTrue(redirectUri.toString().length() > 0, "Is Empty");
```
yuriyz: toString() is important test. In each `toString()` test we should see equals assertion
```
assertEquals("expected value", redirectUri.toString());
```
Also when I look into that method after your refactoring I am not sure logic of the method is not broken. If you cover it with tests correctly (with equals) we will know it exactly. It is main goal of writing these tests. It is not to get good coverage number but really be sure that logic inside is exactly what we expect it to be.

I would suggest:
1. take old version of class
2. cover it with equals test
3. refactor it
4. Run tests from step 2. In this way you will be sure that refactoring does not broke anything.

jans-auth-server/model/src/main/java/io/jans/as/model/jwe/Jwe.java
```diff
@@ -49,6 +49,10 @@ public static Jwe parse(String encodedJwe, PrivateKey privateKey, byte[] sharedS
         return jwe;
     }
 
+    public Jwe encryptJwe(Jwe jwe, JweEncrypter jweEncrypter) throws InvalidJweException {
```
yuriyz, line 51: Lets drop this method. Not clear why do we need it here. If we need higher level of abstraction for encryption then lets add it. Calling jwe drom jweEncrypter or jweEncrypted from jwe does not solve anything.

jans-auth-server/common/pom.xml
jmunozherbas: I have removed the version from jans-common pom,xml, and put it on parent pom.xml.
I have included the dependency in jans-bom pom.xml too

jans-auth-server/common/src/main/java/io/jans/as/common/util/RedirectUri.java
jmunozherbas: The variable was deleted.

jans-auth-server/common/src/main/java/io/jans/as/common/util/RedirectUri.java
jmunozherbas: got it, refactored

jans-auth-server/common/src/main/java/io/jans/as/common/util/RedirectUri.java
jmunozherbas: jwtEncrypter was deleted from class.

jans-auth-server/common/src/main/java/io/jans/as/common/util/RedirectUri.java
jmunozherbas: Got it, thanks

jans-auth-server/common/src/test/java/io/jans/as/common/RedirectUriTest.java
jmunozherbas: More Asserts, was added

jans-auth-server/common/src/test/java/io/jans/as/common/RedirectUriTest.java
jmunozherbas: More Asserts, was added

jans-auth-server/common/src/test/java/io/jans/as/common/RedirectUriTest.java
jmunozherbas: Got it, thanks

jans-auth-server/common/src/test/java/io/jans/as/common/RedirectUriTest.java
jmunozherbas: Got it.

jans-auth-server/common/src/test/java/io/jans/as/common/RedirectUriTest.java
jmunozherbas: I have replaced it, with a method better resumed.

jans-auth-server/common/src/test/java/io/jans/as/common/RedirectUriTest.java
jmunozherbas: Got it, thanks.

jans-auth-server/common/src/test/java/io/jans/as/common/RedirectUriTest.java
jmunozherbas: I have refactored tests, including more asserts and assertEquals in toString tests.
according your suggestion, i have tested previous and last version of the method, the results was successfully.

PR #786, opened 2022-02-09
fix: check alg none to display error JARM issue310

HemantKMehta:
check the algorithm none and display none algorithm is not allowed in FAPI

### DIFF
--- a/jans-auth-server/server/src/main/java/io/jans/as/server/authorize/ws/rs/AuthorizeRestWebServiceImpl.java
+++ b/jans-auth-server/server/src/main/java/io/jans/as/server/authorize/ws/rs/AuthorizeRestWebServiceImpl.java
@@ -407,7 +407,7 @@ public class AuthorizeRestWebServiceImpl implements AuthorizeRestWebService {
                     JsonWebResponse jwr = parseRequestToJwr(request);
                     if (jwr != null) {
                         String checkForAlg = jwr.getClaims().getClaimAsString("alg"); // to handle Jans Issue#310
-                        if (checkForAlg.equals("none")) {
+                        if ("none".equals(checkForAlg)) {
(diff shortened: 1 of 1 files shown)


jans-auth-server/server/src/main/java/io/jans/as/server/authorize/ws/rs/AuthorizeRestWebServiceImpl.java
```diff
@@ -406,6 +406,14 @@ private Response requestAuthorization(
                 } catch (WebApplicationException e) {
                     JsonWebResponse jwr = parseRequestToJwr(request);
                     if (jwr != null) {
+                        String checkForAlg = jwr.getClaims().getClaimAsString("alg"); // to handle Jans Issue#310
+                        if (checkForAlg.equals("none")) {
```
yuriyz: `"none".equals(alg)` - more defensive, it should not be null but claim itself is nullable.

Will `jwr.getClaims().getClaimAsString("alg")` gets inner request object `alg`, it looks like outer `alg` ?

yuriyz (changes requested):

jans-auth-server/server/src/main/java/io/jans/as/server/authorize/ws/rs/AuthorizeRestWebServiceImpl.java
HemantKMehta: Thanks, I will change the condition as suggested.
During debugging, I checked and got the claims value as {alg:none} and this is why we don't get responseMode from claims.
Screenshot from 2022-02-08 18-51-41

HemantKMehta: Screenshot from 2022-02-09 15-33-37

yuriyz (approved):

Issue #777, opened 2022-02-08
Remove duplicate scripts in integrations folder


jgomer2001:
Move folders in `integrations` folder to `./jans-linux-setup/static/extension/person_authentication/other`. If a given script already has a correspondingo one in `person_authentication` it is removed (from `other`) and the README updated accordingly

jgomer2001:
Files with useless or extremely deprecated instructions were removed too.

jgomer2001:
https://github.com/JanssenProject/jans/pull/788/commits/587d8c129957f3bee463ce7cd6b1a1363768997d

PR #788, opened 2022-02-09
chore: refactor main 2fa scripts and move integrations folder #777

jgomer2001:
#777


moabu (approved):
based on a request from Jose I am approving. This is too big to review and things seem to be broken so it might be the right time. 

Overriding normal procedure with guilt 😞

Issue #804, opened 2022-02-11
tests(jans-auth-server): enhance integration tests - easier asserts and easier test build


yuriyz:
### Description

Within this ticket we should concentrate on 2 main goals for `jans-auth-server/client/src/test/java`: 
1. remove duplicated lines in tests. We should make standard asserters for common responses and use it instead of repeated blocks (as shown in comments below)
2. during test build we use repeated lines of register client, send authorization, assert response.
It would be great if you can think about builders for common steps and simplify process of building tests. It's ok if you simplify it with simple static methods (like asserters) but if you can come with builders (like spring and lombok works) it would look both better and more readable for developers. 

Lets make it with 2 PRs, it will be too much in one PR.

![image](https://user-images.githubusercontent.com/2013070/153573929-b87cae67-7f6a-4541-81dc-940e3c887a19.png)

![image](https://user-images.githubusercontent.com/2013070/153574005-c2caed68-1294-4955-abdf-281ee86c3698.png)

yuriyz:
While reducing asserts looks quite good maybe we should think about introducing assert builder?

```
Assert.builder().registerResponse(registerResponse).status(201).registrationUriIsNotNull().check();
```
Because with ` assertRegisterResponseOk(registerResponse, 201, true);` it's not clear what `201` or `true/false` means. We have to look inside method to understand it. With builder we will not have such issue.

Lets apply builder idea on server side tests. Applying it to client tests is up to you.

yuriyz:
Done, PR merged.

PR #828, opened 2022-02-16
test(jans-auth-server): refactor client/src/test/java Issue #804 

jmunozherbas:
Common assert code lines (in different classes) was replaced with next methods :

io.jans.as.client.client.Asserter.assertOk 
io.jans.as.client.client.Asserter.assertBadRequest
io.jans.as.client.client.Asserter.assertBadRequestStatus

https://github.com/JanssenProject/jans/issues/804

### DIFF
--- a/jans-auth-server/client/src/test/java/io/jans/as/client/ciba/BackchannelAuthenticationPingMode.java
+++ b/jans-auth-server/client/src/test/java/io/jans/as/client/ciba/BackchannelAuthenticationPingMode.java
@@ -22,7 +22,6 @@ import io.jans.as.client.TokenRequest;
 import io.jans.as.client.TokenResponse;
 import io.jans.as.client.UserInfoClient;
 import io.jans.as.client.UserInfoResponse;
-import io.jans.as.client.client.Asserter;
 import io.jans.as.model.ciba.BackchannelAuthenticationErrorResponseType;
 import io.jans.as.model.common.AuthenticationMethod;
 import io.jans.as.model.common.BackchannelTokenDeliveryMode;
@@ -55,6 +54,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
 
+import static io.jans.as.client.client.Asserter.assertOk;
 import static io.jans.as.model.register.RegisterRequestParam.BACKCHANNEL_AUTHENTICATION_REQUEST_SIGNING_ALG;
 import static io.jans.as.model.register.RegisterRequestParam.BACKCHANNEL_CLIENT_NOTIFICATION_ENDPOINT;
 import static io.jans.as.model.register.RegisterRequestParam.BACKCHANNEL_TOKEN_DELIVERY_MODE;
@@ -116,7 +116,7 @@ public class BackchannelAuthenticationPingMode extends BaseTest {
         RegisterResponse registerResponse = registerClient.exec();
 
         showClient(registerClient);
-        Asserter.assertOk(registerResponse);
+        assertOk(registerResponse);
 
         assertTrue(registerResponse.getClaims().containsKey(BACKCHANNEL_TOKEN_DELIVERY_MODE.toString()));
         assertTrue(registerResponse.getClaims().containsKey(BACKCHANNEL_AUTHENTICATION_REQUEST_SIGNING_ALG.toString()));
@@ -235,7 +235,7 @@ public class BackchannelAuthenticationPingMode extends BaseTest {
         RegisterResponse registerResponse = registerClient.exec();
 
         showClient(registerClient);
-        Asserter.assertOk(registerResponse);
+        assertOk(registerResponse);
 
         assertTrue(registerResponse.getClaims().containsKey(BACKCHANNEL_TOKEN_DELIVERY_MODE.toString()));
         assertTrue(registerResponse.getClaims().containsKey(BACKCHANNEL_AUTHENTICATION_REQUEST_SIGNING_ALG.toString()));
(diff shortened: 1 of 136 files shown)


jans-auth-server/client/src/test/java/io/jans/as/client/ciba/BackchannelAuthenticationPingMode.java
```diff
@@ -528,11 +513,7 @@ public void backchannelTokenDeliveryModePingIdTokenHintRS256(
         RegisterResponse registerResponse = registerClient.exec();
 
         showClient(registerClient);
-        assertEquals(registerResponse.getStatus(), 201, "Unexpected response code: " + registerResponse.getEntity());
-        assertNotNull(registerResponse.getClientId());
-        assertNotNull(registerResponse.getClientSecret());
-        assertNotNull(registerResponse.getRegistrationAccessToken());
-        assertNotNull(registerResponse.getClientSecretExpiresAt());
+Asserter.assertOk(registerResponse);
```
yuriyz: Not formatted (as well as in many other places).
Lets use static import to have just `assertOk(registerResponse);`.
Otherwise nice changes. It makes our tests much shorter.

jans-auth-server/client/src/test/java/io/jans/as/client/client/Asserter.java
```diff
@@ -51,15 +51,21 @@ public static void assertOk(RegisterResponse response) {
...
-        assertNotNull(registerResponse.getEntity());
-        assertNotNull(registerResponse.getErrorType());
-        assertNotNull(registerResponse.getErrorDescription());
+        assertEquals(registerResponse.getStatus(), 400, "Unexpected response code: " + registerResponse.getEntity());
+        assertNotNull(registerResponse.getEntity(), "The entity is null");
+        assertNotNull(registerResponse.getErrorType(), "The error type is null");
+        assertNotNull(registerResponse.getErrorDescription(), "The error description is null");
+    }
+
+    public static void assertBadRequestStatus(RegisterResponse registerResponse) {
```
yuriyz: We don't need this method, please remove. It does not differ from `public static void assertBadRequest(RegisterResponse registerResponse)`. If error and error description are not null then indeed entity can't be null as well because otherwise from where it took error and description.

jans-auth-server/client/src/test/java/io/jans/as/client/client/Asserter.java
```diff
@@ -51,15 +51,21 @@ public static void assertOk(RegisterResponse response) {
         assertEquals(response.getStatus(), 201, "Unexpected response code: " + response.getEntity());
         assertNotNull(response.getClientId());
         assertNotNull(response.getClientSecret());
-        assertNotNull(response.getClientIdIssuedAt());
         assertNotNull(response.getClientSecretExpiresAt());
+        assertNotNull(response.getRegistrationAccessToken());
```
yuriyz, line 55: I guess we can add also assert for `registrationClientUri` as well.

jans-auth-server/client/src/test/java/io/jans/as/client/client/Asserter.java
```diff
@@ -51,15 +51,21 @@ public static void assertOk(RegisterResponse response) {
         assertEquals(response.getStatus(), 201, "Unexpected response code: " + response.getEntity());
         assertNotNull(response.getClientId());
         assertNotNull(response.getClientSecret());
-        assertNotNull(response.getClientIdIssuedAt());
```
yuriyz, line 54: Why this assert is removed? I guess issued at is always present when client is created, no?

yuriyz (changes requested):
See inlined comments

jans-auth-server/client/src/test/java/io/jans/as/client/client/Asserter.java
jmunozherbas: assertNotNull(response.getClientIdIssuedAt());

was not in most of the asserts that I have replaced, I preferred do not include this assert to avoid problems in current test cases.
I preserved this assert in cases that it was present.
I don't know if this assert was omitted due to specific reasons, or it was forgotten.
who I have to ask to confirm that issuedAt is always present when client is created?

jans-auth-server/client/src/test/java/io/jans/as/client/client/Asserter.java
jmunozherbas: assertNotNull(response.getRegistrationClientUri());
is not used in most test cases, due to it I preferred do not include it in **assertOk**
I preserved this assert in cases that it was present.

jans-auth-server/client/src/test/java/io/jans/as/client/client/Asserter.java
yuriyz: You can ask me. Please add it.

jans-auth-server/client/src/test/java/io/jans/as/client/client/Asserter.java
yuriyz: I see, lets unify it. Add assert for registration access token.

jans-auth-server/client/src/test/java/io/jans/as/client/ciba/BackchannelAuthenticationPingMode.java
jmunozherbas: OK, it was fixed

jans-auth-server/client/src/test/java/io/jans/as/client/client/Asserter.java
jmunozherbas: method was removed.

jans-auth-server/client/src/test/java/io/jans/as/client/client/Asserter.java
jmunozherbas: next asserts was added:
assertNotNull(response.getRegistrationClientUri()); assertNotNull(response.getClientIdIssuedAt());

jans-auth-server/client/src/test/java/io/jans/as/client/client/Asserter.java
jmunozherbas: it was added.

yuriyz (approved):

PR #829, opened 2022-02-16
test: UserService code-coverage and sonar

jmunozherbas:
Refactor code (sonar)
Unit Tests (coverage)

### DIFF
--- a/jans-auth-server/common/src/test/java/io/jans/as/common/service/common/UserServiceImplTest.java
+++ b/jans-auth-server/common/src/test/java/io/jans/as/common/service/common/UserServiceImplTest.java
@@ -1,0 +1,30 @@
+package io.jans.as.common.service.common;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class UserServiceImplTest extends UserService {
+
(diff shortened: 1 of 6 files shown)


jans-auth-server/common/src/test/java/io/jans/as/common/service/common/UserServiceImplTest.java
```diff
@@ -0,0 +1,30 @@
+package io.jans.as.common.service.common;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class UserServiceImplTest extends UserService {
```
yuriyz: `TestUserService`, when we put suffix `*Test` it indicates classes with executable tests while it is not the case here.

yuriyz (changes requested):

jans-auth-server/common/src/test/java/io/jans/as/common/service/common/UserServiceImplTest.java
jmunozherbas: OK, it was fixed.

yuriyz (approved):

PR #879, opened 2022-02-23
test: jans-auth-client - reduce using assertNotNull, replacing with asserters #804

jmunozherbas:
new Asserters methods to cover test cases and replace common code assertions.
class: io.jans.as.client.client.Asserter

module: jans-auth-client 

https://github.com/JanssenProject/jans/issues/804

### DIFF
--- a/jans-auth-server/client/src/test/java/io/jans/as/client/client/Asserter.java
+++ b/jans-auth-server/client/src/test/java/io/jans/as/client/client/Asserter.java
@@ -6,28 +6,33 @@
 
 package io.jans.as.client.client;
 
-import io.jans.as.client.AuthorizationResponse;
-import io.jans.as.client.JwkClient;
-import io.jans.as.client.RegisterResponse;
-import io.jans.as.client.TokenResponse;
+import io.jans.as.client.*;
 import io.jans.as.client.par.ParResponse;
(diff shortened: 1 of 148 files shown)


jans-auth-server/client/src/test/java/io/jans/as/client/client/Asserter.java
```diff
@@ -47,31 +52,77 @@ public static void assertParResponse(ParResponse response) {
...
-        assertEquals(response.getStatus(), 201, "Unexpected response code: " + response.getEntity());
+    public static void assertRegisterResponseOk(RegisterResponse response, int status, boolean checkClientUri) {
+        assertEquals(response.getStatus(), status, "Unexpected response code: " + response.getEntity());
         assertNotNull(response.getClientId());
         assertNotNull(response.getClientSecret());
         assertNotNull(response.getClientSecretExpiresAt());
         assertNotNull(response.getRegistrationAccessToken());
         assertNotNull(response.getClientIdIssuedAt());
-        assertNotNull(response.getRegistrationClientUri());
+        if (checkClientUri) {
```
yuriyz, line 62: Why do we have this `if`? registration uri is always provided.

yuriyz (approved):

PR #880, opened 2022-02-23
fix: jans-linux-setup openbanking

mbaser:


HemantKMehta (approved):
Tested the installation with this update. It worked perfectly.

yuriyz (approved):

PR #897, opened 2022-02-24
chore(charts): update helm package

mo-auto:
Signed-off-by: mo-auto <54212639+mo-auto@users.noreply.github.com>


moabu:
This looks off. I see the config-api values schema removed. Rejecting with a close. Bot needs check

PR #948, opened 2022-03-02
feat: jans-linux-setup check if opendj ports are free

mbaser:


mbaser:
Better to check if backend is OpenDJ

Issue #962, opened 2022-03-03
config api swagger spec to include tag to distinguish plugin api


pujavs:
Swagger spec contains endpoints for config-api as well as plugin endpoins. Need to add a tag to distinguish the plugin endpoints from the core apiy

pujavs:
Changes done [PR](https://github.com/JanssenProject/jans/pull/963)

PR #963, opened 2022-03-03
feat(jans-config-api): swagger spec change to add extension to differentiate plugin en…

pujavs:
Swagger spec contains endpoints for config-api as well as plugin endpoints. Need to add a tag to distinguish the plugin endpoints from the core api
Related issue [962](https://github.com/JanssenProject/jans/issues/962)


pujavs:
@yuriyz, request your review and approval. Addition of extension property is required to be used in jans-cli to not show plugin endpoints if plugin is not deployed.

yuriyz (approved):

Issue #891, opened 2022-02-23
Allow importing custom LDIF file in persistence-loader


iromli:
`docker-jans-persistence-loader` should allow importing custom LDIF files from predefined directory,

PR #1002, opened 2022-03-09
feat: add support to import custom ldif

iromli:
### Description

- Target issue: #891

### Implementation Details

1. the entrypoint will scan for all `.ldif` files under predefined directory `/app/custom_ldif` (where users can mount files into this directory)
1. for every `.ldif` file found, entrypoint will try to generate rendered template (if file contains template context) under `/app/tmp/`
1. entrypoint will import data parsed from the generated template into user-selected persistence (`ldap`/`sql`/`spanner`/`couchbase`)

### DIFF
--- a/docker-jans-persistence-loader/scripts/couchbase_setup.py
+++ b/docker-jans-persistence-loader/scripts/couchbase_setup.py
@@ -4,6 +4,7 @@ import json
 import logging.config
 import os
 import time
+from pathlib import Path
 
 from ldif import LDIFParser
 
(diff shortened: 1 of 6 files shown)


docker-jans-persistence-loader/scripts/couchbase_setup.py
```diff
@@ -326,37 +328,10 @@ def create_indexes(self, bucket_mappings):
                             continue
                         logger.warning("Failed to execute query, reason={}".format(error["msg"]))
 
-    def import_ldif(self, bucket_mappings):
-        ctx = prepare_template_ctx(self.manager)
-        attr_processor = AttrProcessor()
-
+    def import_builtin_ldif(self, bucket_mappings, ctx):
```
moabu, line 331: What happens if the custom ldifs loaded override the builtin ldifs ?

moabu (changes requested):
we need to add a how to to our docs section. I'll add that.

docker-jans-persistence-loader/scripts/couchbase_setup.py
iromli: same behavior with builtin ldif files; data imported into persistence only if not exist

moabu: ok we can improve in operations where the same item might be modified. Like modifying the builtin ldifs in one `key` such as adding a scope to a client.
We can leave this for a separate enhancement.

moabu (approved):

Issue #348, opened 2021-02-20
Support redirect_uri regular expression matching


martynaslawinska:
_nynymike commented on Jul 23, 2019_
We could have an option to support regular expression matching for URLs for the redirect URI on a per client basis. If enabled, redirect_uri's could be evaluated against a regex pattern.

yuriyz:
@Milton-Ch after merge we got 3 test failures. Would you please check?

https://jenkins.jans.io/jenkins/job/jans-auth-server/5807/

![image](https://user-images.githubusercontent.com/2013070/158135250-07835372-1e57-4a7a-adf0-13bbed1ce994.png)

Milton-Ch:
Tests were fixed and changes were merged, closing this ticket.

PR #1005, opened 2022-03-10
feat: support regex client attribute to validate redirect uris

Milton-Ch:
### Description

- Target issue #
https://github.com/JanssenProject/jans/issues/348

- Implementation Details
1. Added client attribute to validate redirect uris using a regex.
2. Added feature flag `redirectUrisRegexEnabled`.

### DIFF
--- a/jans-auth-server/server/src/main/java/io/jans/as/server/service/RedirectionUriService.java
+++ b/jans-auth-server/server/src/main/java/io/jans/as/server/service/RedirectionUriService.java
@@ -115,12 +115,22 @@ public class RedirectionUriService {
                         client.getClientId(), redirectionUri, redirectUris.length);
                 if (isUriEqual(redirectionUri, redirectUris)) {
                     return redirectionUri;
+                } else {
+                    log.debug("RedirectionUri didn't match with any of the client redirect uris, clientId = {}, redirectionUri = {}", client.getClientId(), redirectionUri);
                 }
 
-                //if not found a match, should it print a log ?
-                return redirectionUri.matches(client.getAttributes().getRedirectUrisRegex()) ? redirectionUri : null;
+                if (appConfiguration.getRedirectUrisRegexEnabled()) {
+                    if (redirectionUri.matches(client.getAttributes().getRedirectUrisRegex())) {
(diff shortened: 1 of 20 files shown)


jans-auth-server/server/src/main/java/io/jans/as/server/service/RedirectionUriService.java
```diff
@@ -106,18 +106,19 @@ public String validateRedirectionUri(@NotNull Client client, String redirectionU
...
                     return redirectionUri;
                 }
-            } else {
-                // Accept Request Without redirect_uri when One Registered
-                if (redirectUris != null && redirectUris.length == 1) {
-                    return redirectUris[0];
-                }
+
+                //if not found a match, should it print a log ?
+                return redirectionUri.matches(client.getAttributes().getRedirectUrisRegex()) ? redirectionUri : null;
```
yuriyz: Yes, we definitely has to print in log why we rejected call. Also if you can log other places where it is rejected it will be helpful.

jans-auth-server/server/src/test/java/io/jans/as/server/service/RedirectionUriServiceTest.java
```diff
@@ -0,0 +1,147 @@
...
+    private ErrorResponseFactory errorResponseFactory;
+
+    @Mock
+    private AppConfiguration appConfiguration;
+
+    @Mock
+    private LocalResponseCache localResponseCache;
+
+    @Test
+    public void validateRedirectionUri() throws Exception {
```
yuriyz: I vote for X_Y_Z convention for test names, so it is clear X-what we test, Y - under what condition, Z - what we expect. Exactly same convention we discussed some time ago.

`validateRedirectionUri_withValidRedirectUri_shouldReturnNotNullValue`

yuriyz (changes requested):
Looks good. Some minor feedback in inlined comments.

yuriyz (approved):

moabu (approved):

mbaser (approved):

pujavs (commented):
Approved

PR #1008, opened 2022-03-10
ci: added python_version in martix

nikdavnik:
### Prepare

- [ ] Read contribution guidelines
- [ ] Read license information

-------------------

### Description

- Target issue #
  

- Implementation Details
  

-------------------
### Document the changes

### DIFF
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -21,18 +21,21 @@ jobs:
             asset_prefix: '_'
             asset_path: jans
             sign_cmd: dpkg-sig -s builder -k DE92BEF14A1A4E542F678B64DC3C790386C73900
+            python_version: 3.8
           - name: el8
             asset_suffix: .el8.x86_64.rpm
             build_files: rpm/el8
             asset_prefix: '-'
             asset_path: jans/rpmbuild/RPMS/x86_64
             sign_cmd: rpm --addsign
+            python_version: 3.6
           - name: suse15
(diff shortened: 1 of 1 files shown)


.github/workflows/build-packages.yml
```diff
@@ -42,6 +45,9 @@ jobs:
          mkdir -p jans/jans-src/opt/
          cp -rp packaging/${{ matrix.build_files }}/* jans/
          wget https://raw.githubusercontent.com/JanssenProject/jans/main/jans-linux-setup/jans_setup/install.py -O jans/install.py
+         sudo add-apt-repository ppa:deadsnakes/ppa
+         sudo apt-get update
+         sudo apt-get install -y python${{ matrix.python_version }}
```
moabu, line 50: Use matrix based form with python3 step instead of apt-get 

```
    - name: Set up Python ${{ matrix.python_version }}
      uses: actions/setup-python@v3.0.0
      with:
        python-version: ${{ matrix.python_version }}
```

moabu (changes requested):

.github/workflows/build-packages.yml
nikdavnik: @moabu not able to make install.py work with action/setup-python.
1st it installs python under /opt which works only for current user, when running install.py with sudo it can't find python. I fixed this by updating root PATH, after getting issue with install.py not funding some other python packages that ldap3 etc.

PR #1046, opened 2022-03-14
Fix: Validate redirect_uri blank and client redirect uris single item

Milton-Ch:
### Description

1. Fixing redirect_uri validator issue when single item is in client redirect uris and param received is blank.
2. Enabling `redirectUrisRegexEnabled` to run test in Jenkins.


yuriyz (approved):

yuriyz:
@Milton-Ch commit is not signed

PR #1091, opened 2022-03-21
docs: add README for technical docs

ossdhaval:
### Prepare

-------------------

### Description

- Target issue
Currently technical documentation for Janssen modules doesn't exist. Earlier(before monorepo) documentation existed under docs repo [here](https://github.com/JanssenProject/docs/tree/main/source/janssen-modules). This documentation is scant and not available in all modules.

Under `jans` monorepo, we have to start an effort to put together detailed technical design documentation for all modules.  Under `jans` monorepo, technical design documentation should go under `docs/technical`.

- Implementation Details
This PR attempts to provide basic structure and template that all code owners can follow while writing technical design documents for respective modules:
  - README for main `/technical` directory: This is supposed to capture overall Janssen design guidelines
  - Template for technical design document for code owners

Once we agree on the content and format as laid out in this PR, I can ask code owners to write detailed design docs for respective modules.

### DIFF
--- a/docs/technical/README.md
+++ b/docs/technical/README.md
@@ -19,12 +19,12 @@ Contents:
   
 | Service | REST API | Javadoc |  
 | --- | --- | --- |  
-| Jans Auth Server | | [client](https://jenkins.jans.io/javadocs/jans-scim/main/client/) [model](https://jenkins.jans.io/javadocs/jans-scim/main/model/) [server](https://jenkins.jans.io/javadocs/jans-auth/main/server/) |  
-| Jans Client API | [Swagger](https://gluu.org/swagger-ui/?url=https://raw.githubusercontent.com/JanssenProject/jans-client-api/master/server/src/main/resources/swagger.yaml)| |  
-| Jans Config API | [Swagger](https://gluu.org/swagger-ui/?url=https://raw.githubusercontent.com/JanssenProject/jans-config-api/master/docs/jans-config-api-swagger.yaml)| |  
+| Jans Auth Server | | [server](https://jenkins.jans.io/javadocs/jans-auth/main/server/) [model](https://jenkins.jans.io/javadocs/jans-auth/main/model/) [client](https://jenkins.jans.io/javadocs/jans-auth/main/client/)|  
+| Jans Client API | [Swagger](https://gluu.org/swagger-ui/?url=https://raw.githubusercontent.com/JanssenProject/jans/main/jans-client-api/server/src/main/resources/swagger.yaml)| |  
+| Jans Config API | [Swagger](https://gluu.org/swagger-ui/?url=https://raw.githubusercontent.com/JanssenProject/jans/main/jans-config-api/docs/jans-config-api-swagger.yaml)| |  
 | Jans Core | | [server](https://jenkins.jans.io/javadocs/jans-core/main/io/jans/server/filters/package-summary.html)|  
-| Jans FIDO 2 | [Swagger](https://gluu.org/swagger-ui/?url=https://raw.githubusercontent.com/JanssenProject/jans-fido2/master/docs/jansFido2Swagger.yaml) | |  
-| Jans SCIM API | [Swagger](https://gluu.org/swagger-ui/?url=https:/raw.githubusercontent.com/JanssenProject/jans-scim/master/server/src/main/resources/jans-scim-openapi.yaml) | [client](https://jenkins.jans.io/javadocs/jans-scim/main/client/) [model](https://jenkins.jans.io/javadocs/jans-scim/main/model/) [server](https://jenkins.jans.io/javadocs/jans-scim/main/server/) |  
+| Jans FIDO 2 | [Swagger](https://gluu.org/swagger-ui/?url=https://raw.githubusercontent.com/JanssenProject/jans/main/jans-fido2/docs/jansFido2Swagger.yaml) | |  
+| Jans SCIM API | [Swagger](https://gluu.org/swagger-ui/?url=https://raw.githubusercontent.com/JanssenProject/jans/main/jans-scim/server/src/main/resources/jans-scim-openapi.yaml) | [server](https://jenkins.jans.io/javadocs/jans-scim/main/server/) [model](https://jenkins.jans.io/javadocs/jans-scim/main/model/) [client](https://jenkins.jans.io/javadocs/jans-scim/main/client/) |  
   
 ## Design Consideration and Guidelines
 This section outlines high-level design principles, styles and design choices that Janssen project follow. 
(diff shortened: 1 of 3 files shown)


docs/technical/README.md
```diff
@@ -1,3 +1,36 @@
...
+### Security
+### Scallability and Cloud Infrastructure 
+
+## Technical Documentation Guidelines
+  
+Each module directory will hold detailed technical documentation for that module. It should have README file that follows [Technical Overview Template](./technical-overview-template.md). Also create other directories as required and as mentioned in `Directory Structure` section below. 
+
+### Directory Structure
+Create directories within your module's directory as below to support your documetation
+ - `images` : to hold screenshots etc
```
moabu: Let's specify the image params like size limit . It's very important we don't clog our repo

moabu (changes requested):

docs/technical/README.md
ossdhaval: About clogging of the repo. Should we rethink about what I have proposed in this PR with respect to _where we store our media assets?_ 

In this PR, I have put media assets under `docs` directory. Similar media assets would exist in `user` documentation and `developer` documentation. Yes, we can put recommendation, but there is no way to stop people from uploading larger files (GH allows up to 100MB). Plus, over time, we will add GIF(tutorials), or may be video files for that matter. If we keep media assets at different places, it'll hard to control, review and clean-up. 

So I am having second thoughts about this arrangement. Now I think that keeping all media assets in a separate top-level directory (`jans/media-assets`) would be better. It'll be easy to keep in check the size of one directory. Also, maybe later we can add automation using GH actions where we can have an action trigger on commit in `jans/media-assets`. That action can make sure the newly added asset is in-line with recommended size and type etc.

docs/technical/README.md
```diff
@@ -1,3 +1,45 @@
...
+Contents:
+
+- [API Reference](#api-reference)
+- [Design Consideration and Guidelines](#design-consideration-and-guidelines)
+  - [REST API Design](#rest-api-design)
+  - [Caching](#caching)
+  - [Testing](#testing)
+  - [Deployment](#deployment)
+  - [Security](#security)
+  - [Scalability and Cloud Infrastructure](#scalability-and-cloud-infrastructure)
```
ossdhaval, line 14: @yuriyz It would be great if you can contribute content for these guidelines. Since you have seen Janssen design evolve over time, you would have the right background to drive this. Since these guidelines are applicable to Janssen in general, other code owners can also chip in too. @moabu @yurem

docs/technical/README.md
moabu: I think we should keep all assets in the docs and we shouldn't break it off. As for the size and type just mention that in the docs recs so people know. Its the responsibility of the reviewer of the docs PR to make sure the image/diagram is of spec. We will add a todo to check the assets directories in a PR to make sure it meets the specs we specified.

Two options:
1. Each module will have a predefined set of folders such as `images` and `diagrams`.
2. You have one folder  `docs/assets` that contains all the modules and their folders `images` and `diagram` .

I would go with the first option and hence you need to add empty folders for each module. 

The action we will add will check recursively for any folder that has the name `images` or `diagram`

ossdhaval: I see few problems if we go for multiple directories:
- directories outside of the `docs` may also require to store media files.
- In the future, new directories will have to be created. Like `videos`, `GIFs` etc. And all this will happen in various directories. 
- Yes, PR reviewers should ideally be vigilant about the size fo the uploaded files, but it is highly likely to be missed. Only Automated checks will prove to be helpful in the long run
- Cleaning up would be difficult as it'll be difficult to understand what is the total size of media being held in repo as media will be scattered. 
- And it'll be an additional risk that GH actions would fail if someone has named the directory `image` instead of `images`. Controlling naming standards would be another problem. Instead, if everything is in one place, there will be only one `images` directory and that is it. Developers will just have drop their images in there. Not hassle of creating new directories. Easy.

moabu: Yeah that's option 2 and that sounds fine as long as it's in the `docs/assets` folder

ossdhaval: I am good with option 2. It is a middle path. 
In your description of option 2 above, are you suggesting that we should have a folder for each module under `docs/assets`? I think that is not required because there is no benefit in separating images by modules. `docs/assests` can just have two folders `images` and `diagrams` and that is it. Images for all modules stay together. And developers can refer to them in their documentation. Do you see benefit of separating them by modules?

docs/technical/README.md
yuriyz: Are these guidelines about AS or in general for any janssen project? 
E.g. AS is bound to spec and is developed according to it. While in jans-config-api we are free to build API which are not restricted by anything. 
If idea here is about general consideration then it should come from generic best practices which are not bound to any specific project (e.g. AS, cli, config-api). Otherwise each project needs own specific page where each code owner can describe current design and guidelines. Placing description of everything on single page looks overwhelming. What do you think?

ossdhaval: Idea is to document and publish architectural choices that are to be followed by all modules. 

For example,
- under `Testing` we can mention what mocking frameworks and tools that we use. And advise people against adding new test frameworks and so on.
- under `caching` we can mention that considering cloud-native architecture, what is our approach towards caching.
- under `security` we can mention that when adding a new dependency, please check what is the security profile of that library first. So that later we don't see our security rating go down.
- we can have guidelines around `logging` where we follow a common format.

Hope this makes idea behind this section more clear. 

If we don't have these documented, we may see ourselves repeating in review of every major change. Or worse, risk missing out of some of these.

And this would be an evolving list. Not necessary that we have everything listed here in one go.

yuriyz: Got it. Yes, makes perfect sense. We can start with something and then during reviewes we can fill it with more and more info to not repeat it.

docs/technical/README.md
moabu: yes I think they should be separated by module. The likelihood that they asset is related to a module ( a service in the module) is high. That way we can offer a bit more distinction where these assets are being used . So take a look at it in from a reverse standpoint a document points to a link but with separate modules we also tie the assets to the module. Looking where to edit or where to fix will be a lot easier especially if we have a lot. We also provide a safe guard against miss use that way and drive adoption. Easier for the dev to find so easier to write a doc.  I also want to do this in case later on some of these assets get inherited or used by code. Doing this will certainly help. Otherwise we can prefix with `service_` such as `jans_auth_server_*`. We also need a clear nomenclature for the images. This helps with automation , cleanup ...etc so it should be added to this doc.

ossdhaval: Agreed. :+1: 
I'll make changes in the instructions accordingly.

docs/technical/README.md
ossdhaval: :+1: 
Let's track this documetation effort in a separate issue #1124.

docs/technical/README.md
ossdhaval: Refactored the instructions. Please review.

moabu (approved):

moabu (approved):

Issue #992, opened 2022-03-07
tests(jans-auth-server): jans-auth-model is module used by other projects. We need cover it better with tests.


yuriyz:
tests(jans-auth-server): jans-auth-model is module used by other projects. We need cover it better with tests.

- All utils in model module has to be covered with tests
- All classes where we have methods with logic must be covered with tests

yuriyz:
Lets get rid of sonar issues within this ticket as well in  jans-auth-model

yuriyz:
It's all utility classes and classes/methods where we have logic. I will take over this ticket.

ossdhaval:
Love this ticket. Much needed. 

@yuriyz @nynymike For community engagement in Jans, we need to have more issues with low complexity marked as `Good First issue`. I have always imagined that Sonar issues could be such issues. Developers who want to contribute to open source,  can spot these and work on them.

Here is what I would suggest. Our core team members like @yuriyz , take up more complex tasks like the adding of tests. And we have a separate issue for fixing sonar bugs which we mark as a `good first issue` and let the community pick it up. 

I have had a positive experience with the use of the `good first issue` label and I can see it attracts developers. Two examples where we got contributions to Jans from the community are [here](https://github.com/JanssenProject/jans/issues/4120) and [here](https://github.com/JanssenProject/jans/issues/2521). We need to do more of this.

Any thoughts are welcome.

PR #1130, opened 2022-03-30
test: jans-auth-model cleaning sonar issues

jmunozherbas:
Refactor cleaning sonar issues in model module.
Target Issue #
https://github.com/JanssenProject/jans/issues/992

This is the second PR for this issue, please do not close the Issue.
Main changes:
- Change interfaces with only constants, to classes with public final constants.
- Add Particular Exceptions, cleaning general Exceptions.
- Refactor complex methods.
- Remove direct reflection modifiers.
- Add Type to Generic Class declarations.
- Changes in other modules are minimum, to adapt to jans-auth-model changes.


jans-auth-server/model/src/main/java/io/jans/as/model/authorize/AuthorizeRequestParam.java
```diff
@@ -10,41 +10,43 @@
  * @author Javier Rojas Blum
  * @version October 7, 2019
  */
-public interface AuthorizeRequestParam {
+public class AuthorizeRequestParam {
```
yuriyz: Constant classes must be also `final`

jans-auth-server/model/src/main/java/io/jans/as/model/authorize/AuthorizeResponseParam.java
```diff
@@ -10,29 +10,31 @@
  * @author Javier Rojas Blum
  * @version July 28, 2021
  */
-public interface AuthorizeResponseParam {
+public class AuthorizeResponseParam {
```
yuriyz: declare `final`

jans-auth-server/model/src/main/java/io/jans/as/model/configuration/ConfigurationResponseClaim.java
```diff
@@ -10,73 +10,76 @@
  * @author Javier Rojas Blum
  * @version September 30, 2021
  */
-public interface ConfigurationResponseClaim {
+public class ConfigurationResponseClaim {
```
yuriyz: declare `final`

jans-auth-server/model/src/main/java/io/jans/as/model/ciba/BackchannelAuthenticationRequestParam.java
```diff
@@ -10,71 +10,74 @@
  * @author Javier Rojas Blum
  * @version May 28, 2020
  */
-public interface BackchannelAuthenticationRequestParam {
+public class BackchannelAuthenticationRequestParam {
```
yuriyz: Declare `final`

jans-auth-server/model/src/main/java/io/jans/as/model/crypto/encryption/BlockEncryptionAlgorithm.java
```diff
@@ -14,8 +14,8 @@
  */
 public enum BlockEncryptionAlgorithm {
 
-    A128CBC_PLUS_HS256("A128CBC+HS256", "CBC", "AES/CBC/PKCS5Padding", "SHA-256", "HMACSHA256", 256, 128, 128),
-    A256CBC_PLUS_HS512("A256CBC+HS512", "CBC", "AES/CBC/PKCS5Padding", "SHA-512", "HMACSHA512", 512, 128, 256),
+    A128CBC_PLUS_HS256(new BasicEncryptionData("A128CBC+HS256", "CBC", "AES/CBC/PKCS5Padding", 256, 128), "SHA-256", "HMACSHA256",  128),
+    A256CBC_PLUS_HS512(new BasicEncryptionData("A256CBC+HS512", "CBC", "AES/CBC/PKCS5Padding", 512, 128), "SHA-512", "HMACSHA512", 256),
```
yuriyz: Nothing really changed. I propose to drop this redundant `BasicEncryptionData` class and suppress warning. For me it does not make any sense to produce extra code for enum contructor which can't be called outside of enum itself.
Use for enum constructor that has more then 7 arguments `@SuppressWarning(java:S107"")`.

jans-auth-server/model/src/main/java/io/jans/as/model/discovery/WebFingerParam.java
```diff
@@ -9,26 +9,28 @@
 /**
  * @author Javier Rojas Blum Date: 09.16.2013
  */
-public interface WebFingerParam {
+public class WebFingerParam {
```
yuriyz, line 12: Declare `final`

jans-auth-server/model/src/main/java/io/jans/as/model/crypto/encryption/BasicEncryptionData.java
```diff
@@ -0,0 +1,57 @@
+package io.jans.as.model.crypto.encryption;
+
+public class BasicEncryptionData {
```
yuriyz: We should remove this class, see comment in [BlockEncryptionAlgorithm.java](https://github.com/JanssenProject/jans/pull/1130/files#diff-3b399cb6a0a68b787e62b1c7e1d67d141be0f2bcb5eabf42b693cb74e97803bd)

jans-auth-server/model/src/main/java/io/jans/as/model/error/ErrorResponseFactory.java
```diff
@@ -98,7 +97,8 @@ public String getErrorAsJson(IErrorType type) {
 
     public String errorAsJson(IErrorType type, String reason) {
         final DefaultErrorResponse error = getErrorResponse(type);
-        error.setReason(appConfiguration.getErrorReasonEnabled() ? reason : "");
+        boolean errorReasonEnabled = appConfiguration.getErrorReasonEnabled() != null && appConfiguration.getErrorReasonEnabled().booleanValue();
+        error.setReason(errorReasonEnabled ? reason : "");
```
yuriyz: `error.setReason(BooleanUtils.isTrue(appConfiguration.getErrorReasonEnabled()) ? reason : "")`

jans-auth-server/model/src/main/java/io/jans/as/model/json/JsonApplier.java
```diff
@@ -51,9 +55,7 @@ private void apply(Object source, Map<String, String> target, PropertyDefinition
                 return;
             }
 
-            Field field = source.getClass().getDeclaredField(property.getJavaTargetPropertyName());
-            field.setAccessible(true);
-            Object value = field.get(source);
+            Object value = invokeReflectionGetter(source, property.getJavaTargetPropertyName());
```
yuriyz, line 58: Looks good but run both `JsonApplierServerTest` and `JsonApplierTest` to check nothing is broken after this change.

jans-auth-server/model/src/main/java/io/jans/as/model/token/JsonWebResponse.java
```diff
@@ -22,8 +22,8 @@
 
     private static final long serialVersionUID = -4141298937204111173L;
 
-    protected JwtHeader header;
-    protected JwtClaims claims;
+    protected transient JwtHeader header;
+    protected transient JwtClaims claims;
```
yuriyz: header and claims has to be done serializable instead of declaring transient. Otherwise not clear what we serialize.

jans-auth-server/model/src/main/java/io/jans/as/model/util/StringUtils.java
```diff
@@ -46,15 +46,10 @@ public static String nullToEmpty(String str) {
...
-        } else if (str1 == null && str2 != null) {//note: str2!=null is always NOT null, see previous 'if' statement*/
-            return false;
-        } else if (str1 != null && str2 == null) {  //note: str1!=null is ALWAYS true
-            return false;
-        } else {
-            return str1 != null && str1.equals(str2);
+        if (str1 == null) {
+            return str2 == null;
         }
+        return str1.equals(str2);
```
yuriyz: Probably it can be directly `org.apache.commons.lang3.StringUtils#equals` and mark this method deprecated and remove later.

jans-client-api/uma-rs-resteasy/src/main/java/io/jans/ca/rs/protect/resteasy/ResourceRegistrar.java
```diff
@@ -112,7 +112,7 @@ private void register(RsResource rsResource) {
 
         } catch (Exception ex) {
             LOG.error(ex.getMessage(), ex);
-            throw ex;
+            throw (RuntimeException) ex;
```
yuriyz: It is problem, not every exception is RuntimeException. Why do we have to change this line at all ?

yuriyz (changes requested):
Please check inlined comments.

jmunozherbas:
@yuriyz  your comments was solved, please check the new commit.

jans-client-api/uma-rs-resteasy/src/main/java/io/jans/ca/rs/protect/resteasy/ResourceRegistrar.java
jmunozherbas: One sonar issue was https://rules.sonarsource.com/java/type/Code%20Smell/RSPEC-112
due to RuntimeException throw, I prefered to supress the sonar warning to avoid more changes.

jans-auth-server/model/src/main/java/io/jans/as/model/token/JsonWebResponse.java
jmunozherbas: I have implemented writeObject and readObject to allow serialization. 
It was because Map<String,Object> in JwtClaimSet (parent) is not serializable.
I have implemented one unit test too, to check serialization.

yuriyz (approved):

PR #1150, opened 2022-04-01
feat(jans-auth): added jwkThumbprint attribute to JSONWebKey to fix j…

pujavs:
**Issue:** Jenkins build failure for jans-config-api

**Analysis:** jans-config-api is failing because of `jwkThumbprint` attribute being **removed** from https://github.com/JanssenProject/jans/blob/main/jans-auth-server/model/src/main/java/io/jans/as/model/jwk/JSONWebKey.java
This is causing error-
`01-04 12:13:08.795 ERROR jans.configapi.configuration.ConfigurationFactory ConfigurationFactory.java:436- Failed to convert json value '{"keys":[....' to object of type class io.jans.as.model.config.WebKeysConfiguration`

**Logs:**  [2022_04_01.jetty.log](https://github.com/JanssenProject/jans/files/8398296/2022_04_01.jetty.log)

**Fix:** Added  `jwkThumbprint` attribute **back** in  https://github.com/JanssenProject/jans/blob/main/jans-auth-server/model/src/main/java/io/jans/as/model/jwk/JSONWebKey.java

### DIFF
--- a/jans-auth-server/model/src/main/java/io/jans/as/model/jwk/JSONWebKey.java
+++ b/jans-auth-server/model/src/main/java/io/jans/as/model/jwk/JSONWebKey.java
@@ -93,9 +93,8 @@ public class JSONWebKey {
     }
 
     /**
-     * Returns the Key ID. The Key ID member can be used to match a specific key.
-     * This can be used, for instance, to choose among a set of keys within the JWK
-     * during key rollover.
+     * Returns the Key ID. The Key ID member can be used to match a specific key. This can be used, for instance,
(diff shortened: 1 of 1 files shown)


yuriyz:
Interesting, from where do we got `jwkThumbprint` in the key json. It is dynamically calculated value. Is it to avoid re-computation?

yuriyz:
jwkThumbprint was serialized by mistake. Avoid serialization in this PR : https://github.com/JanssenProject/jans/pull/1164

Closing this PR as redundant. config-api build is successful now (as told by mister jenkins).

PR #1183, opened 2022-04-08
fix: update api-admin permissions from config api yaml

mbaser:
We were using static permissions for role based auth. Since it is static, developers have to update manually when they add new scope.  With this PR scopes are updated form jans-config-api swagger yaml file.


pujavs (approved):

jans-linux-setup/jans_setup/templates/jans-auth/role-scope-mappings.json
```diff
@@ -0,0 +1,380 @@
...
+        "https://jans.io/oauth/config/database/sql.write",
+        "readonly",
+        "https://jans.io/oauth/config/stats.readonly",
+        "jans_stat"
+      ]
+    },
+    {
+      "role": "api-admin",
+      "permissions": [
+        "https://jans.io/oauth/config/attributes.readonly",
```
duttarnab, line 319: Please add below `permission` to `api-admin` role mapping
https://jans.io/oauth/jans-auth-server/config/properties.readonly

devrimyatar: > Please add below `permission` to `api-admin` role mapping https://jans.io/oauth/jans-auth-server/config/properties.readonly

Don't worry about it. All scopes from config-api swagger yaml file is added to `api-admin`

yuriyz (approved):

Issue #1057, opened 2022-03-16
feat: auth Server should Support Client Registration Language Metadata


nynymike:
[Section 2.1 of OpenID Client Reg](https://openid.net/specs/openid-connect-registration-1_0.html#LanguagesAndScripts) states:

   Human-readable Client Metadata values and Client Metadata values that reference human-readable values MAY be 
   represented in multiple languages and scripts. For example, values such as client_name, tos_uri, policy_uri, logo_uri, and 
   client_uri might have multiple locale-specific values in some Client registrations.

   To specify the languages and scripts, [BCP47](https://openid.net/specs/openid-connect-registration-1_0.html#RFC5646) 
   [RFC5646] language tags are added to Client Metadata member names, delimited by a # character. The same syntax is used 
   for representing languages and scripts for Client Metadata as is used for Claims, as described in Section 5.2 (Claims 
   Languages and Scripts) of [OpenID Connect Core 1.0](https://openid.net/specs/openid-connect-registration-
   1_0.html#OpenID.Core) [OpenID.Core].

   If such a human-readable field is sent without a language tag, parties using it MUST NOT make any assumptions about the 
   language, character set, or script of the string value, and the string value MUST be used as-is wherever it is presented in a 
   user interface. To facilitate interoperability, it is RECOMMENDED that any human-readable fields sent without language tags 
   contain values suitable for display on a wide variety of systems.

* The highest priority claims are `client_name` and `client_uri`.

qbert2k:
The Client metadata values: client_name (displayName), logo_uri (jansLogoURI), client_uri (jansClntURI), policy_uri (jansPolicyURI), and tos_uri (jansTosURI) now stores a JSON with values for multiple languages and scripts. For example:

### MySQL, Couchbase, Spanner, etc.

##### client_name (displayName)
![image](https://user-images.githubusercontent.com/2474616/164330221-a95dcd62-8be9-42e1-be35-434a8c4adbd3.png)

##### logo_uri (jansLogoURI)

![image](https://user-images.githubusercontent.com/2474616/164330837-d1b7c51f-4a39-48f8-b0f2-3cff7acf5a25.png)

##### client_uri (jansClntURI)

![image](https://user-images.githubusercontent.com/2474616/164331213-5b7f69fd-a52e-475d-bdde-8986e961d876.png)

##### policy_uri (jansPolicyURI)

[...]

PR #1237, opened 2022-04-20
Jans auth server client registration language metadata

qbert2k:
### Prepare

- [ ] Read [PR guidelines](https://github.com/JanssenProject/jans/blob/main/docs/CONTRIBUTING.md#prs)
- [ ] Read [license information](https://github.com/JanssenProject/jans/blob/main/LICENSE)

-------------------

#### Target issue
https://github.com/JanssenProject/jans/issues/1057

#### Implementation Details
See https://github.com/JanssenProject/jans/issues/1057

-------------------


jans-orm/model/src/main/java/io/jans/orm/model/base/ClientMetadataValue.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.io.Serializable;
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * @author Javier Rojas Blum
+ * @version April 18, 2022
+ */
+public class ClientMetadataValue implements Serializable {
```
yuriyz: What do you think about naming it `LocalizedString` ? In this way it is more generic and can be used in other places where we need localized string without concern.

jans-orm/model/src/test/java/io/jans/orm/model/base/ClientMetadataValueTest.java
```diff
@@ -0,0 +1,58 @@
...
+        clientMetadataValue.setValue("Client name", Locale.forLanguageTag("en-NZ"));
+        clientMetadataValue.setValue("Nom du client", Locale.FRENCH);
+        clientMetadataValue.setValue("Nom du client", Locale.CANADA_FRENCH);
+        clientMetadataValue.setValue("Nom du client", Locale.FRANCE);
+        clientMetadataValue.setValue("クライアント名", Locale.forLanguageTag("ja-Jpan-JP"));
+        clientMetadataValue.setValue("カナ姓※", Locale.forLanguageTag("ja-Kana-JP"));
+        clientMetadataValue.setValue("漢字姓※", Locale.forLanguageTag("ja-Hani-JP"));
+
+        System.out.println(clientMetadataValue);
+        assertEquals(clientMetadataValue.size(), 12);
```
yuriyz: We should check serialized data contains format which we expect.

jans-auth-server/client/src/main/java/io/jans/as/client/RegisterRequest.java
```diff
@@ -1505,6 +1648,20 @@ public static RegisterRequest fromJson(JSONObject requestObject) throws JSONExce
         return result;
     }
 
+    private static void clientMetadataValueFromJson(
```
yuriyz: We need static generic method in `LocalizedString` (`ClientMetadataValue`), it can be used in many places.

```
private static void fromJson(JSONObject requestObject, String key, BiFunction<String, Locale, Void> function) {
    ...
}
```

jans-auth-server/client/src/main/java/io/jans/as/client/RegisterRequest.java
```diff
@@ -1790,4 +1938,28 @@ public String getRedirectUrisRegex() {
...
+        if (clientMetadataValue != null && clientMetadataValue.size() > 0) {
+            for (String languageTag : clientMetadataValue.getLanguageTags()) {
+                if (StringUtils.isBlank(languageTag)) {
+                    parameters.put(registerRequestParam.toString(), clientMetadataValue.getValue());
+                } else {
+                    parameters.put(registerRequestParam.toString() + "#" + languageTag, clientMetadataValue.getValue(languageTag));
+                }
+            }
+        }
+    }
```
yuriyz: ClientMetadataValue -> LocalizedString.
We can add method to `LocalizedString` and reuse it.
```
    public Map<String, String> addToMap(Map<String, String> map, String key) {
        if (!values.isEmpty) {
            return map;
        }

        for (String languageTag : getLanguageTags()) {
            if (StringUtils.isBlank(languageTag)) {
                map.put(key, getValue());
            } else {
                map.put(key + "#" + languageTag, getValue(languageTag));
            }
        }
        return map;
    }
```

jans-auth-server/client/src/main/java/io/jans/as/client/RegisterRequest.java
```diff
@@ -1790,4 +1938,28 @@ public String getRedirectUrisRegex() {
...
+
+    private void addClientMetadataParam(JSONObject parameters, RegisterRequestParam registerRequestParam, ClientMetadataValue clientMetadataValue) {
```
yuriyz: If add `addToMap()` method then this I guess can be tricked as (pseudo code that has to be checked.): 

`new JSONObject(localicalizedString.addToMap(jsonObject.ToMap(), registerRequestParam.toString()));`

Advantage is that we can re-use `addToMap` in other places.

yuriyz (changes requested):
Added some improvements proposals, please check.

qbert2k:
Hi, @yuriyz
Regarding the sonar issues:

1. The dupplicated lines of code are between Client.java and RegisterRequest.java and because they are not related and I think hineritance is not a good solution. I am thinking to use Lombok, what do you think?
2. There is annother compilation issue bloking the merge request. A compilation issue in jans-client-api but it seems not related to the changes in jans-auth-server.

yuriyz:
Yes, I like lombok idea to reduce boilerplate code. Re: jans-client-api - ignore it. @jmunozherbas will correct it.

But if PR is ready we can let it go. Lets add lombok in separate PR. I don't want to put everything in one bucket. Duplicated lines is not end of the world. @qbert2k otherwise PR is ready, right ?

yuriyz (approved):

PR #1275, opened 2022-04-30
refactor: remove all Casa manifests

iromli:
### Description

The manifests to create config, client, script for Casa integration is now removed from the core as they should be taken care by
`gluufederation/casa` image.

Note, `CN_CASA_ENABLED` env var is no longer needed.


moabu (approved):
Charts need to be updated . Assigning myself on a different issue.

PR #1295, opened 2022-05-05
feat: jans-cli obtain list of attrbiutes from server when creating user

mbaser:
When creating user we add each attrbiute as **customAttributes** schema  in the following format:
```
     {
        "name": "attribName",
        "multiValued": false,
        "values": [
          "attribValue"
        ]
      }
```

Admin should type **attribName* but he does not know which attributes are allowed. In this PR, CLI obtains list of attribute names from endpoint `/jans-config-api/api/v1/attributes` of jans-config-api, and provides as suggestion:

image

### DIFF
--- a/jans-cli/cli/config_cli.py
+++ b/jans-cli/cli/config_cli.py
@@ -303,11 +303,13 @@ class JCA_CLI:
         self.enums()
 
     def enums(self):
-        self.enum_dict = {}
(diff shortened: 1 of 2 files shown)


moabu (approved):
Looks good . We need to dig deep in adding documentation strings and also sphinx

PR #1301, opened 2022-05-06
chore(charts): update helm package

mo-auto4:
This is an automated message propagated by a change in the parent Gluu chart. It will be automerged. Don't think about touching me before reaching out to @moabu.


moabu:
issue in the chart generation. Closing

Issue #1307, opened 2022-05-06
feat(docker): review ownership and permission


iromli:
## Problems

Current images adjust ownership and permission for several directories and files for non-root user. The issue is some of the commands are using `chown`, for example:

```
RUN chown -R 1000:1000 /opt/jans/jetty
```

Unfortunately, due to docker build limitation, the `chown` command will create extra layer and add the size as huge as the `chown`-ed directory/file size.

**Build using chown command**

For illustration, inspecting image layers using `docker images` will show the `SIZE`:

```
» docker images | grep config-api
REPOSITORY                               TAG           IMAGE ID       CREATED          SIZE
janssenproject/config-api                1.0.0_dev     3b4b039ce33e   27 hours ago     632MB
```

By using `docker history` command, we know that adjusting ownership and permission via `chown` increased the size of the image (approx. 164MB), as seen below:

```
» docker history janssenproject/config-api:1.0.0_dev                                                                                                  
IMAGE          CREATED        CREATED BY                                      SIZE      COMMENT                                                                          
<missing>      27 hours ago   RUN |5 JETTY_VERSION=11.0.8 JETTY_HOME=/opt/…   164MB     buildkit.dockerfile.v0      
```

**Build without using chown command**

Temporarily commenting out `chown` command produces smaller image size:

```
» docker images | grep config-api
REPOSITORY                               TAG           IMAGE ID       CREATED          SIZE
janssenproject/config-api                1.0.0_dev     dabb9115a60c   3 minutes ago    468MB
```

Now the history shows small size while adjusting permission:

```
» docker history janssenproject/config-api:1.0.0_dev                                                                                                  
IMAGE          CREATED         CREATED BY                                      SIZE      COMMENT                                                      
<missing>      5 minutes ago   RUN |5 JETTY_VERSION=11.0.8 JETTY_HOME=/opt/…   4.66kB    buildkit.dockerfile.v0                               
```

## Goals

- Review the usage of `chown` or similar command that changes ownership and permission of directories and files
- If `chown` is inevitable, only adjust specific directories and files (i.e. `chown /opt/jans/jetty/jans-config-api/logs` instead of `/opt/jans/jetty/jans-config-api`; the latter contain `webapps/jans-config-api.war` which is a huge file)
- Re-test to avoid regression

PR #1312, opened 2022-05-09
feat: adjust ownership and permission to avoid bloated images

iromli:
### Description

The changeset fixes incorrect usage of ownership/permission (in docker context) as reported in #1307 

#### Implementation Details

- remove `chown` command as it increases the size of the image
- replace `chgroup -R` command with `chmod -R g=u` to set directory ownership/permission
- use `chmod 664` command to set file permission


moabu:
reviewing....im going to run some tests before merging.

moabu (approved):

PR #1401, opened 2022-05-19
fix(jans-fido2): temp switch to bom snapshot

yuriyz:
fix(jans-fido2): temp switch to bom snapshot


yuriyz:
We will build it via branch, closing.

Issue #260, opened 2020-11-26
Re-design oxd Connect part to work inside user-agent instead of as backchannel app.


duttarnab:
Refer : https://github.com/GluuFederation/oxd/issues/462

yuriyz:
Latest dropwizard still sticks to jetty 9 (`<jetty.version>9.4.45.v20220203</jetty.version>`).
It can be right time to drop dropwizard and migrate to weld/resteasy and get it works exactly in same way as AS and config-api work.

https://github.com/dropwizard/dropwizard/blob/release/2.1.x/dropwizard-dependencies/pom.xml#L52

yuriyz:
Yes, it is good amount of work. Also deployment scripts has to be corrected. Instead of standalone run it would be deployment of war to jetty (same story as with AS and jans-config-api).

yuriyz:
Done within https://github.com/JanssenProject/jans/pull/1319

PR #1475, opened 2022-05-31
fix(jans-orm):sql localizedstring persistence SqlEntryManager

jmunozherbas:
### Description
SqlEntryManager: method to get LocalizedString values was modified similar to SqlLdapEntryManager

detected trying to persist sites with mysql DB.

#### Origin Issue
https://github.com/JanssenProject/jans/issues/260

### DIFF
--- a/jans-orm/sql/src/main/java/io/jans/orm/sql/impl/SqlEntryManager.java
+++ b/jans-orm/sql/src/main/java/io/jans/orm/sql/impl/SqlEntryManager.java
@@ -18,6 +18,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 
+import io.jans.orm.model.base.LocalizedString;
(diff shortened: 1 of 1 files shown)


yuriyz (approved):

yurem:
@yuriyz @jmunozherbas Why we need this method on ORM layer. It not works with DB. The better place for it is in reusable service layer.

yurem:
@qbert2k Can you share design related to `LocalizedString`?

Why we simply can't use
    @JsonObject
    @AttributeName(name = "attrbute_name")
    private LocalizedString attributeValueWhichOrmWill convertToJson;

jmunozherbas:
> @jmunozherbas in PR description you added reference to #260 But it's about another update. Is there design of this update?

I mentioned Issue, because it was the origin of the change, but doesn't have a specific Issue.

PR #1568, opened 2022-06-16
ci: jans automation test using jans-cli

manojs1978:
### Prepare

-------------------

### Description
created automation script to test jans-cli using expect and shell scripts
#### Implementation Details
created script to check what is machine os ,
delete existing jans server
install latest jans server using install.py
test jans-cli command line utility with all modules.
created log folder to store all logs ( installation,jans test log)
create test report.( pass/Fail) in html format.
**Input:**
setup script needs 2 inputs:

./setup.sh -h 64.227.6.77 -u root -p Pass@1222.

other scripts called within setup.sh

**OS supported**

ubuntu 20.04
centos 8
Suse 15
RedHat Linux

-------------------


automation/tests/report/result.txt
```diff
@@ -0,0 +1,23 @@
...
+couchbaseDB-configuration Passed
+openID-connnect Passed
+UMA Failed
+OAuth-Scopes Passed
+Statistics Passed
+health Passed
+server-Statistics Passed
+scim-user-mgmt Failed
+Organization-Configuration Passed
+AuthServerHealth Passed
```
moabu: remove this file completely and parent folder

automation/tests/install.exp
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/expect -f
```
moabu: keep it simple. A clean bash script. The installation process should not be longer than 50 lines and this is prone to error. If upstream changes a prompt this script would fail. Lets use a properties file and execute what a user would.

Create a function and do something like this 

```bash

install_jans() {
  echo "*****   Writing properties!!   *****"
  echo "ip=${IP_ADDRESS}" | sudo tee /install/community-edition-setup/setup.properties > /dev/null
  echo "hostname=${HOSTNAME}" | sudo tee -a /install/community-edition-setup/setup.properties > /dev/null
  echo "orgName=${ORG_NAME}" | sudo tee -a /install/community-edition-setup/setup.properties > /dev/null
  echo "admin_email=${EMAIL}" | sudo tee -a /install/community-edition-setup/setup.properties > /dev/null
  echo "city=${CITY}" | sudo tee -a /install/community-edition-setup/setup.properties > /dev/null
  echo "state=${STATE}" | sudo tee -a /install/community-edition-setup/setup.properties > /dev/null
  echo "countryCode=${COUNTRY}" | sudo tee -a /install/community-edition-setup/setup.properties > /dev/null
  # shellcheck disable=SC2016
  echo 'ldapPass=${ADMIN_PASS}' | sudo tee -a /install/community-edition-setup/setup.properties > /dev/null
  # shellcheck disable=SC2016
  # after passing the params execute the install specifying the properties file
  
 ...
 ...
 }
 install_jans

File: automation/tests/install.exp [hunk 3]
```diff
@@ -0,0 +1,109 @@
...
+set timeout -1
+spawn $env(SHELL)
+match_max 100000
+expect  "$"
+send -- "ssh ${USERNAME}@${HOST}"
+expect -exact "${HOST}"
+send -- "\r"
+expect -re "(.*)\n"
+send -- "curl https://raw.githubusercontent.com/JanssenProject/jans/main/jans-linux-setup/jans_setup/install.py > install.py"
+expect -exact "curl https://raw.githubusercontent.com/JanssenProject/jans/main/jans-linux-setup/jans_setup/install.py > install.py"
```
moabu: This is fine but allow to pass an arg to the script specifying the branch instead of hardcoding it to `main` . In that sense when we execute it we can specify the branch to execute from too.

automation/tests/jans-cli-test.exp
```diff
@@ -0,0 +1,1310 @@
+#!/usr/bin/expect -f
```
moabu, line 1: Same comment here. Instead of expecting a prompt, pass an argument. loop through the different arguments for the different tests. This should drop around 1000 lines.

automation/tests/setup.sh
```diff
@@ -0,0 +1,315 @@
...
+helpFunction()
+{
+   echo ""
+   echo "Usage:  -h servername -u username -p passwd"
+   echo -e "\t-host name of the host server -u username of host -p passwd of user "
+   exit 1 # Exit script after printing help
+}
+
+MODULE="Attribute defaultMethods cacheconfig memcacheconf memory-cache-configue redis-cache-configuration in-Memory-cache-configuration native-persist-cacheconf Configuration-property fido2-configuration SMTP-configuration Logconfiguration LDAP-configuration couchbaseDB-configuration openID-connnect UMA OAuth-Scopes Statistics health server-Statistics scim-user-mgmt Organization-Configuration AuthServerHealth"
+
```
moabu: Modify to fit the above comments. This file should be around 100-150 lines.

automation/tests/setup.sh
```diff
@@ -0,0 +1,315 @@
...
+		if [[ `grep -ir "$i tests passed" $LOG_LOCATION/jans-cli_test.log` ]]
+		then
+			echo "$i Passed" >>./report/result.txt
+		else
+			echo "$i Failed" >>./report/result.txt
+
+		fi
+	done
+
+os_name=`uname -v | awk {'print$1'} | cut -f2 -d'-'`
```
moabu, line 64: 1. don't use html
2. Use markdown but don't write it here. Instead create a template markdown file in the directory. Call it and replace the temp values with the results. That should make it readable and easily updated. You would only be modifying the results instead of updating the whole doc in bash.

automation/tests/setup.sh
```diff
@@ -0,0 +1,315 @@
...
+
+
+			Red)
+				echo "Login to RHEL8 Host"
+				
+				 if ( ssh  ${USERNAME}@${HOST} 'ls /opt/jans/jans-cli/config-cli.py' )
+                                        then
+
+expect<<EOF
+spawn /usr/bin/ssh $USERNAME@$HOST
```
moabu: You can group the setup/uninstall in functions and call that.

moabu (changes requested):

manojs1978:
thanks for reviewing,
i ll modify these scripts

moabu:
@manojs1978 I think its more relevant and easier to use the swagger spec to generate the different possibile commands. I think this can later be converted to run from within the jans-cli itself with an options such as `--run-functional-tests` which would basically output and run all possible jans cli operations against a Jans AS.

automation/tests/install.exp
manojs1978: i think there is option to set branch while installation.

usage: install.py [-h] [-use-downloaded] [-upgrade] [-uninstall] [--args ARGS] [-yes] [--keep-downloads] [--profile {jans,openbanking}]  [-download-exit] [--setup-branch SETUP_BRANCH]

automation/tests/report/result.txt
manojs1978: this is done

automation/tests/install.exp
manojs1978: this is done

automation/tests/install.exp
manojs1978: this is done

automation/tests/setup.sh
manojs1978: this is done used  markdown template and updated result

automation/tests/setup.sh
manojs1978: modified script and file contains now 250-260 lines

automation/tests/setup.sh
manojs1978: this is done,grouped in setup file

Issue #1593, opened 2022-06-23
fix (jans-config-api) : error in POST, PUT endpoint of user management plugin if `birthdate` attribute is present in request


duttarnab:
**Describe the bug**
The user management plugin has apis to CRUD users. The POST, PUT endpoint of this plugin throws parsing error if the `birthdate` attribute is present in the request. The issue is due to the date parsing error at the backend.  

The LDAP backend supports date only in `yyyyMMddHHmmss.SSSZ` format but MySQL backend support only `yyyy-MM-ddTHH:mm:ss.SSSZ` format. So user management plugin should persist/update `birthdate` as a Date object (not String) to prevent this error.

duttarnab:
Done

PR #1594, opened 2022-06-23
fix: error in POST, PUT endpoint of user management plugin if birthda…

duttarnab:
https://github.com/JanssenProject/jans/issues/1593


duttarnab:
Need to add few more changes.

Issue #1532, opened 2022-06-09
feat(jans-client-api): use injectable operations and remove ServiceProvider class


yuriyz:
### Description

feat(jans-client-api): use injectable operations and remove ServiceProvider class

We should have `@RequestScoped` operations which can serve request directly without `ServiceProvider`, `Processor` wrappers.

`Authorization` header has to be processed in Filter or in Pre-Procesor unless we really need authorization inside operation.

```
@RequestScoped
@Named
public class GetAccessTokenByRefreshTokenOperation {
   ...
}

...
public class RestResource {
   
    @Inject
    GetAccessTokenByRefreshTokenOperation getAccessTokenByRefreshTokenOp;

    @POST
    @Path("/get-access-token-by-refresh-token")
    @Produces(MediaType.APPLICATION_JSON)
    @Consumes(MediaType.APPLICATION_JSON)
    public Response getAccessTokenByRefreshToken(@HeaderParam("Authorization") String authorization, @HeaderParam("AuthorizationRpId") String authorizationRpId, String params) {
        logger.info("Api Resource: /get-access-token-by-refresh-token  Params: {}", params);
        return getAccessTokenByRefreshTokenOp.process(params, authorizationRpId);
    }
}
```

PR #1643, opened 2022-06-30
feat(jans-client-api): Use injectable operations and remove serviceprovider

jmunozherbas:
### Prepare

-------------------

### Description
use injectable operations and remove ServiceProvider class

#### Target issue
https://github.com/JanssenProject/jans/issues/1532

#### Implementation Details
Refactoring code
Operations are injected directly in request

-------------------


yuriyz (changes requested):
Lets keep authorization check in Filter.

jans-client-api/server/src/main/java/io/jans/ca/server/filter/AuthorizationFilter.java
```diff
@@ -0,0 +1,82 @@
...
+
+        log.info("======" + context.getMethod() + " " + info.getPath() + " FROM IP " + request.getRemoteAddr());
+        log.info("======PERFORMING AUTHORIZATION=========================================");
+        String authorizationHeader = context.getHeaderString(HttpHeaders.AUTHORIZATION);
+        String authorizationRpIdHeader = context.getHeaderString(AUTHORIZATION_RP_ID);
+
+        log.info("\n\n\n AuthorizationFilter::filter() - authorizationHeader = " + authorizationHeader + " , authorizationRpIdHeader = "
+                + authorizationRpIdHeader + " \n\n\n");
+        try {
+            authorizationService.processAuthorization(info.getPath(), context.getMethod(), request.getRemoteAddr(), authorizationHeader, authorizationRpIdHeader);
```
yuriyz, line 69: If we have filter which performs authorization why do we still authorize again here `io.jans.ca.server.op.BaseOperation#getObjectForJsonConversion` ? 
```
 if (getCommandType().isAuthorizationRequired()) {
            final ApiAppConfiguration conf = jansConfigurationService.find();
            String authorization = httpRequest.getHeader("Authorization");
            String authorizationRpId = httpRequest.getHeader("AuthorizationRpId");
            validateAccessToken(authorization, safeToRpId((HasRpIdParams) params, authorizationRpId), conf);
        }
```

jans-client-api/server/src/main/java/io/jans/ca/server/op/AuthorizationCodeFlowOperation.java
```diff
@@ -54,6 +47,16 @@ public IOpResponse execute(AuthorizationCodeFlowParams params) {
         return null;
     }
 
+    @Override
+    public Class<AuthorizationCodeFlowParams> getParameterClass() {
+        return AuthorizationCodeFlowParams.class;
+    }
+
+    @Override
+    public CommandType getCommandType() {
```
yuriyz: Since now it's injectable operation we don't need commandType because we should not have central processing. Isn't it? Remove `CommandType` and all info should go to operation (e.g. return type).

yuriyz (changes requested):

jans-client-api/server/src/main/java/io/jans/ca/server/filter/AuthorizationFilter.java
jmunozherbas: In case `AuthorizationRpId` header is null, then it is read from body params:
https://github.com/JanssenProject/jans/blob/e44ed18da41f7621489cbeabb47491341de8a82c/jans-client-api/server/src/main/java/io/jans/ca/server/op/BaseOperation.java#L84

I tried to read the body content in filter, but then the body is not available for `io.jans.ca.server.op.BaseOperation#process` method.

`io.jans.ca.server.filter.AuthorizationFilter` process the token validation only when `AuthorizationRpId` header is not null.

One option could be to establish `AuthorizationRpId` header as mandatory.
Otherwise, we can look for a way to read body in filter, without losing it.

jans-client-api/server/src/main/java/io/jans/ca/server/op/AuthorizationCodeFlowOperation.java
jmunozherbas: OK I will remove.

jans-client-api/server/src/main/java/io/jans/ca/server/filter/AuthorizationFilter.java
yuriyz: Right, lets read `authorizationRpId` from header and move it out of body. In general everything was in body because oxd was designed as backchannel app, we should re-think it since now it's web app.

jmunozherbas: Now, `authorizationRpId` is mandatory in header.

jmunozherbas: It is ready for review

yuriyz (approved):

Issue #1575, opened 2022-06-17
feat (jans-config-api):  add support for date ranges in statistic endpoint


duttarnab:
**Is your feature request related to a problem? Please describe.**
Currently stat endpoint support explicit listing of months, e.g. month=202202 202203 which is inconvenient for a long date range (eg: from 202102 to 202202 ). 

**Describe the solution you'd like**
Idea is to add range support, `e.g. start-month=202102, end-month=202202` has to include all months in between.

Ranges support is added to jans-auth statistic endpoint. Please refer https://github.com/JanssenProject/jans/issues/1089 for deatils.

duttarnab:
Done

PR #1653, opened 2022-07-02
feat: add support for date ranges in statistic endpoint #1575

duttarnab:
https://github.com/JanssenProject/jans/issues/1575

### DIFF
--- a/jans-auth-server/client/src/main/java/io/jans/as/client/service/StatService.java
+++ b/jans-auth-server/client/src/main/java/io/jans/as/client/service/StatService.java
@@ -16,12 +16,20 @@ import jakarta.ws.rs.core.MediaType;
 public interface StatService {
     @GET
     @Produces({MediaType.APPLICATION_JSON})
-    JsonNode stat(@HeaderParam("Authorization") String authorization, @QueryParam("month") String month, @QueryParam("format") String format);
+    JsonNode stat(@HeaderParam("Authorization") String authorization,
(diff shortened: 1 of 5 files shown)


jans-auth-server/client/src/main/java/io/jans/as/client/service/StatService.java
```diff
@@ -16,12 +16,20 @@
...
     @GET
     String statOpenMetrics(@HeaderParam("Authorization") String authorization, @QueryParam("month") String month, @QueryParam("format") String format);
 
     @POST
     @Produces({MediaType.APPLICATION_JSON})
-    JsonNode statPost(@HeaderParam("Authorization") String authorization, @FormParam("month") String month, @FormParam("format") String format);
+    JsonNode statPost(@HeaderParam("Authorization") String authorization,
+                      @FormParam("month") String month,
+                      @QueryParam("start-month") String startMonth,
+                      @QueryParam("end-month") String endMonth,
```
yuriyz: For POST it has to be FormParam, not QueryParam

yuriyz (changes requested):

jans-auth-server/client/src/main/java/io/jans/as/client/service/StatService.java
duttarnab: My bad. I will be more careful next time.

yuriyz (approved):

PR #1637, opened 2022-06-29
fix(forgot_password): update script according to jans specs

christian-hawk:
### Prepare

- [x] Read [PR guidelines](https://github.com/JanssenProject/jans/blob/main/docs/CONTRIBUTING.md#prs)
- [x] Read [license information](https://github.com/JanssenProject/jans/blob/main/LICENSE)

-------------------

#### Target issue
ref issues #1601 #1631

#### Implementation Details
  

-------------------
### Test and Document the changes
- [ ] Static code analysis has been run locally and issues have been fixed
- [ ] Relevant unit and integration tests have been added/updated
- [ ] Relevant documentation has been updated if any (i.e. user guides, installation and configuration guides, technical design docs etc)

PR #1662, opened 2022-07-04
fix: jans-linux-setup disable script Forgot_Password_2FA_Token

devrimyatar:
Script **Forgot_Password_2FA_Token** is not working properly. Disabled until it is fixed.
It can be enabled after installation by `python3 /opt/jans/jans-setup/setup.py -enable-script=B270-381E` for testing.


moabu (approved):
disabling this as its breaking the installation. We will enable it once [pr #1637 ](https://github.com/JanssenProject/jans/pull/1637) is resolved.

Issue #207, opened 2021-02-18
Cache should do entry update instead of remove/insert when DB is SQL


yurem:
We need to add new option to control cache entry update. When this option is enabled it should do entry update instead of remove/insert new entry. This can be useful for SQL in ordee to avoid DB extra grow up.

PR #1787, opened 2022-07-13
FEAT / Added config to disable attempt to update before insert in cache

Milton-Ch:
### Description

Added config to disable attempt to update before insert in cache.

#### Target issue
https://github.com/JanssenProject/jans/issues/207

### DIFF
--- a/jans-core/cache/src/main/java/io/jans/service/cache/NativePersistenceCacheProvider.java
+++ b/jans-core/cache/src/main/java/io/jans/service/cache/NativePersistenceCacheProvider.java
@@ -82,6 +82,9 @@ public class NativePersistenceCacheProvider extends AbstractCacheProvider<Persis
             // CouchbaseEntryManagerFactory.PERSISTENCE_TYPE
             skipRemoveBeforePut = "couchbase".equals(persistenceType);
             attemptUpdateBeforeInsert = "sql".equals(persistenceType);
+            if (cacheConfiguration.getNativePersistenceConfiguration().isDisableAttemptUpdateBeforeInsert()) {
+                attemptUpdateBeforeInsert = false;
+            }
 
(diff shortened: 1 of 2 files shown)


jans-core/cache/src/main/java/io/jans/service/cache/NativePersistenceCacheProvider.java
```diff
@@ -82,6 +82,9 @@ public void create() {
             // CouchbaseEntryManagerFactory.PERSISTENCE_TYPE
             skipRemoveBeforePut = "couchbase".equals(persistenceType);
             attemptUpdateBeforeInsert = "sql".equals(persistenceType);
+            if (cacheConfiguration.getNativePersistenceConfiguration().isDisableAttemptUpdateBeforeInsert()) {
```
yuriyz, line 85: I guess we don't need explicit configuration for it. Existing `attemptUpdateBeforeInsert = "sql".equals(persistenceType);` is enough. With this we already updating entry (for all further calls except initial one).

@yurem is it what you meant in https://github.com/JanssenProject/jans/issues/207 ?

Milton-Ch: Actually, I talked with him before doing these changes, but let's see Yure's feedback in case I misunderstood something.

yurem: yes, for #207 we already updated code. But this new property will allow us to check at runtime which approach is better for specific DB MySQL/Postgress/etc

yuriyz: Agreed, yes, it can be useful.

PR #1801, opened 2022-07-14
docs: updated introspection script methods

IdentityMethods-UwarajWaman:
Added details to 'Introspection' interception script with sample code snippet and a use case.


nynymike (changes requested):
Please review [these two scripts](https://github.com/GluuFederation/community-edition-setup/tree/master/static/extension/introspection).  The longer one shows how to work backwards from a grant to user information. Could we also get information about the original authorization request, like the `state`? 

I don't understand the use case for your script. Do people commonly post a Userinfo JWT when they are requesting or introspecting a token? `ujwt` is not a valid request parameter for [introspection](https://datatracker.ietf.org/doc/html/rfc7662#section-2.1). Does the introspection endpoint even allow custom parameters? Net-net, I don't think this is the best place to show to validate a JWT--we would have to do this in the Client Registration script for software statement validation. 

Another very common use case for the introspection script is calling an external backchannel API to add a dynamic claim to the access token. 

It might also be interesting to know what other sessions a user might have. For example, the session for an access token is associated with a Chrome browser. But the user might have another active session in Firefox. Could we figure out how to find other sessions for the owner of that access token?

nynymike (changes requested):
These changes don't align with my feedback. The overview section is lacking. That should be a clear English. You can normally find some good text from the RFC itself.

In the methods section, I think you should have a link to the object we are inheriting. For example, what are the parameters for `init`. 

Still waiting for the code samples too... especially the Java one.

IdentityMethods-UwarajWaman:
@nynymike I am still working on the changes you suggested. I set up a working Jans Auth Server locally and trying out different code snippets before adding to the documentation. Its bit of a learning curve with Jans AS CLI.

IdentityMethods-UwarajWaman:
@nynymike Updated the introspection script doc as per your suggestions. Let me know if we are heading in right direction. If this is good to proceed then I'll add few more use cases in both python and java.
@OliverSothcott FYI.

IdentityMethods-UwarajWaman:
@nynymike Updated the document as per last review comments and placed the document and code samples in respective locations as per new structure. Please review.

IdentityMethods-UwarajWaman:
@nynymike Updated the 'Methods' section as per last review comments

IdentityMethods-UwarajWaman:
@nynymike Added doc for OpenID-Config (Discovery) script along with both python and java code samples. Please review.

nynymike (changes requested):
Line 49: the params need a space after the comma (std python formatting).
This looks like a bad find/replace during the migration: https://github.com/JanssenFederation/oxauth 

Author:  Uwaraj Waman is not ok. This is work for hire. The author should be Gluu.

nynymike (changes requested):
* Remove "oxAuth is available under the MIT License (2008). See http://opensource.org/licenses/MIT for full text." 

* Copyright should be "Copyright (c) 2022, Janssen Project"

* Per other comment: Author: Gluu -- this is a work for hire.

IdentityMethods-UwarajWaman:
@nynymike updated the code samples as per review comments

shmorri (approved):

Issue #776, opened 2022-02-08
upgrade to MDS3 in fido2 server implementation


maduvena:
### Description

- Development target issue
 #242 

- Implementation Details
### New version of Janssen Server:

1. on startup of fido2 server we need to do the following

`wget https://mds.fidoalliance.org/ -O /etc/jans/conf/fido2/mds/toc/toc.jwt`

2. `wget http://secure.globalsign.com/cacert/root-r3.crt -O /etc/jans/conf/fido2/mds/cert/root-r3.crt`

3. This folder `/etc/jans/conf/fido2/mds/toc/toc.jwt`  should have appropriate permissions because there will be a Timer Task will update this file toc.jwt by downloading the latest copy from the same location

### For upgrade script:

1. Delete whatever is present in /etc/jans/conf/fido2/mds/toc/toc.jwt

2. Get the latest - `wget https://mds.fidoalliance.org/ -O /etc/jans/conf/fido2/mds/toc/toc.jwt`

3. `wget http://secure.globalsign.com/cacert/root-r3.crt -O /etc/jans/conf/fido2/mds/cert/root-r3.crt`

4. This folder `/etc/jans/conf/fido2/mds/toc/toc.jwt`  should have appropriate permissions because there will be a Timer Task will update this file toc.jwt by downloading the latest copy from the same location

-------------------

### Prepare

- [X ] Read contribution guidelines
- [ X] Read license information

------------------

### Identified code changes

- [ ] follow steps listed in implementation

-------------------

### Test cases and code coverage `NA`

- [ ] Write unit test to cover added/changed code
- [ ] Update integration tests to cover added/changed code

-------------------

### Document the changes `NA`

- [ ] #785
- [ ] task for updating installation and configuration guides if needed `NA`
- [ ] task for updating developer documentation if needed `NA`
- [ ] task for updating technical documentation if needed `NA`

moabu:
This needs implementation in CN .

Cc @iromli

PR #2503, opened 2022-09-29
fix: #776

maduvena:
#776


jans-auth-server/server/src/main/java/io/jans/as/server/service/MDSUpdateTimer.java
```diff
@@ -0,0 +1,67 @@
...
+import io.jans.service.timer.schedule.TimerSchedule;
+import org.slf4j.Logger;
+
+/**
+ * @author madhumitas
+ *
+ */
+@ApplicationScoped
+@Named
+public class MDS3UpdateTimer {
```
yuriyz: It's not clear why timer with event is put in jans-auth-server. Shouldn't it be in fido2-server? fido2-server is separate war and can carry it.

jans-auth-server/server/src/main/java/io/jans/as/server/service/MDSUpdateTimer.java
```diff
@@ -0,0 +1,67 @@
...
+	@Inject
+	private Logger log;
+
+	@Inject
+	private Event<TimerEvent> timerEvent;
+
+	@Inject
+	private TocService tocService;
+
+	public void initTimer() {
```
yuriyz: Formatting is messed. (Check other java files, e.g. spaces (vs tabs), curly brackets (starting bracket is always in same line))

yuriyz (changes requested):

jans-auth-server/server/src/main/java/io/jans/as/server/service/MDSUpdateTimer.java
maduvena: This was in error, corrected

yuriyz (changes requested):
fido2-server app initializer has to be used instead of jans-auth-server

yuriyz (approved):
Approved since changes are moved out of AS. Now we should let @yurem check fido related changes.

yurem (approved):

PR #2581, opened 2022-10-10
feat(jans-core): port Gluu ORM-based document store

shekhar16:
feat(jans-core): port Gluu ORM-based document store

### DIFF
--- a/jans-core/document-store/src/main/java/io/jans/service/document/store/provider/DBDocumentStoreProvider.java
+++ b/jans-core/document-store/src/main/java/io/jans/service/document/store/provider/DBDocumentStoreProvider.java
@@ -17,7 +17,7 @@ import io.jans.service.document.store.conf.DBDocumentStoreConfiguration;
 import io.jans.service.document.store.conf.DocumentStoreConfiguration;
 import io.jans.service.document.store.conf.DocumentStoreType;
 import io.jans.service.document.store.service.DBDocumentService;
-import io.jans.service.document.store.service.OxDocument;
+import io.jans.service.document.store.service.Document;
 import io.jans.util.StringHelper;
 import jakarta.annotation.PostConstruct;
 import jakarta.enterprise.context.ApplicationScoped;
@@ -73,9 +73,9 @@ public class DBDocumentStoreProvider extends DocumentStoreProvider<DBDocumentSto
 		if (StringHelper.isEmpty(DisplayName)) {
 			throw new IllegalArgumentException("Specified path should not be empty!");
 		}		
-		OxDocument oxDocument = null;
+		Document oxDocument = null;
 		try {
-			oxDocument = documentService.getOxDocumentByDisplayName(DisplayName);
+			oxDocument = documentService.getDocumentByDisplayName(DisplayName);
 		} catch (Exception e) {
 			log.error("Failed to check if path '" + DisplayName + "' exists in repository", e);
 		}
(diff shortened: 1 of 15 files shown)


jans-core/document-store/src/main/java/io/jans/service/document/store/provider/DBDocumentStoreProvider.java
```diff
@@ -0,0 +1,213 @@
...
+			}
+			oxDocument.setDisplayName(destinationDisplayName);
+			documentService.updateOxDocument(oxDocument);
+			OxDocument oxDocumentDestination = documentService.getOxDocumentByDisplayName(destinationDisplayName);
+			if(oxDocumentDestination == null) {
+				log.error("Failed to rename to destination file '{}'", destinationDisplayName);
+				return false;
+			}
+		} catch (Exception e) {
+			log.error("Failed to rename to destination file '{}'", destinationDisplayName);
```
yuriyz, line 183: Something is wrong here. In case of exception, we got further in line 187 and return `true` even if it's not renamed.
Also we should log exception for easier troubleshooting.

jans-core/document-store/src/main/java/io/jans/service/document/store/provider/DBDocumentStoreProvider.java
```diff
@@ -0,0 +1,213 @@
...
+			String inum = documentService.generateInumForNewOxDocument();
+			oxDocument.setInum(inum);	
+			String dn = "inum="+ oxDocument.getInum() +",ou=document,o=gluu";
+			oxDocument.setDn(dn);
+			oxDocument.setDescription(name);
+			oxDocument.setOxEnabled("true");
+			oxDocument.setOxModuleProperty(moduleList);
+			documentService.addOxDocument(oxDocument);
+			return true;
+		} catch (IOException e) {
```
yuriyz: IOException is redundant because it has exactly same statements as super `Exception`.

jans-core/document-store/src/main/java/io/jans/service/document/store/service/OxDocument.java
```diff
@@ -0,0 +1,150 @@
...
+ * oxDocument
+ * 
+ * @author Shekhar L. Date: 01.10.2022
+ */
+
+
+@DataEntry(sortBy = { "displayName" })
+@ObjectClass(value = "oxDocument")
+@JsonInclude(Include.NON_NULL)
+public class OxDocument extends Entry implements Serializable {
```
yuriyz: In jans we removed ox prefix. It's replace with jans prefix instead.
Lets rename `OxDocument` with just `Document`. It's inside `document.store.service` package so it's clear to what functionality it belongs. Object Class can be renamed `oxDocument` -> `jansDocument`. Please add changes in `jans_schema.json`. It seems to be missed in this PR.

jans-core/document-store/src/main/java/io/jans/service/document/store/service/OxDocument.java
```diff
@@ -0,0 +1,150 @@
...
+	private String document;
+	
+	@AttributeName
+	private Date creationDate;
+	
+	@AttributeName
+	private List<String> oxModuleProperty;
+
+	@AttributeName
+	private String oxLevel;
```
yuriyz: This attribute is not present in jans. As well as other attributes refixed here as `ox*`. It must be changed. Please check existing schema in `jans_schema.json`. Lets re-use existing attributes.

yuriyz (changes requested):

jans-linux-setup/jans_setup/schema/jans_schema.json
```diff
@@ -4286,6 +4297,33 @@
...
+		"jansEnabled",
+		"jansAlias"				
+      ],
+      "must": [
+        "objectclass"
+      ],
+      "names": [
+        "jansDocument"
+      ],
+      "oid": "oxObjectClass",
```
yuriyz: jansObjClass instead of oxObjectClass?

jans-linux-setup/jans_setup/schema/jans_schema.json
```diff
@@ -4286,6 +4297,33 @@
...
+        "ou",
+        "displayName",
+		"description",
+		"document",
+		"creationDate",
+		"jansModuleProperty",
+		"jansLevel",
+		"jansRevision",
+		"jansEnabled",
+		"jansAlias"
```
yuriyz: It would be great to use spaces instead of tabs all over this PR. Otherwise you can see how github represents changes in weird way.

yuriyz (changes requested):

yuriyz (approved):

Issue #2487, opened 2022-09-28
Script-catalog refactoring


maduvena:
1. All interception scripts will be placed under https://github.com/JanssenProject/jans/tree/main/docs/script-catalog under a folder representing the **Type** of the script (for e.g. [UpdateTokenType](https://github.com/JanssenProject/jans/blob/main/jans-core/script/src/main/java/io/jans/model/custom/script/type/token/UpdateTokenType.java),  [ClientRegistrationType](https://github.com/JanssenProject/jans/blob/main/jans-core/script/src/main/java/io/jans/model/custom/script/type/client/ClientRegistrationType.java), [PersonAuthenticationType](https://github.com/JanssenProject/jans/blob/main/jans-core/script/src/main/java/io/jans/model/custom/script/type/auth/PersonAuthenticationType.java) etc)

2. All [PersonAuthenticationType](https://github.com/JanssenProject/jans/blob/main/jans-core/script/src/main/java/io/jans/model/custom/script/type/auth/PersonAuthenticationType.java) scripts will be under https://github.com/JanssenProject/jans/tree/main/docs/script-catalog/person_authentication

3. Similarly other [CustomScriptTypes](https://github.com/JanssenProject/jans/blob/main/jans-core/script/src/main/java/io/jans/model/custom/script/CustomScriptType.java) will be under their respective folders.

This will be the **central location** for all scripts and the use of `integrations` folder is deprecated

Subsequently, all setups will pick the script from the catalog 
1. The CE setup folder will not have a duplicated version of this file.
2. Neither will there be duplicates on CN.

TODOs:
- [ ] Modify the CE setup to pick scripts from the catalog

- [ ] Modify the CN setup to pick scripts from the catalog

moabu:
Lets push this through.  We need to keep moving forward to lower the duplicates in the repo

CC @iromli

maduvena:
You only need to edit file `jans-linux-setup/jans_setup/templates/scripts.ldif` after moving the script.
Rule for template name
1. directories are merged with underscore and dashes are replaced by underscore, e.g., directory-1/another-directory_2 becomes directory_1_another_directory_2
2. append lowered file name without extension with underscote to directory., e.g., file directory-1/another-directory_2/SampleScript.py becomes directory_1_another_directory_2_samplescript

PR #2975, opened 2022-11-14
fix: #2487 - removing inwebo

maduvena:
#2487


yuriyz (changes requested):
Shouldn't it be moved to script-catalog? PR contains only remove. Or we don't need inwebo anymore?

https://github.com/JanssenProject/jans/tree/main/docs/script-catalog

devrimyatar (approved):

maduvena:
@yuriyz - it hasnt been tested in years and we dont have a test account. It is not being used at all.

yuriyz:
I see, then approving.

yuriyz (approved):

PR #2999, opened 2022-11-15
fix: #2487 scripts-catalog folder restructuring

maduvena:
#2487


devrimyatar (changes requested):
Pleaceholders in `scripts.ldif` are not correct. Please use these rules

1. seperate directories with `_`
2. replace `-` with `_` in directory names
3. use only script name, omit extension and use lowercase

Using these rules placeholder for script:
`person_authentication/user-cert-external-authenticator/UserCertExternalAuthenticator.py`
is 
`person_authentication_user_cert_external_authenticator_usercertexternalauthenticator`

When I try to install from this branch, setup exited with error:
```
  File "/opt/jans/jans-setup/setup_app/utils/setup_utils.py", line 492, in render_template
    rendered_text = self.fomatWithDict(template_text, format_dict)
  File "/opt/jans/jans-setup/setup_app/utils/setup_utils.py", line 385, in fomatWithDict
    return text % dictionary
KeyError: 'person_authentication_usercertexternalauthenticator'
```

devrimyatar (approved):

yuriyz (approved):

PR #3328, opened 2022-12-15
ci: introduce release number marker for documentation

### LINKED ISSUE #3327: docs: automation required for updating current release references
Currently, all the references to releases in the documentation are either pointing to `head` or they are hard coded. This does not allow content to be release specific. 

For example, in the documentation for any release before 1.0.5 and the URLs for swagger API reference only point to swagger spec at `head` release. URL documented in v1.0.4 doesn't point to the spec of the same release but instead, it points to `head`. 

The problem can be fixed by manually updating all such places with corresponding release numbers but this would not be a practical approach in the long run when there will be many such touch points. It'll be error-prone too. 

Automation is needed to make sure this is a systematic way of referring to current release and updating the number correctly.

ossdhaval:
### Prepare

-------------------

#### Target issue
  
closes #3327 

#### Implementation Details
Added a release marker as documented in CONTRIBUTION.md

CI flow will replace the marker with the appropriate release number.

-------------------

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -86,15 +86,27 @@ jobs:
 
       - name: Generate docs
         continue-on-error: true
+        env:
+          VERSION: ${{ github.event.release.tag_name }}
         run: |
          echo "Custom work on generating docs can go here."
+          
          # Run cn docs 
          sudo bash ./automation/docs/generated-cn-docs.sh
+          
          # Generate property docs and push to main
          sudo bash ./automation/docs/generate-property-docs.sh 
          cd docs
          git add . || echo "Nothing to add"
          git commit -a -S -m "docs: automated property doc generation" || echo "Nothing to commit"
+         git push
+         cd ..
+          
+         # Replace release number markers with actual release number
+         cd docs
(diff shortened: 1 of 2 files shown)


.github/workflows/build-docs.yml
```diff
@@ -86,17 +86,29 @@ jobs:
...
          # Generate property docs and push to main
          sudo bash ./automation/docs/generate-property-docs.sh 
          cd docs
          git add . || echo "Nothing to add"
          git commit -a -S -m "docs: automated property doc generation" || echo "Nothing to commit"
          git push
          cd ..
+          
+         # Replace release number markers with actual release number
+         cd docs
```
moabu: add a condition here. If the VERSION is empty then you shouldn't run the substitution because this will cause it to be replaced by an empty string which then becomes difficult to fix.

moabu (changes requested):

.github/workflows/build-docs.yml
ossdhaval: Added condition check.

moabu (approved):

Issue #3088, opened 2022-11-28
"One step" authentication flow using Super Gluu


maduvena:
![image](https://user-images.githubusercontent.com/12072533/207381064-85008dfb-0b53-4bc2-8abe-2ad9947a6d66.png)

Flow is as follows:
1. user visits login page, which displays the QR code
2. User scans the QR code using SG app and approves it
3. User is prompted a login-pwd page so that a user can be attached to this credential
4. Subsequent authentications takes place in one step , where the user scans a QR code and approves the same on the mobile app (find user by it’s keyHandle and do user login)

**In this flow we store user device enrollment in temporary record in DB/enroll user/associate enrolled device with newly created user**

maduvena:
![image](https://user-images.githubusercontent.com/12072533/207635881-74e021d9-7c02-4e0c-b97f-9df7b47d779c.png)
Incase of one_step, this page should directly show the QR code to be scanned

PR #3337, opened 2022-12-16
fix(jans-auth): Issue 3088 - "One step" authentication flow using Super Gluu

maduvena:
#3088


yuriyz (changes requested):
In jans fido is seperate project. Thus I vote to keep these changes in https://github.com/JanssenProject/jans/tree/main/jans-fido2

yurem:
> In jans fido is seperate project. Thus I vote to keep these changes in https://github.com/JanssenProject/jans/tree/main/jans-fido2

I agree with this. Also I wonder if we really need this old code in Jans. It's our first version which support U2F (v1) only. Now Fido2 server supports both U2F and Fido2. @maduvena I think we need to apply our custom U2F parts to Fido2 server code. Our long target is to move Super Gluu to use Fido2 server. This update will be steep in this direction.

nynymike:
We want to remove FIDO U2F endpoints from Auth Server... the idea was to move U2F to the FIDO server...

yuriyz (approved):
For jans-auth-server looks good. We need Yura to review fido2 changes part.

PR #3411, opened 2022-12-24
fix: jans-linux-setup opPolicyUri and opTosUri

### LINKED ISSUE #3410: fix: jans-linux-setup opPolicyUri and opTosUri
**Describe the bug**
In jans-auth config we have:
```
  "opPolicyUri": "http://www.jans.io/doku.php?id=jans:policy",
  "opTosUri": "http://www.jans.io/doku.php?id=jans:tos",
We need to update these default values… these URLs refer to (very) old DocuWiki
```
Ideally, we should make these https://%s/opPolicy % op_domain_name

and https://%s/tos % op_domain_name

devrimyatar:
closes #3410


iromli:
@devrimyatar no changes for openbanking config? It has `opPolicyUri` too.

yuriyz (approved):

Issue #3717, opened 2023-01-30
fix(jans-auth): Runtime issue in config-api 


maduvena:
Refactoring causing issues in `jans-config-api`
1. RegistrationPersistenceService has been recently added for the the sake of SG script in jans-auth common
2. It caused a side effect in config-api. Config-api dosnt start up
```
2023-01-30 07:34:01.816:WARN :oejw.WebAppContext:main: Failed startup of context o.e.j.w.WebAppContext@45a37759{jans-config-api.xml,/jans-config-api,[file:///opt/jetty-11.0/temp/jetty-localhost-8074-jans-config-api_war-_jans-config-api-any-7755429429673189795/webapp/, jar:file:///opt/jetty-11.0/temp/jetty-localhost-8074-jans-config-api_war-_jans-config-api-any-7755429429673189795/webapp/WEB-INF/lib/jakarta.faces-3.0.2.jar!/META-INF/resources, jar:file:///opt/jans/jetty/jans-config-api/custom/libs/scim-plugin.jar!/META-INF/resources],UNAVAILABLE}{/opt/jans/jetty/jans-config-api/webapps/jans-config-api.war}
org.jboss.weld.exceptions.DeploymentException: WELD-001409: Ambiguous dependencies for type UserService with qualifiers @Default|  at injection point [BackedAnnotatedField] @Inject private io.jans.as.common.service.common.fido2.RegistrationPersistenceService.userService|  at io.jans.as.common.service.common.fido2.RegistrationPersistenceService.userService(RegistrationPersistenceService.java:0)|  Possible dependencies: |  - Managed Bean [class io.jans.configapi.plugin.mgt.service.UserMgmtService] with qualifiers [@Default @Any @Named],|  - Managed Bean [class io.jans.configapi.plugin.fido2.service.UserFido2Service] with qualifiers [@Default @Named @Any]|
        at org.jboss.weld.bootstrap.Validator.validateInjectionPointForDeploymentProblems(Validator.java:386)
        at org.jboss.weld.bootstrap.Validator.validateInjectionPoint(Validator.java:290)
        at org.jboss.weld.bootstrap.Validator.validateGeneralBean(Validator.java:143)
        at org.jboss.weld.bootstrap.Validator.validateRIBean(Validator.java:164)
        at org.jboss.weld.bootstrap.Validator.validateBean(Validator.java:526)
        at org.jboss.weld.bootstrap.ConcurrentValidator$1.doWork(ConcurrentValidator.java:64)
[...]

PR #3725, opened 2023-01-30
fix(jans-config-api): runtime exceptions in config-api at startup

maduvena:
#3717


pujavs (approved):

jans-config-api/plugins/fido2-plugin/src/main/java/io/jans/configapi/plugin/fido2/service/RegistrationPersistenceService.java
```diff
@@ -0,0 +1,29 @@
...
+@ApplicationScoped
+public class RegistrationPersistenceService extends io.jans.as.common.service.common.fido2.RegistrationPersistenceService {
+
+    public String getUserInum(String userName)
+    {
+    	return userService.getUserInum(userName);
+    }
+
+    @Inject
+	private  UserService userService;
```
yurem: Is it works well without `@Named("userFido2Srv")` ?

yurem: It's not clear because we use the same `io.jans.as.common.service.common.UserService` in old `RegistrationPersistenceService`  and there

yurem (approved):

Issue #3812, opened 2023-02-08
fix(jans-fido2): decoding authData incase of SuperGluu


maduvena:
U2F V1.2 spec needs to be checked to verify whether to use base64 or base64-URL to decode AuthData

Current error : 
```

7-02 13:38:44.645 DEBUG [qtp6519275-17] [jans.fido2.ws.rs.controller.AssertionController] (AssertionController.java:122) - Prepared U2F_V2 authentication options request: {"authenticateRequests":[{"type":"public-key","transports":["net","qr"],"appId":"https://maduvena-rational-jennet.gluu.info/.well-known/openid-configuration","userVerification":"preferred","challenge":"T6pUdkYhz_AGMxuDak6BWcR-I-lPkBF_lmtpccAfa60","keyHandle":"zq04MC8_3itYKRPO71L0ouoEGRKET-mO-r6IgooH3ZOVu2_n7w9mY4ykE51baDCQYmqObCmkXDXxsdNPNNs_WA","version":"U2F_V2"}]}
07-02 13:38:45.060 DEBUG [qtp6519275-16] [jans.fido2.ws.rs.controller.AssertionController] (AssertionController.java:136) - Finish authentication: username = madhu1, tokenResponse = {"signatureData":"AQAAAAEwRgIhANCfVml4Q31qQSxRWufPRYf5rKezPwWK01zwbzsZT0YmAiEA5_yZua9HaTfKwj1-e4Gv-aQhe8vmgRBDqcZ97YEygvg","clientData":"eyJ0eXAiOiJuYXZpZ2F0b3IuaWQuZ2V0QXNzZXJ0aW9uIiwiY2hhbGxlbmdlIjoiVDZwVWRrWWh6X0FHTXh1RGFrNkJXY1ItSS1sUGtCRl9sbXRwY2NBZmE2MCIsIm9yaWdpbiI6Imh0dHBzOlwvXC9tYWR1dmVuYS1yYXRpb25hbC1qZW5uZXQuZ2x1dS5pbmZvXC8ud2VsbC1rbm93blwvb3BlbmlkLWNvbmZpZ3VyYXRpb24ifQ","keyHandle":"zq04MC8_3itYKRPO71L0ouoEGRKET-mO-r6IgooH3ZOVu2_n7w9mY4ykE51baDCQYmqObCmkXDXxsdNPNNs_WA"}
[...]

PR #3814, opened 2023-02-08
fix(jans-fido2): For SG authenticatorData use base64 decoding instead…

maduvena:
… of base64URL decoding #3812


maduvena:
@yurem has added fix for this in https://github.com/JanssenProject/jans/pull/3815

Issue #4253, opened 2023-03-21
[Tent] use only META_URL and remove unneeded fields from config.py


christian-hawk:
`config.py` has fields like `CLIENT_AUTH_URI`, `TOKEN_URI`,  `USERINFO_URI` that are needed only when there is no `SERVER_META_URL`.

So as there should **always** be a configuration endpoint (`SERVER_META_URL`), those fields are unnecessary.

PR #4281, opened 2023-03-23
fix(tent): use params according to OpenID specs

### LINKED ISSUE #4254: [Tent] `ACR_VALUES` should be optional
**Describe the bug**
ACR_VALUES should be optional as OpenID Specs, but they are required by tent's code.

ACR_VALUES are being defined and loaded from `config.py` when calling `oauth.register`. So if there is no ACR_VALUES, it will break.   ACR_VALUES should be validated, and if ACR_VALUE is `None`, don't register any `ACR_VALUE`.

christian-hawk:
Required params should not be required (such as ACR_VALUES). 

Also removed everything that is already found at configuration endpoint

Closes #4254 #4253


nynymike (approved):
ok

PR #4342, opened 2023-03-28
fix: fix push config name

yuremm:


yurem (approved):

mo-auto:
Error: Hi @yuremm, You did not reference an open issue in your [PR](4342). I attempted to create an [issue](#4343) for you.    
 Please update that [issues'](#4343) title and body and make sure I correctly referenced it in the above PRs body.

nynymike (commented):
`https://%(hostname)s/identity/register` is also Gluu 4 specific. Maybe just remove `/identity` 

Also, is this `notification_service_mode`specific to Super Gluu? If so, should the value be `super-gluu` not `jans` ?

Issue #4396, opened 2023-03-30
fix: update UserService to correclty add user when DB is not LDAP


yurem:

yurem:
implemented

PR #4397, opened 2023-03-30
fix: update UserService to correclty add user when DB is not LDAP #4396

yuremm:


mo-auto:
Error: Hi @yuremm, You did not reference an open issue in your [PR](4397). I attempted to create an [issue](#4398) for you.    
 Please update that [issues'](#4398) title and body and make sure I correctly referenced it in the above PRs body.

PR #4417, opened 2023-03-31
feat(jans-auth-server): log httpresponse body configurated by httpLoggingResponseBodyContent #349

### LINKED ISSUE #349: When HTTP logging mode is enabled, Auth Server should also dump HTTP response body.
_aliaksander-samuseu commented on Jul 30, 2019_
At the moment, `httpLoggingEnabled` property in conjunction with `loggingLevel` makes oxAuth to dump complete HTTP request into log file on disk for easy review. That is a huge step forward in terms of troubleshooting and ease of providing support to customers that (in theory) may make usage of cumbersome complex tools like intercepting proxies and Apache's `mod_dumpio` not necessary in most cases.

The only downside to this currently is that this feature won't dump HTTP response's body along with the request - and this severely limiting its troubleshooting potential in many cases.

According to @yurem, such limitation was added with optimization in mind, to prevent log from consuming disk space rapidly; there also are some implementation difficulties to make that happen (but still may be possible).

Overall, I believe implementing this additional functionality may worth a try. To tackle the disk space issue at least, there could be an additional key in JSON configuration array which would allow enabling HTTP response body logging separately from the rest of the HTTP logging functionality.

### LINKED ISSUE #4418: fix: feat(jans-auth-server): log httpresponse body configurated by httpLoggingResponseBodyContent -autocreated
This issue was created because no open issues were referenced in your PR (#4417).

jmunozherbas:
### Prepare

-------------------

### Description
A new AppConfiguration param for logging response body.

#### Target issue
https://github.com/JanssenProject/jans/issues/349
  
closes #349 

#### Implementation Details
The response body is logged if param "httpLoggingResponseBodyContent" is true in jans-auth AppConfiguration.
(ConfDynamic)
If param not exist or is false, then response body is not logged.

-------------------
- [X] Relevant documentation has been updated if any (i.e. user guides, installation and configuration guides, technical design docs etc)

Closes #4418,


mo-auto:
Error: Hi @jmunozherbas, You did not reference an open issue in your [PR](4417). I attempted to create an [issue](#4418) for you.    
 Please update that [issues'](#4418) title and body and make sure I correctly referenced it in the above PRs body.

jans-auth-server/server/src/main/java/io/jans/as/server/audit/debug/ServletLoggingFilter.java
```diff
@@ -119,10 +122,17 @@ protected String getRequestDescription(RequestWrapper requestWrapper, Duration d
...
-    protected String getResponseDescription(ResponseWrapper responseWrapper) {
+
+    protected String getResponseDescription(HttpServletResponseWrapper responseWrapper) {
         HttpResponse httpResponse = new HttpResponse();
-        httpResponse.setStatus(responseWrapper.getStatus());
-        httpResponse.setHeaders(responseWrapper.getHeaders());
+        if (appConfiguration.getHttpLoggingResponseBodyContent()) {
+            httpResponse.setStatus(((ReadableResponseWrapper) responseWrapper).getStatus());
+            httpResponse.setHeaders(((ReadableResponseWrapper) responseWrapper).getHeaders());
+            httpResponse.setBody(((ReadableResponseWrapper) responseWrapper).readBodyValue());
```
yuriyz: @jmunozherbas would you please extensively test this change and confirm it works good ? Because this line sounds like we read from writer and close it. So down the chain filters will not be able to process it. As result it can cause not returing response out of the server. It's similar problem as reading input stream in request which can make it unavailable further.

See this https://stackoverflow.com/questions/10210645/http-servlet-request-lose-params-from-post-body-after-read-it-once

yuriyz (changes requested):
See inline comment

jans-auth-server/server/src/main/java/io/jans/as/server/audit/debug/ServletLoggingFilter.java
jmunozherbas: I have applied some changes.
I have tested with jans-auth-client heap of tests, and results are similar than before the change.
Clarification.- The changes and new flows just be applied if the parameter "httpLoggingResponseBodyContent" is enabled, if this parameter doesn´t exist or is false, then the flow is executed as usual.

The code to read body from response is based on next example:

search "Logging requests and responses" in next page:
https://spaces.at.internet2.edu/display/Grouper/Grouper+Web+Services

code repository:
https://github.com/Internet2/grouper/tree/master/grouper-ws/grouper-ws/src/grouper-ws/edu/internet2/middleware/grouper/ws/j2ee

maven repository of this code library:
https://mvnrepository.com/artifact/edu.internet2.middleware.grouper/grouper-ws/5.0.3

There is an example to read request body too, maybe could be implemented.
The log printting in just one line (request and response) could be implemented too, similar to the example code.

yuriyz: If all tests are passed with `httpLoggingResponseBodyContent=true` then I guess we are good to merge.

yuriyz:
Lets correct sonar reports before merge.

yuriyz (approved):

PR #5124, opened 2023-05-30
feat(jans-tarp): migrate tarp from vanilla javascript to react.js and typescript #5123

### LINKED ISSUE #5123: feat (jans-tarp): migrate tarp from vanilla javascript to react.js and typescript
- [x]  migrate tarp from vanilla javascript to react.js and typescript

duttarnab:
closes #5123


demos/jans-tarp/src/options/oidcClientDetails.tsx
```diff
@@ -0,0 +1,264 @@
...
+            })
+        });
+    }
+
+    async function triggerCodeFlowButton() {
+        setLoading(true);
+        const redirectUrl = chrome.identity.getRedirectURL()
+        const { secret, hashed } = await generateRandomChallengePair();
+        chrome.storage.local.get(["oidcClient"]).then(async (result) => {
+            if (result.oidcClient != undefined) {
```
jv18creator: please add optional chaining before accessing `oidcClient` value and we can change the condition to just `!result.oidcClient` which will handle where result may not exist instead of explicit value `undefined`

moabu (approved):

PR #5295, opened 2023-06-20
chore: fido2 server should allow enrollment if session is

yuremm:


mo-auto:
Error: Hi @yuremm, You did not reference an open issue in your [PR](5295). I attempted to create an [issue](#5296) for you.    
 Please update that [issues'](#5296) title and body and make sure I correctly referenced it in the above PRs body.

PR #6099, opened 2023-09-21
fix: remove content-type in header from /retrieve GET request #6096

### LINKED ISSUE #6096: fix(jans-config-api): remove content-type in header from /retrieve GET request
- remove content-type in the header from /retrieve GET request
https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET

- remove code duplication

duttarnab:
closes #6096


jans-config-api/plugins/admin-ui-plugin/src/main/java/io/jans/ca/plugin/adminui/service/license/LicenseDetailsService.java
```diff
@@ -508,4 +513,16 @@ private io.jans.as.client.TokenResponse generateToken(String opHost, String clie
             return null;
         }
     }
+
+    private Optional<GenericResponse> handleLicenseApiNotAccessible(Response response) {
+        if (response.getStatus() == 404) {
+            log.error("{}", LICENSE_APIS_404);
```
yuriyz, line 519: @duttarnab you don't need to involve formatter, it's as simple as `log.error(LICENSE_APIS_404);`

yuriyz (approved):

PR #6101, opened 2023-09-21
feat: adding folder for android and ios inside jans-chip #6100

### LINKED ISSUE #6100: feat(jans-chip): adding folder for android and ios inside jans-chip
- Adding folder for Android and ios projects inside jans-chip.
image

duttarnab:
closes #6100


NazarYavornytskyy (approved):
Good, integration of MVVM will in next stage

PR #6167, opened 2023-09-29
feat: include app integrity check through DCR #6166

### LINKED ISSUE #6166: feat(jans-chip): include app integrity check through DCR
1. User Google Play integrity API (https://developer.android.com/google/play/integrity/overview) to verify the integrity of app installed.
2. Pass the integrity result with attested DCR (in evidence parameter) so that it can be verified at Auth server.

duttarnab:
closes #6166

To mitigate the risk of app tampering and the use of fraudulent devices during client registration, it is highly advisable to integrate the native app with the Play Integrity API. The Play Integrity API helps you check that interactions and server requests are coming from your genuine app binary running on a genuine Android device.

The integrity verdict obtained through the Play Integrity API can be incorporated as a claim within the evidence parameter for attested DCR. This claim can then be subjected to verification by the designated verifier to guarantee the trustworthiness of both the app and the Android device. Based on this verification, the DCR may either be allowed or rejected accordingly.

Combining a classic request to check app integrity with the use of attestation in a DCR, we can create the following sequence diagram.

image


demos/jans-chip/android/app/src/main/java/io/jans/chip/RegisterActivity.java
```diff
@@ -51,28 +72,77 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            registerButton.setEnabled(false);
+        } else if (appIntegrity.getError() != null) {
+            appIntegrityText.setText(appIntegrity.getError());
+            registerButton.setEnabled(false);
+        } else if (appIntegrity.getResponse() != null) {
+            try {
+                JSONObject jo = new JSONObject(appIntegrity.getResponse().toString());
+                if (jo.has("appIntegrity")) {
+                    if(jo.getJSONObject("appIntegrity").has("appRecognitionVerdict")) {
+                        appIntegrityText.setText(jo.getJSONObject("appIntegrity").getString("appRecognitionVerdict"));
```
NazarYavornytskyy: I think better way there is to decode json into DAO structure and use class directly instead of using has("data_key")

demos/jans-chip/android/app/src/main/java/io/jans/chip/services/PlayIntegrityService.java
```diff
@@ -0,0 +1,214 @@
...
+
+                if (json.has("error")) {
+                    hasError = true;
+                    Log.e("Response from App server :: ", "Response body has error");
+                    //Toast.makeText(context, "Api request error :: " + json.getString("error"), Toast.LENGTH_SHORT).show();
+                    dbH.addAppIntegrity(setErrorInAppIntegrityObj("Response from App server has error :: " + json.getString("error")));
+                    return new String[]{"Api request error", json.getString("error")};
+                }
+
+                if (!json.has("appIntegrity")) {
```
NazarYavornytskyy: Please use decodable object there instead of has("data_key")

demos/jans-chip/android/app/src/main/java/io/jans/chip/RegisterActivity.java
duttarnab: Now using DTO to get appIntegrity from API.

demos/jans-chip/android/app/src/main/java/io/jans/chip/services/PlayIntegrityService.java
duttarnab: Now using DTO to get appIntegrity from API.

NazarYavornytskyy (approved):

Issue #6334, opened 2023-10-19
feat(jans-keycloak-link): implement password decryption


shekhar16:
Add functionality to check client secret and password is encrypted or not and handle both scenarios.

shekhar16:
code has been merged, closing this ticket. please let me know if anything needed.

PR #6335, opened 2023-10-19
feat(jans-keycloak-link): implement password decryption

shekhar16:
#6334 added code to handle encrypted and non encrypted password and client secret key.


jans-link/service/src/main/java/io/jans/link/util/PropertyUtil.java
yurem, line 1: We need to use io.jans.link.service.EncryptionService method decrypt.
PropertyUtil is mainly with static methods

jans-keycloak-link/server/src/main/java/io/jans/keycloak/link/server/service/KeycloakService.java
```diff
@@ -70,4 +85,27 @@ public List<UserRepresentation> getUserList() {
 
         return finalUsersResourceList;
     }
+
+    public Keycloak getKeycloakInstanceDecrypt() {
```
yurem: Is this method needed? There is no references to it in this PR

yurem (approved):

PR #6433, opened 2023-10-31
feat(jans-auth-server): enabled JWT response at introspection endpoint configured by AS and client config

### LINKED ISSUE #3658: feat(jans-auth-server): Enable JWT response at introspection point
```
   The introspection response, as specified in OAuth 2.0 Token
   Introspection [RFC7662], is a plain JSON object.  However, there are
   use cases where the resource server requires stronger assurance that
   the authorization server issued the token introspection response for
   an access token, including cases where the authorization server
   assumes liability for the content of the token introspection
   response.  An example is a resource server using verified person data
   to create certificates, which in turn are used to create qualified
   electronic signatures.

```

https://datatracker.ietf.org/doc/draft-ietf-oauth-jwt-introspection-response/12/

yuriyz:
### Description

feat(jans-auth-server): enabled JWT response at introspection endpoint configured by AS and client config

#### Target issue
  
closes #3658

### DIFF
--- a/jans-auth-server/client/src/main/java/io/jans/as/client/RegisterRequest.java
+++ b/jans-auth-server/client/src/main/java/io/jans/as/client/RegisterRequest.java
@@ -105,6 +105,9 @@ public class RegisterRequest extends BaseRequest {
     private SignatureAlgorithm userInfoSignedResponseAlg;
     private KeyEncryptionAlgorithm userInfoEncryptedResponseAlg;
     private BlockEncryptionAlgorithm userInfoEncryptedResponseEnc;
+    private SignatureAlgorithm introspectionSignedResponseAlg;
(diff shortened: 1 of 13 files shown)


ossdhaval:
Hi @yuriyz , should this change be mentioned on any of the documents ( I found relevant doc and section [here](https://docs.jans.io/v1.0.20/admin/auth-server/endpoints/introspection/#response-as-jwt)). 

Though I don't fully understand the changes but it seems to be adding additional capabilities. If so, it should find its place somewhere in the docs.

yuriyz:
Changes are : 
- publish introspection algs (used by jwt) on discovery page
- control algs via client. The same way as User Info and ID Token.

Discovery page is self-explanatory. And for client - swagger is updated. But I guess you are right, docs can mention it and say more.

PR #6769, opened 2023-11-28
feat(jans-auth-server): set feature flags state according to list discussed in #6611

### LINKED ISSUE #6611: fix: report correct state of feature flags on TUI
On a fresh install of Janssen Server(build: 1.0.21-nightly), feature flags are shown as below in TUI.

image

All checkboxes are unchecked. Giving an impression that all flags are disabled. At the same time, on AS all the flags are enabled by default. 

We need to fix this inconsistency. TUI should always show which flags are enabled/disabled on AS. There has to be a way of TUI to know which all flags are available on AS and out of those, which ones are enabled at that point in time.

Context: [discussion here](https://chat.gluu.org/group/service_jans_tui?msg=pjeJz6AFfKrD46uR8)

yuriyz:
### Description

feat(jans-auth-server): set feature flags state according to list discussed in #6611

#### Target issue
  
closes #6611


mo-auto:
Error: Hi @yuriyz, You did not reference an open issue in your [PR](6769). I attempted to create an [issue](#6774) for you.    
 Please update that [issues'](#6774) title and body and make sure I correctly referenced it in the above PRs body.

PR #7424, opened 2024-01-18
fix: remove extra end in nginx-ingress

### LINKED ISSUE #7423: fix(janssen-all-in-one): nginx-ingress deployment fails
`Error: UPGRADE FAILED: parse error at (janssen-all-in-one/templates/nginx-ingress.yaml:665): unexpected {{end}}`

misba7:
closes #7423


charts/janssen-all-in-one/templates/nginx-ingress.yaml
```diff
@@ -662,4 +662,3 @@ spec:
                 port:
                   number: 8080
 {{- end }}
-{{- end }}
```
iromli: This is the closing `end` of main `if` clause. It should not be deleted. Instead remove line 615. See https://github.com/JanssenProject/jans/commit/4ad8fc5b14f7ffcfac152d2f75de8e136cb0fa0a for comparison.

Note, this file can be reverted.

iromli (changes requested):

PR #7492, opened 2024-01-22
feat(jans-cli-tui): Support custom urls for Janssen services

davispuh:
### Prepare

-------------------

#### Target issue

I can host Janssen services on any `scheme://host:port/path` combination but `jans-cli` won't be able to interact with them.

#### Implementation Details

This PR implements `--auth_url`, `--config_url`, `--scim_url` parameters that solves this limitation.
This change is fully backwards compatible since `--host` parameter is still kept which has lower precedence.

For example:
```
$ jans-cli-tui --config_url http://config.internal:8080/config --auth_url http://auth.internal:8080/auth --scim_url http://scim.internal:8080/scim
```

-------------------


ossdhaval:
Hi @davispuh, Thanks for submitting this PR. The code owner will review it soon. At the same time, I wanted to request you to add DCO, guidelines [here](https://docs.jans.io/v1.0.22/CONTRIBUTING/#commits), to your commits.

ossdhaval:
Also, if this PR gets merged after the review, the documentation should be updated to reflect these additional switches.

davispuh:
Added DCO, but which documentation you mean? Where exactly it would need to be updated?

Note that `--help` is implemented and will output it correctly.

```
$ jans-cli --help
[...]
options:
  -h, --help            show this help message and exit
  --host HOST           Hostname of server
  --auth_url AUTH_URL   URL to Jans Auth server
  --config_url CONFIG_URL
                        URL to Jans Config API server
  --scim_url SCIM_URL   URL to Jans SCIM server
[...]
```

nynymike (commented):
These changes look good to me...

devrimyatar (commented):
@davispuh 

1. I did not see any benefit for setting custom janssen urls. We set hostname, that is all we need. Can you explain if there is some use cases?
2.  We can't make **host** as global variable at https://github.com/JanssenProject/jans/pull/7492/files#diff-8b4ac4f3964d1431edfcea09b382fa0e80c5353c86de9d278df5d0274e6eaae5R264 since JCA_CLI is initiated by TUI that gets host from configuration window. See https://github.com/JanssenProject/jans/blob/main/jans-cli-tui/cli_tui/jans_cli_tui.py#L323-L329

davispuh:
1. Using different domains/hosts and paths when hosting different services. I can host Config instance on `https://config.examle.org:14000/config` and SCIM on `https://scim.examle.org:17000/scim`. Currently CLI doesn't support using different port or path. Also this PR allows to configure server directly without going thru reverse proxy thus `http://localhost:8080/jans-config-api` works aswell.

2. That doesn't make it global variable, also it works correctly even for that part. It keeps same behavior as before when `--config_url` etc is not specified. `host` is first loaded from config file `~/.config/jans-cli.ini` and then used to create `endpoints` so it will work like before.

moabu (approved):
This works and I think can address certain limitation of the TUI atm. This helps if people have ephemeral envs and want to call the internal address of kuberentes services for each service.

moabu:
@davispuh I still think this is a valuable addition. Can you please rebase this  ? 


@devrimyatar can you please review after this PR has been rebased

PR #7555, opened 2024-01-25
feat(jans-linux-setup): unique indexes for uid and mail

### LINKED ISSUE #7419: feat: uid and email should have unique contains/index in SQL DBs by default
We need to add uniqueness constrains for `uid` and `email` to SQL DB: MySQL/PostgreSQL/Spanner.

For some DB is preferable to use indexes for this. For example (PostgreSQL):

```
CREATE UNIQUE INDEX ON "jansPerson" ("uid");
CREATE UNIQUE INDEX ON "jansPerson" ("mail");
```

The reason for this that we already add these indexes to DB. Also I think there similar answers [here](https://stackoverflow.com/questions/23542794/postgres-unique-constraint-vs-index).

devrimyatar:
closes #7419


yuriyz (approved):

yurem (approved):

mo-auto:
Error: Hi @devrimyatar, You did not reference an open issue in your [PR](7555). I attempted to create an [issue](#7558) for you.    
 Please update that [issues'](#7558) title and body and make sure I correctly referenced it in the above PRs body.

PR #7647, opened 2024-02-05
fix: for authorization with multiple redirects the one with redirect_url should be considered #7646

### LINKED ISSUE #7646: fix(jans-tarp): for authorization with multiple redirects the one with redirect_url should be considered
**Describe the bug**
There are cases where after authorization there are multiple redirects. In that case, the tarp should only consider the URL with redirect URL for fetching the authorization `code`.

duttarnab:
closes #7646


nynymike (approved):
Bueno!

moabu (approved):

PR #7670, opened 2024-02-07
fix: fix array java.lang.IndexOutOfBoundsException

### LINKED ISSUE #7446: feat(jans-config-api): user is able to see the user claim on create/edit user form even if the claim is disabled
1. The disabled user claim should not be obtained using `/mgt/configuser` `GET` API (also in response of POST and PUT). 

**Note:** If the user claim has been disabled after assigning its value for one or more user then also API should not send the claim value for those users.

Clipboard - January 19, 2024 6_02 PM
Clipboard - January 19, 2024 6_03 PM

duttarnab:
Parent issue# https://github.com/GluuFederation/flex/issues/1566
closes #7446

### DIFF
--- a/jans-config-api/plugins/user-mgt-plugin/src/main/java/io/jans/configapi/plugin/mgt/service/UserMgmtService.java
+++ b/jans-config-api/plugins/user-mgt-plugin/src/main/java/io/jans/configapi/plugin/mgt/service/UserMgmtService.java
@@ -26,6 +26,7 @@ import jakarta.inject.Inject;
 import jakarta.inject.Named;
 import org.apache.commons.beanutils.BeanUtils;
 import org.apache.commons.beanutils.BeanUtilsBean;
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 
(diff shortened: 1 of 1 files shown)


jans-config-api/plugins/user-mgt-plugin/src/main/java/io/jans/configapi/plugin/mgt/service/UserMgmtService.java
```diff
@@ -443,4 +443,8 @@ private String getDnForAttribute(String inum) {
         }
         return String.format("inum=%s,%s", inum, attributesDn);
     }
+
+    private static boolean isEmptyOrNullCollection(Collection<?> collection) {
+        return (collection == null || collection.isEmpty());
+    }
```
yuriyz: Approving but in future you can re-use `org.apache.commons.collections.CollectionUtils#isEmpty`

yuriyz (approved):

yuriyz (approved):

jans-config-api/plugins/user-mgt-plugin/src/main/java/io/jans/configapi/plugin/mgt/service/UserMgmtService.java
duttarnab: Done

yurem (approved):

Issue #7777, opened 2024-02-19
bug(saml): Need to parse SP metadata from XML to post to KC in individual elements


nynymike:
**Describe the bug**

![image](https://github.com/JanssenProject/jans/assets/3717101/5e8df173-f9bd-485b-8ddc-80a1340fc25e)

This does not work. Config API cannot parse SAML TR XML file. 

**Expected behavior**

Need to parse out individual data and post to KC API.

PR #7862, opened 2024-02-27
fix(keycloak-integration): saml metadata parsing #7777

uprightech:
This issue is in fullfillment for issue #7777 . 
There is a duplication of saml metadata parser classes which will be resolved  in subsequent versions.


jans-core/saml/src/main/java/io/jans/saml/metadata/model/LocalizedText.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class LocalizedText {
+
+    private String language;
+    private String text;
+    
+    public LocalizedText() {
+
+        this.language = language;
+        this.text = text;
```
yuriyz: Is this code auto-generated? Otherwise it's not clear what code above means. Empty constructor with assigning variables to itself.

Also when I look on all these boilderplate builders from "io.jans.saml.metadata.builder" package I think about lombok. If we plan to heavily rely on builders maybe it's good time to add lombok to not waste time on builders in future.

yuriyz (changes requested):

moabu (approved):

jans-core/saml/src/main/java/io/jans/saml/metadata/model/LocalizedText.java
uprightech: @yuriyz this has been fixed. Could you please accept the changes ?

yuriyz (approved):

PR #7971, opened 2024-03-05
perf(jans-pycloudlib): handle google secret versioning

### LINKED ISSUE #7944: perf(jans-pycloudlib): handeling secret versioning
**Describe the bug**
The secret class currently for google does not compare the existent data in the secret and hence keeps creating new versions each time the secret class is called. This results in very high number of active versions held in the secret manager. This causes huge costs on the end user as the versions are not intuitively being destroyed or disabled. 

**Expected behavior**

Two things need to be done on this side:
1. Only create a new version if new info is entering the secret . This does endure an API call but keeps the secret clean.
2. Disable the old secret once the new secret version has been posted or limit the number of active secrets to `5`.

**Additional context**
Current workaround is to execute the following :

```bash
#!/bin/bash
gcloud secrets versions list flex-secret --filter="state = enabled" --filter="createTime < '2024-03-02'" | grep "NAME:" | tr -d "NAME: " > versions_to_disable.txt
while read -r line; do
  gcloud secrets versions destory "$line" --secret=flex-secret --quiet
done < "versions_to_disable.txt"
```

iromli:
### Prepare

-------------------

#### Target issue
  
  
  
  
closes #7944 

#### Implementation Details
  

-------------------


moabu:
I do have a concern regarding possible leftover versions from race conditions or failed operations. I'm thinking we add a check in certmanager to run it after key rotation which would make sure old versions are cleaned up and only the last CN_GOOGLE_SECRET_MAX_VERSIONS are kept

iromli:
It is taken care by lower-level jans-pycloudlib google wrapper.

PR #8416, opened 2024-04-30
fix(charts): fix indentation for JAVA_OPTIONS in deployments.

### LINKED ISSUE #8415: fix(charts): Incorrectly indented template files causing helm install to fail.
When trying to install the Helm janssen v1.1.1 charts I encounter an error:
```
helm.go:84: [debug] error converting YAML to JSON: yaml: line 30: did not find expected key
YAML parse error on janssen/charts/janssen/charts/config-api/templates/deployment.yaml
```

To reproduce the error try installing the helm chart with an added userEnvs section, e.g.:
```yaml
  config-api:
    usrEnvs:
      normal:
        CN_CONFIG_API_PLUGINS: scim,user-mgt
```

This would normally install fine, but a change recently seems to have thrown out the indentation and causes the yaml file to be invalid. The change I think that has broken things is in this commit: https://github.com/JanssenProject/jans/commit/044290af76dbf1bb173cba237a7e39ac0f47e945

The added env for `CN_CONFIG_API_JAVA_OPTIONS` is indented only by 10, though the existing yaml includes are indented by 12. So if you have any custom user envs or secrets it results in the yaml file becoming corrupted.

The same issue exists for the following environment variables across several different deployments:
- CN_AUTH_JAVA_OPTIONS
- CN_CASA_JAVA_OPTIONS
- CN_CONFIG_API_JAVA_OPTIONS
- CN_FIDO2_JAVA_OPTIONS
- CN_LINK_JAVA_OPTIONS
- CN_SAML_JAVA_OPTIONS
- CN_SCIM_JAVA_OPTIONS

I wonder if there's possibly a linter or testing that can be done to prevent such issues in the future. The fix should be pretty easy for the charts, I'm pretty sure the env array just needs to be indented to the same depth and we should all be good.

saernz:
#### Target issue
Fixes #8415
  
Closes #8415

If someone else could test this fixes the issue that would be great. I've so far tested it for the config-api by manually editing my janssen charts folder with the change, which fixed the issue for me. I haven't tested any of the other deployments.


moabu:
Thank you @saernz ! We will review this PR.

moabu:
We would truly appreciate it if you can sign your commits and DCO basically use the `-s and -S` options in `git`. Thanks again @saernz

saernz:
Sorry, should be all good now. Was not too keen to use my email but I found out how github does it privately.

charts/janssen/charts/auth-server/templates/deployment.yml
misba7, line 1: 2 more whitespaces needed in auth-server

charts/janssen/charts/fido2/templates/deployment.yml
misba7, line 1: 2 more whitespaces needed in fido2

charts/janssen/charts/scim/templates/deployment.yml
misba7, line 1: 2 more whitespaces needed in scim

misba7 (changes requested):

charts/janssen/charts/auth-server/templates/deployment.yml
saernz: Hi @misba7 I think you are technically correct, but something strange is going on with the indentations of the charts.

If you look at the actual raw source of the deployment chart on the line there is only 10 spaces of indentation, though the template is using an indent function value of 12.

If you actually download the chart with helm, unzip the templates and open them up, they actually have an indentation of 12 spaces, even though in the raw source the lines only had an indent of 10.

I've found out this is because the raw template in the source code does not indent the array of containers, though in the unzipped charts, there is an indentation of 2 spaces for the containers array. I think somewhere in the pipeline the charts formatting is being corrected, so even though those lines only have an indent of 10 in the raw source, when they make it into the released charts, they end up with a correct indentation of 12.

I'm not sure where this formatting is happening, though it probably needs to be fixed to minimize confusion. Once it is fixed, someone's probably going to have to fix all the indenting in the raw source files too.

saernz: Ah I forgot to add, you can see what's happening by running: 

`helm template janssen jans.io/janssen -s charts/config-api/templates/deployment.yaml --set config-api.usrEnvs.normal.CN_CONFIG_API_PLUGINS="scim" --debug`

if you have the jans.io repo added.

saernz:
I think as per my reasoning in the resolved requested change we should probably be able to move ahead with this now. I'm not sure if there is a way to run a test release for the helm charts to see what happens?

As per my comments in the requested change, I believe there is something in the helm chart release pipeline that fixes the indentation of the raw source files, so even though the JAVA_OPTIONS are only indented by 10 in the raw source, once they pass through that pipeline, they become indented by 12. This is because the parent array of containers is indented differently in the raw helm templates in GitHub from what comes out the other side in the released helm chart where it has +2 spaces for array items. You can unzip the released helm charts and compare the differences from there.

I think whatever is happening to cause that is probably better fixed in another issue, though for now this fix should at least maintain the status quo.

iromli (approved):

moabu (approved):
I think @saernz is right and I believe it's coming from a source that's using a windows computer 🫢

misba7 (approved):

Issue #8671, opened 2024-06-06
feat(jans-lock): add metadata/autit/config/sse endpoints


yurem:
Add metadata/autit/config/sse endpoints

yurem:
Implemented

PR #8672, opened 2024-06-06
Lock Endpoint Scaffolding

yurem:
#8671


nynymike (approved):
Looks good to me. You don't need an endpoint to return the Lock Master metadata (i.e. .well-known/lock-master-configuration) ?

yuriyz (approved):

jans-lock/lock-master/service/src/main/java/io/jans/lock/service/util/ServerUtil.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Janssen Project software is available under the Apache License (2004). See http://www.apache.org/licenses/ for full text.
```
nynymike, line 2: Please align with this format: https://www.apache.org/licenses/LICENSE-2.0#apply

PR #8792, opened 2024-06-27
feat(jans-keycloak-integration): update kc-saml integration installation for ce and cn #8776

### LINKED ISSUE #8776: feat(jans-keycloak-integration): update kc-saml integration installation for ce and cn
Completes issue #8614 .
Completes issue #7399 .

#### Update to SPIs

 The keycloak SPIs have been merged into a single SPI to reduce duplicate dependencies between SPIs and bring them under one umbrella. The artifact that bundles them all is called `kc-jans-spi` with the binaries being built here https://jenkins.jans.io/maven/io/jans/kc-jans-spi/1.1.3-SNAPSHOT/ 
 Remove reference to the following SPIs:
 - `kc-jans-authn-plugin`
 - `kc-jans-storage-plugin`
 
 Instead of copying their files to `/opt/keycloak/providers` , copy the files for `kc-jans-spi` in the same directory. 
 
 #### Quarkus properties file for CDI 
 
 A quarkus properties file has been added in `jans-linux-setup/jans_setup/templates/jans_saml/quarkus.properties`. During setup , copy said file to `/opt/keycloak/conf/`
 
 #### Changes to service file
 
Changes have been made to `kc.service` ,specifically the startup command `ExecStart`. All it's parameters have been removed and moved to `jans-linux-setup/jans_setup/templates/jans-saml/keycloak.conf`
Make sure the file is properly rendered with the correct parameters. 
In addition , make sure the parameter `jansBaseFolder` is available when the `kc.service` service template is being rendered. 
This contains the base configuration folder for jans (`/etc/jans`) usually. 

#### Remove reference to  Jans SCIM Client

 There was a requirement to create a scim client called `Jans SCIM Client for SAML`. We will have to remove it as it is no more necessary.
 
 #### Switch keycloak from dev to production. 
 
  In `kc.service` , the start argument is `start-dev` , but ought to be `start` so we run in production mode , but this requires switch from the H2 file database used by keycloak to a production database. This may require further discussion , but I think it can be done already for postgresql.  Supported databases by keycloak can be found here https://www.keycloak.org/server/db.
 But when the database we use to install is postgresql , for now , let's do the following:
  - In the kc.service file , in the startup command , switch from `start-dev` to `start`
  - create a postgresql user called `keycloak` and a database of the same name. 
  - add the above to the parameters to render the file `keycloak.conf` (check the `db`,`db-username`,`db-password` and `db-url` in the same file). The `db-url` should be of the format `jdbc:postgresql://<server>:<port>/<db-name>`
  
  In addition during startup , it's important to run `/opt/keycloak/bin/kc.sh build` before starting the keycloak service.

uprightech:
Closes #8776


jans-linux-setup/jans_setup/app_info.json
```diff
@@ -17,6 +17,6 @@
   "CRYPTOGRAPHY": "https://files.pythonhosted.org/packages/20/8b/66600f5851ec7893ace9b74445d7eaf3499571b347e339d18c76c876b0f9/cryptography-37.0.4-cp36-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.whl",
   "TWILIO_MAVEN": "https://repo1.maven.org/maven2/com/twilio/sdk/twilio/",
   "TWILIO_VERSION": "7.17.0",
-  "KC_VERSION": "24.0.3",
+  "KC_VERSION": "25.0.1",
```
moabu, line 20: @iromli we need to sync in with KC versions in CN

moabu (dismissed):

jans-linux-setup/jans_setup/app_info.json
iromli: Yes, we can. For other assets, they would be sync after this PR merged.

iromli (approved):

moabu (approved):

PR #8819, opened 2024-06-28
feat(jans-lock): Acquire config from JS, Improved codebase structure

sokorototo:
### Prepare

-------------------

\nPlease check the below before submitting your PR. The PR will not be merged if there are no commits that start with `docs:` to indicate documentation changes or if the below checklist is not selected.\n- [x] **I confirm that there is no impact on the docs due to the code changes in this PR.**


jans-lock/cedarling/src/authz/token2entity.rs
```diff
@@ -0,0 +1 @@
+
```
SafinWasi: Blank files should not be committed. If this is a placeholder please leave a comment in the file

jans-lock/cedarling/src/lock_master/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+
+		// send
+		let res = http::post(&openid_config.registration_endpoint, http::PostBody::Json(client_req), &[])
+			.await
+			.expect_throw("Unable to register client");
+		res.into_json().await.expect_throw("Unable to parse client registration response")
+	};
+
+	let grant: types::OAuthGrantResponse = {
+		// https://gluu.org/docs/gluu-server/4.0/admin-guide/oauth2/
```
SafinWasi: As this is under the Janssen project, we should be referring to Janssen docs: https://docs.jans.io/v1.1.2/admin/auth-server/endpoints/token/

jans-lock/cedarling/src/lock_master/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+	// Get OAuthConfig
+	let res = http::get(&lock_master_config.oauth_as_well_known, &[]).await.expect_throw("Unable to fetch LockMasterConfig from URL");
+	let openid_config = res.into_json::<types::OAuthConfig>().await.unwrap_throw();
+
+	let client: types::OAuthDynamicClient = {
+		// OpenID dynamic client registration
+		let client_req = types::OAuthDynamicClientRequest {
+			client_name: application_name,
+			application_type: "web",
+			redirect_uris: &[],
```
SafinWasi: Redirect URI array must be non empty. In this case, since we're using the client credentials grant it is sufficient to use the OpenID provider's hostname.

jans-lock/cedarling/src/lock_master/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+			client.client_id
+		};
+
+		let grant = types::OAuthGrantRequest {
+			scope: "cedarling",
+			grant_type: "client_credentials",
+		};
+
+		// send
+		let url = format!("{}/token", openid_config.token_endpoint);
```
SafinWasi: This might be malformed. The token endpoint for jans is `https://hostname/jans-auth/restv1/token`, which would add an extra `/token` to the end

jans-lock/cedarling/src/lock_master/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+		sse.set_onopen(Some(onopen.unchecked_ref()));
+	}
+
+	// Get OAuthConfig
+	let res = http::get(&lock_master_config.oauth_as_well_known, &[]).await.expect_throw("Unable to fetch LockMasterConfig from URL");
+	let openid_config = res.into_json::<types::OAuthConfig>().await.unwrap_throw();
+
+	let client: types::OAuthDynamicClient = {
+		// OpenID dynamic client registration
+		let client_req = types::OAuthDynamicClientRequest {
```
SafinWasi, line 48: The `"grant_types": ["client_credentials"]` field is missing from the dynamic client registration request as per https://www.rfc-editor.org/rfc/rfc7591#section-2.

jans-lock/cedarling/src/lock_master/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+			.expect_throw("Unable to register client");
+		res.into_json().await.expect_throw("Unable to parse client registration response")
+	};
+
+	let grant: types::OAuthGrantResponse = {
+		// https://gluu.org/docs/gluu-server/4.0/admin-guide/oauth2/
+		let token = if let Some(client_secret) = client.client_secret {
+			format!("{}:{}", client.client_id, client_secret)
+		} else {
+			console::warn_1(&JsValue::from_str("Client Secret is not provided"));
```
SafinWasi, line 70: I'm not sure why this check is needed. You always get a client ID and secret back from DCR.

jans-lock/cedarling/src/lock_master/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+			client.client_id
+		};
+
+		let grant = types::OAuthGrantRequest {
+			scope: "cedarling",
```
SafinWasi: Incorrect scope. Please follow https://gluu.org/swagger-ui/?url=https://raw.githubusercontent.com/JanssenProject/jans/main/jans-lock/lock-master.yaml to see what scopes you'll use.

SafinWasi (changes requested):
A readme is also essential so that we can run this locally and test

kdhttps (changes requested):
1. It will be good to add unit test cases(TDD) for every function 
2. I think we should follow the DDD and clean architecture. You can start writing usecases that means cover basic functionality and functions first like I added in TODOs in lock-project .
image

jans-lock/cedarling/src/lock_master/mod.rs
sokorototo: Ah, thanks for the doc reference!

jans-lock/cedarling/src/lock_master/mod.rs
sokorototo: Oh yeah, that's a malformed endpoint

jans-lock/cedarling/src/lock_master/mod.rs
sokorototo: And what would that be? I'm not sure how this works outside the Web

jans-lock/cedarling/src/lock_master/mod.rs
sokorototo: Fixed!

jans-lock/cedarling/src/lock_master/mod.rs
sokorototo: I actually don't remember what frame of reference I got this from, but it _can_ miss in some circumstances. Besides, it's a check for that edge case and doesn't affect normal behaviour

jans-lock/cedarling/src/lock_master/mod.rs
sokorototo: image

What does this mean? According to the diagram I need an Access Token with a `cedarling` scope to fetch a Policy Store from the Lock Master. Given I don't have the Lock Master's actual OpenId config, I'm not sure what scope I need to access this resource

sokorototo:
@kdhttps while I do understand the importance of Clean Architecture and writing sufficient tests, can we get something that works first then we can start the polish work? I'll definitely do it, I just don't want to get bogged down in structural work right now

sokorototo:
@SafinWasi `https://lock.master.gluu.cloud` doesn't seem to point to anything, so I can't get the relevant OpenId config, and thus can't get the supported scopes. I might resolve to reading through the source code, but I'm not sure if that's a forward-looking solution. 

On the matter of writing the README, I'm restoring my Typora license then I'll get to it.

kdhttps:
> @kdhttps while I do understand the importance of Clean Architecture and writing sufficient tests, can we get something that works first then we can start the polish work? I'll definitely do it, I just don't want to get bogged down in structural work right now

I'm sorry, but I must respectfully disagree with you. At least we should write unit test cases for each function.

sokorototo:
@kdhttps, can we agree then that the tests are not final and are written to fail.

kdhttps:
@sokorototo you can check out the [cedar-wasm code](https://github.com/cedar-policy/cedar/blob/main/cedar-wasm/src/schema_and_entities_and_context.rs), it is a good example of TDD. It has small functions, accepting everything in parameters and respecting true/false test cases. The goal is code should look and execute health(bug free) and it is only possible by adding test cases.

jans-lock/cedarling/src/lock_master/mod.rs
SafinWasi: This would be the hostname of wherever you're registering. So let's say you're registering at https://account.gluu.org/jans-auth/restv1/register, so your hostname would be https://account.gluu.org, and your redirect_uris array would be `["https://account.gluu.org"]`

jans-lock/cedarling/src/lock_master/mod.rs
SafinWasi: image

You'll want to request a token with these two scopes.

jans-lock/cedarling/src/lock_master/mod.rs
SafinWasi: Sure, but leaving possibly dead code will make it harder to audit later on. Please annotate with a todo specifying the edge case and if this can be removed in the future.

SafinWasi (approved):
Looks good to me. @kdhttps can you check functionality and JS bindings?

kdhttps:
> Looks good to me. @kdhttps can you check functionality and JS bindings?

@sokorototo I am getting this on my HTML page.

image


My JS Code looks like:
```
<script type="module">
      import init from "./out/cedarling.js";

      const config = {
        policyStore: {
          // can be "local", "remote" or "lock-master",
          // each strategy requires different parameters, see below
          strategy: "remote",
          url: "https://raw.githubusercontent.com/JanssenProject/jans/main/jans-lock/cedarling/policy-store/default.json",
        },
        // if policy-store.json is compressed using deflate-zlib
        decompressPolicyStore: false,
        // [OPTIONAL] How often, in milliseconds, will the cedarling refresh it's TrustStore. The cedarling won't refresh it's trust store if not omitted
        trustStoreRefreshRate: 2000,
        // Set of jwt algorithms that the cedarling will allow, empty if omitted
        supportedAlgorithms: ["HS256", "HS384", "RS256"],
      };

      async function run() {
        await init(config);
      }

      run();
    </script>
```

sokorototo:
@kdhttps to build for the Web, you use `wasm-pack build --release --target web`. I was testing with node, so I forgot to include this, let me patch it in.

kdhttps:
ok, thanks but now getting this @sokorototo . 

image

sokorototo:
@kdhttps can you check if your dev server is setup correctly? Is the wasm binary being loaded in the Network Section of the dev tools? Since everything seems to load on my side

kdhttps:
I investigate the `cedarling.js`. It is not invoking the wasm binary load function. so I manually added it to my HTML page. Not sure if this way is right or wrong. My HTML JS code is:
```
    <script type="module">
      import __wbg_init, { init } from "./out/cedarling.js";

      const config = {
        policyStore: {
          strategy: "remote",
          url: "https://raw.githubusercontent.com/kdhttps/cedar-policies/main/example-policy-store.json",
        },
        decompressPolicyStore: false,
        trustStoreRefreshRate: 2000,
        supportedAlgorithms: ["HS256", "HS384", "RS256"],
        policyStoreId: "gluu_store",
      };

      async function run() {
        await __wbg_init();
        await init(config);
      }

      run();
    </script>
```

`await __wbg_init();` after manually invoking the wasm binary I can now see it in the network tab. and now looks like it is actually calling the wasm init function but getting an error:

image

kdhttps (approved):

SafinWasi (approved):

moabu (approved):

Issue #8908, opened 2024-07-09
ProjectPasskeys: Add support for WebAuthn Hints


yackermann:
In the latest iterations of the WebAuthn specification, a new mechanism for managing experience was added called "hints".

https://w3c.github.io/webauthn/#dom-publickeycredentialcreationoptions-hints
https://w3c.github.io/webauthn/#enum-hints
This mechanism allows RP to control what user experience they with to push, such as enforce "security-key", or just include "client-device". Together with authenticatorSelection, and `transports` field, RP will be able to carefully control resulting experience.

shekhar16:
its done closing this ticket.

PR #9233, opened 2024-08-20
feat(jans-fido): #8908 add support for webauthn hints

shekhar16:
#8908

### DIFF
--- a/jans-fido2/server/src/main/java/io/jans/fido2/service/operation/AttestationService.java
+++ b/jans-fido2/server/src/main/java/io/jans/fido2/service/operation/AttestationService.java
@@ -9,6 +9,7 @@ package io.jans.fido2.service.operation;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.jans.entry.PublicKeyCredentialHints;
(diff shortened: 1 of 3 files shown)


jans-fido2/server/src/main/java/io/jans/fido2/service/verifier/CommonVerifiers.java
```diff
@@ -99,6 +100,18 @@ public String verifyRpDomain(JsonNode params) {
         return documentDomain;
     }
 
+    public PublicKeyCredentialHints verifyHints(JsonNode params) {
+        String documentDomain;
+        PublicKeyCredentialHints hints = null;
+        if (params.hasNonNull("user-agent")) {
+            documentDomain = params.get("user-agent").asText();
```
yurem, line 107: We need to change variable name to conform real value

jans-fido2/server/src/main/java/io/jans/fido2/service/operation/AttestationService.java
```diff
@@ -161,6 +162,11 @@ public ObjectNode options(JsonNode params) {
 			log.debug("Put excludeCredentials {}", excludedCredentials);
 		}
 
+		//put hints
+		PublicKeyCredentialHints hints = commonVerifiers.verifyHints(params);
+		optionsResponseNode.put("hints",hints.getValue());
```
yurem, line 167: Is it right place in result to add hint node?

PR #9267, opened 2024-08-26
feat(jans-core): use DB document store by default

### LINKED ISSUE #9179: feat: add pull mechanism for custom assets pushed to the DB
Copying asset files on server: `documentStoreService.saveDocumentStream()` should not be done by the configAPI as it is done today. The current design assumes no micro services and that it lives in a VM which is not true. The design needs to work for all sort of installations: 

1. On start of each service it would pull the data from the persistence and place it in the right place. Whether thats VM or CN it should be the same. Each service needs to handle this .

2. On save each service should save it also appropriately just like what we have atm inside the pod or VM. The config api should not do this. If it fails then we can say a restart of the service is required. But each service needs to on some sort of cron check the table and attempt to pull the data associated to it from the table and place it appropriately. Meaning it needs to handle its domain work. `documentStoreService.saveDocumentStream()` should be called by each service following the assets that belong to it.

3. If there is an option store or not file in file system for future processing service should try to avoid storing file in file system.

We need to add generic DocumentStoreManager service which should pull files which belong to specific service periodically. And put them into file system. By default it should pull all files to file system unless application can specify which files store in memory store. To do quick check for changes we need to implement https://github.com/JanssenProject/jans/issues/9186

### LINKED ISSUE #9268: fix: feat(jans-core): use DB document store by default -autocreated
This issue was created because no open issues were referenced in your PR (#9267).

yurem:
closes #9179

Closes #9268,


devrimyatar (approved):

mo-auto:
Error: Hi @yurem, You did not reference an open issue in your [PR](9267). I attempted to create an [issue](#9268) for you.    
 Please update that [issues'](#9268) title and body and make sure I correctly referenced it in the above PRs body.

yuriyz (approved):

PR #9575, opened 2024-09-23
feat(jans-cedarling): Policy Store: Parse Schema and Policies

### LINKED ISSUE #9568: feat(jans-cedarling): Policy Store: Parse Schema - both valid and invalid
Implement Policy Store: Parse Schema - both valid and invalid.

* [x] Validate the Schema from the Policy Store
* [x] Log if Success
* [x] If failure, log specific message -- ideally what part of the schema did not validate

olehbozhok:
### Prepare

-------------------

#### Target issue
  [link](https://github.com/JanssenProject/jans/issues/9568)
  
closes #9568

#### Implementation Details
  Added parsing a policy store with only schema key.
  Currently, support loading only from memory

-------------------


jans-cedarling/cedarling/src/init/policy_store.rs
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Copyright (c) 2024, Gluu, Inc.
+ */
+
+use crate::models::policy_store::{PolicyStore, PolicyStoreMap};
+use crate::models::policy_store_config::{PolicyStoreConfig, PolicyStoreSource};
+
+/// Error cases for loading policy
+#[derive(Debug, thiserror::Error)]
+pub enum ErrorLoadPolicyStore {
```
duttarnab, line 13: Rather than creating different enums for showing error messages, can we just make a single enum for all error messages inside `\jans-cedarling\cedarling\src\models\enum\error`.

**for Example:**
```
pub enum CedarlingApplicationError {
 #[error("{0}")]
    LoadPolicyStoreJsonParseError(#[from] serde_json::Error),
    #[error("store policy is empty")]
    LoadPolicyStoreWithEmptyPolicy,
    #[error("the `store_key` is not specified and the count on policies more than 1")]
    LoadPolicyStoreMoreThanOnePolicy,
    #[error("could not found policy by id: {0}")]
    LoadPolicyStorePolicyNotFound(String),
}
```

jans-cedarling/cedarling/src/init/cedar_schema.rs
```diff
@@ -0,0 +1,81 @@
+/*
+ * This software is available under the Apache-2.0 license.
+ * See https://www.apache.org/licenses/LICENSE-2.0.txt for full text.
+ *
+ * Copyright (c) 2024, Gluu, Inc.
+ */
+
+use base64::prelude::*;
+
+const MSG_UNABLE_DECODE_SCHEMA_BASE64: &str = "unable to decode cedar policy schema base64";
```
duttarnab: These errors should be handled using the `CedarlingApplicationError ` enum . ( as discussed in https://github.com/JanssenProject/jans/pull/9575#discussion_r1772861453)

duttarnab (changes requested):
Please see the inline comments.

Please remove the following from README. We can add these once the bindings  are added.

image

jans-cedarling/cedarling/src/init/policy_store.rs
olehbozhok: I like the idea to return the error struct in functions that cover all possible error cases for only the current situation. Because it will be easier to maintain when application grows.

And If we have somethinglike `CedarlingApplicationError`
It will have all possible errors for all application. 

And when we will be wanted to handle the error cases with match statement, it will be hard to understand what exactly cases current function returns. Because `CedarlingApplicationError` have all possible cases in the application.

jans-cedarling/cedarling/src/init/cedar_schema.rs
olehbozhok: We use it in the function `parse_cedar_schema`. It is wrapper to seamless deserialize `cedar_policy::Schema` from out JSON policy store.

Here how it is used in the macros:
```
#[derive(Debug, serde::Deserialize)]
pub struct PolicyStore {
    #[serde(deserialize_with = "parse_cedar_schema")]
    pub schema: cedar_policy::Schema,
}
```

In function, we can't return our custom type. Because we return generic of `serde::Deserializer`
```
pub(crate) fn parse_cedar_schema<'de, D>(deserializer: D) -> Result<cedar_policy::Schema, D::Error>
where
    D: serde::Deserializer<'de>,
{
...
```

Here is [link](https://serde.rs/error-handling.html) with documentation about custom error handling in the `serde`.
So to return error from custom `Deserializer` we need to use implementation
```
impl de::Error for Error {
    fn custom<T: Display>(msg: T) -> Self {
        Error::Message(msg.to_string())
    }
}
```
Where error is casting to the string.

And actually it was made with
```
        serde::de::Error::custom(format!("{MSG_UNABLE_DECODE_SCHEMA_BASE64}: {}", err,))
```

and it was cowered by:
```
pub enum ErrorLoadPolicyStore {
    #[error("{0}")]
    JsonParce(#[from] serde_json::Error),
    ...
```

olehbozhok:
> Please see the inline comments.
> 
> Please remove the following from README. We can add these once the bindings are added.
> 
> image

Removed

jans-cedarling/cedarling/src/init/policy_store.rs
duttarnab: > And If we have something like CedarlingApplicationError
It will have all possible errors for all application.

This is what we practise in most of applications. We do not clutter our project with multiple Enums for error messages.

> And when we will be wanted to handle the error cases with match statement, it will be hard to understand what exactly cases current function returns. Because CedarlingApplicationError have all possible cases in the application.

The naming convention of enum items should be sufficient of make its purpose clear. We don't need to make multiple error enums for this. 

And I could see `ErrorLoadPolicyStore` in cedar_schem.rs and policy_store.rs. Code duplication does not make sense.

jans-cedarling/cedarling/src/init/cedar_schema.rs
duttarnab: I was just saying that we should maintain uniformity of code everywhere. If we are taking the error messages as enum items somewhere then let's make it a practice everywhere. and we can easily convert an enum to a string. ref: `https://stackoverflow.com/questions/32710187/how-do-i-get-an-enum-as-a-string`.

olehbozhok: Otherwise, we need to create struct with RAW data and manually call a function to decode each field.
Which approach do you think is better?

duttarnab: as discussed, we can do something like the below to solve `serde` error handling
```
fn custom_deserialization_error(err: base64::DecodeError, errorEnum: ErrorLoadPolicyStore) -> de::value::Error {
    let message = errorEnum.to_string(); // Using the enum variant
    serde::de::Error::custom(format!("{}: {}", message, err))
}
```

olehbozhok: We discussed and decided to stay with const messages approach

duttarnab (approved):

nynymike (changes requested):
The cedarling should not init if the schema is invalid.

nynymike (approved):
Ok

Lilit0x (changes requested):
requested changes have been made in #9584

moabu (approved):

Issue #9369, opened 2024-09-04
ProjectPasskeys: Deprecate folder properties from fidoConfiguration.


shekhar16:
Deprecate mdsCertsFolder
 Deprecate mdsTocsFolder
 Deprecate authenticatorCertsFolde

shekhar16:
Currently we are saving certs and jwt files in folders at the time of installation, so we need to make few changes in setup installers too to save certs and jwt files in DB instead of folders.
for example 
https://github.com/JanssenProject/jans/blob/main/jans-linux-setup/jans_setup/setup_app/installers/fido.py#L17
https://github.com/JanssenProject/jans/blob/main/jans-linux-setup/jans_setup/setup_app/installers/fido.py#L18

shekhar16:
code has been merged and files are getting stored in DB now.
Closing this ticket.

Issue #9610, opened 2024-09-27
feat(jans-fido-setup) ProjectPasskeys : move fidoconfig folder properties to db


shekhar16:
As a part of ticket #9369, we are moving the fido certs files to DB instead of folders, we need to make changes in setup installers too.
here is the few examples:
https://github.com/JanssenProject/jans/blob/main/jans-linux-setup/jans_setup/setup_app/installers/fido.py#L17
https://github.com/JanssenProject/jans/blob/main/jans-linux-setup/jans_setup/setup_app/installers/fido.py#L18

if more changes needed in setup for this need to be make.

devrimyatar:
@shekhar16 When we base64 encode file `https://mds.fidoalliance.org/`, it becomes around 5MB and not possible to store in OpenDJ. I am getting "Broken pipe" error.

yurem:
I think we need hold on work on this issue. I will work on https://github.com/JanssenProject/jans/issues/224

devrimyatar:
@yurem the file downloaded from https://mds.fidoalliance.org/ is jwt, not binary. Does it matter?

devrimyatar:
Implemented in commit https://github.com/JanssenProject/jans/commit/e23db2dcfa4216496d61f898d82d62ea5763e4de

devrimyatar:
Implemented in commit https://github.com/JanssenProject/jans/commit/e23db2dcfa4216496d61f898d82d62ea5763e4de

PR #9611, opened 2024-09-27
feat(jans-fido): move fidoconfig folder properties to db #9369

shekhar16:
#9369


shekhar16:
With this PR we also need to make setup changes here is the ticket https://github.com/JanssenProject/jans/issues/9610

yurem (approved):

PR #9694, opened 2024-10-09
feat(jans-cedarling): Parse Cedar Schema for Access Token and Workload entities

### LINKED ISSUE #9677: feat(jans-cedarling): Parse Cedar Schema for Access Token and Workload entities
We need to create `access token` and `Workload` entity for cedar-policy based on the policy store scheme. To not hard-code creating entities in the application.

Based on the documentation [json-schema-grammar](https://docs.cedarpolicy.com/schema/json-schema-grammar.html) need to create own parser for cedar_policy scheme and create entities based on it.

We hard-code only entity type and what field is created for ID

 `Jans::Workload`  ID is field `client_id`
 `Jans::Access_token` ID is field `aud`

**Additional context**
entities
```
    entity Workload  = {
        client_id: String,
        iss: TrustedIssuer,
        name: String,
        org_id: String,
    };
    entity Access_token  = {
        aud: String,
        exp: Long,
        iat: Long,
        iss: TrustedIssuer,
        jti: String,
        nbf: Long,
        scope: String,
        org_id: String,
    };
    ```

As a sample resource we could use an "Issue" or "Ticket" and the policy can be something like allow action Update when workload org_id == ticket.org_id

### PR DESCRIPTION
**Pull request by olehbozhok**
### Prepare

-------------------

#### Target issue
  [link](https://github.com/JanssenProject/jans/issues/9677)
  
  
  
closes #9677

#### Implementation Details
  
  added custom parsing the cedar schema to JSON and create Access Token and Workload entities based on it

-------------------

### CODE REVIEW

**Review by rmarinn — DISMISSED** [review]
Please keep your PR within scope and localized to as few files as possible next time... I think there are multiple commits here that are not related to the [target issue](https://github.com/JanssenProject/jans/issues/9677) which made your PR challenging to review...

**Review by duttarnab — CHANGES_REQUESTED** [review]
- The code has compilation errors at  `jans-cedarling\bindings\cedarling_python\src\authorize\request.rs` and `jans-cedarling\bindings\cedarling_python\src\cedarling.rs`

- `cargo test --workspace` is failing.

**Comment by olehbozhok** [comment]
> Please keep your PR within scope and localized to as few files as possible next time... I think there are multiple commits here that are not related to the [target issue](https://github.com/JanssenProject/jans/issues/9677) which made your PR challenging to review...

I'll give it a try, but it will be quite challenging to verify if everything works correctly without the actual implementation of the authorize method.

[Here ](https://chat.gluu.org/group/service_cedarling?msg=jsqE62jxK3rQfEWBi) our discussion with @nynymike about 
```
As a sample resource we could use an "Issue" or "Ticket" and the policy can be something like allow action Update when workload org_id == ticket.org_id
```
What I made.

File: jans-cedarling/cedarling/src/authz/entities/meta.rs [hunk 1]
```diff
@@ -7,26 +7,12 @@
...
-                cedar_policy_type: CedarPolicyType::EntityUid {
-                    entity_type: "Jans::TrustedIssuer",
-                },
-            },
-        ]
-    );
+    pub(crate) static ref WorkloadEntityMeta: EntityMetadata<'static> =
+        EntityMetadata::new("Jans::Workload", "client_id",);
+    pub(crate) static ref AccessTokenMeta: EntityMetadata<'static> =
+        EntityMetadata::new("Jans::Access_token", "aud",);
```
nynymike: I'm not sure I like this... it's possible that there could be more then one access token in the future. It's also possible that clients will use a stable client_id (i.e. a URI) across multiple IDPs. In this case, you'd have a collission with the `Jans::Access_token`. It would actually be best to use the `jti` of the access token which is a unique identifier for that JWT.

olehbozhok: resolved

rmarinn:
> I'll give it a try, but it will be quite challenging to verify if everything works correctly without the actual implementation of the authorize method. Because we need to use cedar_policy schema.

I think it's better to define [traits](https://doc.rust-lang.org/book/ch10-02-traits.html) that define your expected behavior instead of trying to implement everything in one PR. Then if you want to make unit-tests, just write structs that implement the trait you made.

Defining traits would also make it easier for other people know your intention on how your could should be used :)

If ever we find the project to be bloated in the future because of these traits we just just fix it then... in another _issue_.  For now, since we're still building things from the ground up, it's expected that a lot of things you would need would be missing.

duttarnab (approved):

jans-cedarling/bindings/cedarling_python/example.py
```diff
@@ -53,8 +53,3 @@
 # show logs
 print("Logs stored in memory:")
 print(*instance.pop_logs(), sep="\n\n")
-
-example_access_token = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJzb21lX3VuaXF1ZV9pZCIsImlzcyI6Imh0dHA6Oi8vZXhhbXBsZS5jb20ifQ.0EBtuIHAKtPZRKqX6LBXOZ52erMB2xuMTfVSoaYHK-o'
-request = Request(access_token=example_access_token)
-# if
-instance.authorize(request)
```
SafinWasi, line 60: Are we not using `authorize` as an example call?

SafinWasi (commented):
Tests work and python code looks good, just a bit confused why `authorize` call was removed from `example.py`

jans-cedarling/bindings/cedarling_python/example.py
olehbozhok: Yes, it is removed because it will be implemented in the 
https://github.com/JanssenProject/jans/issues/9706

current PR is already huge

SafinWasi (approved):
Looks good to me

PR #9865, opened 2024-10-21
refactor(jans-cli-tui): display popup for save action

### LINKED ISSUE #7983: refactor(TUI): confirmation message for `save` action
After changing any configuration using TUI, when we hit `save`, the confirmation message about changes getting successfully saved, is delivered in either of these two ways on different screens.

1. A pop-up

image

2. A message in the bottom toolbar

image

So, the user experience is inconsistent. I would request:

- Pick one approach to deliver this message and implement it consistently across TUI all screens
- I would recommend the first approach with a dedicated pop-up as the bottom bar's tooltip is hard to notice. In this case, to the user, the save button seems to be doing nothing till the time they spot the message at the bottom. Resulting in bad user experience.

devrimyatar:
closes #7983

### DIFF
--- a/jans-cli-tui/cli_tui/plugins/010_auth_server/main.py
+++ b/jans-cli-tui/cli_tui/plugins/010_auth_server/main.py
@@ -1026,17 +1026,7 @@ class Plugin(DialogUtils):
                 data=pathches
                 )
             self.app.app_configuration = response.json()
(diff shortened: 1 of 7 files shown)


yuriyz (approved):

yuriyzz (approved):

mo-auto:
Error: Hi @devrimyatar, You did not reference an open issue in your [PR](9865). I attempted to create an [issue](#9866) for you.    
 Please update that [issues'](#9866) title and body and make sure I correctly referenced it in the above PRs body.

PR #9936, opened 2024-10-25
feat(cloud-native): add support for legacy and simple JSON data

### LINKED ISSUE #9933: feat(cloud-native): add support for legacy and simple JSON data
Since `jans-orm` introduces simple JSON data format with the following features:

1. JSON data `{"v": []}` is deprecated in favor of `[]`
2. Backward-compatibility with older JSON data

The cloud-native installation must support these scenarios:

1. OCI images should use new JSON data by default
2. charts must provide attribute to choose using old or new JSON data format

iromli:
### Prepare

-------------------

#### Target issue
  
  
  
  
closes #9933 

#### Implementation Details
  

The changeset adds support for legacy and new simple JSON data format using auto-detection of column's value.

**How to test:**

1. Jans 1.1.5 and lower:

   - install jans 1.1.5 chart; ensure format `jansAppConf.jansSmtpConf` value is `{"v": []}`
   - upgrade to chart in this PR (change image to `1.1.6_json`); ensure format `jansAppConf.jansSmtpConf` value is still `{"v": []}` after upgrade

1. Jans 1.1.6+:

   - install chart in this PR (change image to `1.1.6_json`); ensure format `jansAppConf.jansSmtpConf` value is `[]`
   - do upgrade; ensure format `jansAppConf.jansSmtpConf` value is still `[]` after upgrade

-------------------


charts/janssen-all-in-one/values.yaml
```diff
@@ -173,6 +173,8 @@ configmap:
   cnOpaUrl: http://opa.opa.svc.cluster.cluster.local:8181/v1
   # -- Message type (one of POSTGRES, REDIS, or DISABLED)
   cnMessageType: DISABLED
+  # -- Enable simple JSON data format of multivalued columns (JSON array instead of object) in MySQL. Default to true. Existing installation should set the value to false.
+  cnSqlSimpleJson: true
```
moabu: This flag can potentially create issues. I suggest we determine that via the command run on the charts i.e if its `helm install` or `helm upgrade —install` then set it to `true` . If its `helm upgrade` set it to `false`  

We can make use of `.Release.IsInstall` and `.Release.IsUpgrade` to change these values accordingly in the main configMap. 


https://helm.sh/docs/chart_template_guide/builtin_objects/

moabu (changes requested):

charts/janssen-all-in-one/values.yaml
iromli: @moabu that causes more issues when using `v1.1.6` and upgrade to another version (i.e. `v1.1.7`). See example below:


1. Fresh install `v1.1.6`:

    ```
    » helm -n jans upgrade --install --force jans ../../charts/janssen -f values.yaml
    Release "jans" does not exist. Installing it now.
    NAME: jans
    LAST DEPLOYED: Wed Oct 30 01:23:17 2024
    NAMESPACE: jans
    STATUS: deployed
    REVISION: 1
    TEST SUITE: None
    
    » kubectl -n jans exec jans-auth-server-84dd64bb44-lgvqv -- env | rg SIMPLE_JSON
    MYSQL_SIMPLE_JSON=true
    ```
    As we can see, the `MYSQL_SIMPLE_JSON` is properly set to `true` because `.Release.IsInstall` is true.

1.  Upgrade existing `v1.1.6` installation:

    ```
    » helm -n jans upgrade --install jans ../../charts/janssen -f values.yaml
    Release "jans" has been upgraded. Happy Helming!
    NAME: jans
    LAST DEPLOYED: Wed Oct 30 01:27:22 2024
    NAMESPACE: jans
    STATUS: deployed
    REVISION: 2
    TEST SUITE: None
    
    » kubectl -n jans exec jans-auth-server-84dd64bb44-g94tt -- env | rg SIMPLE_JSON
    MYSQL_SIMPLE_JSON=false
    ```

    As we can see, the `MYSQL_SIMPLE_JSON` is set to `false` because `.Release.IsInstall` is false.

iromli: Our concern is how the JSON data formatted in persistence, not the way chart is upgraded/installed. In future version, we may introduce automatic upgrade from legacy to simple JSON data format. But as stated by Yuriy Movchan, prior to MySQL 8.4, the top-level indexing of JSON data might be slower.

iromli: I think to preserve backward-compatibility, we should set `cnSqlSimpleJson` attribute to `false` by default.

iromli: For second thought, we actually want to use simple JSON for new installation, so keeping `cnSqlSimpleJson: true` is probably desired.

moabu: But its tied.  Don’t we want it to be set to `false` for existing installations ? That means the flags do the job. However, what I’m hearing is that its version tied. We want the flag to be `true` for all installations and upgrades past 1.1.6. And we want it to be false for upgrading from <1.16 versions. 


Otherwise, my concern stems from that this flag can be atomic. It doesn’t make sense from the user experience side . Why should the user be concerned with this flag ? Thats what I’m trying to do. Because otherwise our values.yaml will become bloated over time.

iromli: we are now supporting MYSQL_SIMPLE_JSON via env var or auto-detect

moabu (approved):
Worked like a charm!

PR #10169, opened 2024-11-18
Jans cedaling issue 10151

olehbozhok:
### Prepare

-------------------

#### Target issue
  
  
  
  
closes #issue-number-here

#### Implementation Details
  

-------------------


mo-auto:
Error: Hi @olehbozhok, You did not reference an open issue in your [PR](10169). I attempted to create an [issue](#10170) for you.    
 Please update that [issues'](#10170) title and body and make sure I correctly referenced it in the above PRs body.

docs/cedarling/cedarling-policy-store.md
```diff
@@ -196,17 +202,82 @@ The Token Entity Metadata Schema defines how tokens are mapped, parsed, and tran
...
 ```
 
-- **role_mapping**: (String, *Optional*) Indicates which field in the token should be used for role-based access control. If not needed, set to an empty string (`""`).
-- **claim_mapping:** Defines how to extract and transform specific claims from the token. Each claim can have its own parser (`regex` or `json`) and type (`Acme::Email`, `Acme::URI`, etc.).
+#### Role mapping
 
+- **role_mapping**: (String OR Array of String, *Optional*) Indicates which field in the token should be used for role-based access control. If not needed, set to an empty string (`""`).
 
-**Note**: You can include a `role_mapping` in each token but only the first one that get parsed will be recognized by Cedarling. Cedarling parses the `role_mapping`s for each token in this order:
+You can include a `role_mapping` in each token but only the first one that get parsed will be recognized by Cedarling. Cedarling parses the `role_mapping`s for each token in this order:
```
nynymike, line 209: Why? If the id_token says roles are contained in the `role` claim, and the access_token says in the `member_of` claim, why can't the cedarling instantiate add'l roles?

nynymike (approved):
This looks pretty good to me. I think we need to test it to see.

djellemah (approved):
Looks ok to me.

docs/cedarling/cedarling-policy-store.md
olehbozhok: We never discussed that we should get roles from different tokens. And it looks little strange if different tokens generated for same person have different roles set...

PR #10331, opened 2024-12-04
refactor(jans-cedarling): improve WASM compatibility

### LINKED ISSUE #10330: refactor(jans-cedarling): improve WASM compatibility

Cedarling can't compile to WASM because of
1. the usage of `std::time`.
2. the usage of `reqwest::blocking::Client`

- The [`chrono`](https://crates.io/crates/chrono) crate could be used instead of `std::time` to fetch the time since it has WASM support.
- Use the `reqwest::Client` with the `tokio` runtime to remove the blocking api.

An alternative solution is to write our own interface to the JS Date API.

**Additional context**
N/A

rmarinn:
### Prepare

-------------------

### Description

This PR improves WASM compatibility by:
 - using the `chrono` crate which has WASM support instead of `std::time` for getting the time.
 - using replacing `reqwest::blocking::Client` with `reqwest::Client`.

#### Target issue

target issue: refactor(jans-cedarling): improve WASM compatibility [#10330](https://github.com/JanssenProject/jans/issues/10330)
  
closes #10330

#### Implementation Details

- the `HttpClient` create will store a reference to a `tokio` runtime which it will use to make async requests so compiling to WASM would be possible.

-------------------

### DIFF
--- a/jans-cedarling/cedarling/src/jwt/http_client.rs
+++ b/jans-cedarling/cedarling/src/jwt/http_client.rs
@@ -5,10 +5,14 @@
  * Copyright (c) 2024, Gluu, Inc.
  */
 
-use reqwest::blocking::Client;
-use std::{thread::sleep, time::Duration};
+use reqwest::Client;
+use serde::de::DeserializeOwned;
+use tokio::{
+    runtime::{Builder as RtBuilder, Runtime},
+    time::Duration,
+};
 
-/// A wrapper around `reqwest::blocking::Client` providing HTTP request functionality
+/// A wrapper around [`reqwest::Client`] providing HTTP request functionality
 /// with retry logic.
 ///
 /// The `HttpClient` struct allows for sending GET requests with a retry mechanism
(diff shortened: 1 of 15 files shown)


djellemah (commented):
Locally, there are build failures with `cargo test --target wasm32-wasip1` and `cargo test --target wasm32-unknown-unknown`. Is that expected right now?

Looks to me the cause is `jsonwebkey`, which pulls in an older version of `jsonwebtoken` which then pulls in an older version of `ring`.

rmarinn:
> Locally, there are build failures with cargo test --target wasm32-wasip1 and cargo test --target wasm32-unknown-unknown. Is that expected right now?

Yeah... turns out we need to change more things (other than the `jsonwebkey` issue) before it compiles correctly which is why i converted this to draft for now. 

> Looks to me the cause is jsonwebkey, which pulls in an older version of jsonwebtoken which then pulls in an older version of ring.

we're only really using `jsonwebkey` for tests so I'll probably just replace that when I get the time.

jans-cedarling/cedarling/src/jwt/http_client.rs
```diff
@@ -5,24 +5,63 @@
...
+            .map_err(HttpClientError::DeserializeJson)?;
+        Ok(resp_json)
+    }
+
+    /// Deserializes the response into a String.
+    pub fn text(self) -> Result<String, HttpClientError> {
+        let resp_text = self
+            .rt
+            .block_on(async { self.resp.text().await })
+            .map_err(HttpClientError::DeserializeJson)?;
```
abaghinyan, line 54: maybe it should be better to use `DeserializeText` instead of `DeserializeJson` here

jans-cedarling/cedarling/src/jwt/http_client.rs
```diff
@@ -5,24 +5,63 @@
...
 }
 
 impl HttpClient {
     pub fn new(max_retries: u32, retry_delay: Duration) -> Result<Self, HttpClientError> {
+        let rt = RtBuilder::new_current_thread()
+            .enable_all()
+            .build()
+            .expect("Failed to create Tokio runtime");
         let client = Client::builder()
             .build()
```
abaghinyan, line 66: It's a good practice to add a timeout in during an http request like: 
```
let client = Client::builder()
.timeout(Duration::from_sec(20))
.build()
.map_err(HttpClientError::Initialization)?;
```

jans-cedarling/cedarling/src/jwt/http_client.rs
```diff
@@ -5,24 +5,63 @@
...
+
+impl Response<'_> {
+    /// Deserializes the response into <T> from JSON.
+    pub fn json<T>(self) -> Result<T, HttpClientError>
+    where
+        T: DeserializeOwned,
+    {
+        let resp_json = self
+            .rt
+            .block_on(async { self.resp.json::<T>().await })
```
abaghinyan, line 44: In general it's better to use async then block_on. Here you can do something like :
```
let resp_json = self
            .resp
            .json::<T>()
            .await
            .map_err(HttpClientError::DeserializeJson)?;
        Ok(resp_json)
```
Moreover, you can remove the lifetime parameter <'rt> from Response.

olehbozhok:
Implemented in the https://github.com/JanssenProject/jans/pull/10542

PR #10422, opened 2024-12-14
feat(jans-cedarling): implement loading role from many JWT tokens

### LINKED ISSUE #10418: feat(jans-cedarling): implement loading role from many JWT tokens
In current implementation we can load role only from one JWT token.

In policy store we have field `role_mapping` for each JWT token.
All JWT tokens where mapping is defined should be mapped.

olehbozhok:
### Prepare

-------------------

### Description
In this PR was implemented loading role mapping from all tokens, if `role_mapping` was specified in policy store for token.
If none `role_mapping` defined the `Cedarling` will try to find role in `userinfo` token in field `role`.

#### Target issue
  
  [link](https://github.com/JanssenProject/jans/issues/10418)
  
  
  
closes #10418

#### Implementation Details
  
  Added loading role mapping from all tokens if `role_mapping` was specified in policy store for token.

-------------------


docs/cedarling/cedarling-policy-store.md
```diff
@@ -206,12 +206,8 @@ The Token Entity Metadata Schema defines how tokens are mapped, parsed, and tran
...
 - **role_mapping**: (String OR Array of String, *Optional*) Indicates which field in the token should be used for role-based access control. If not needed, set to an empty string (`""`).
 
-You can include a `role_mapping` in each token but only the first one that get parsed will be recognized by Cedarling. Cedarling parses the `role_mapping`s for each token in this order:
-
-1. `access_tokens`
-2. `id_tokens`
-3. `userinfo_tokens`
-4. `tx_tokens`
+You can include a `role_mapping` in each token, all of them will be executed by Cedarling.
+If none `role_mapping` defined the `Cedarling` will try to find role in `userinfo` token in field `role`.
```
rmarinn, line 210: Shouldn't Cedarling error if the `role` claims are different for each token instead of trying to authorize with each one? This seems insecure since they can effectively *try* multiple values for the `role` in each request -- and if the request should be denied using the `access_token` because of a wrong `role`, Cedarling could still allow the request to go through as long as the `role` is correct for the other tokens.

jans-cedarling/cedarling/src/tests/mapping_entities.rs
```diff
@@ -295,3 +296,80 @@ fn test_failed_userinfo_token_mapping() {
...
+    let roles_left = cedarling
+        .authorize_entities_data(&request)
+        .expect("should get authorize_entities_data without errors")
+        .role_entities
+        .into_iter()
+        .map(|entity| entity.uid().id().escaped())
+        // if role successfully removed from `expected_role_ids` we filter it
+        .filter(|uid| !expected_role_ids.remove(uid.as_str()))
+        .map(|v| v.to_string())
+        .collect::<Vec<_>>();
```
rmarinn, line 364: maybe it's just not clear to me, but can you explain why if an `access_token` is saying that it's role is a `Guest` and the `Userinfo` token is saying that it's an `Admin`, why is the request still approved? Isn't the information provided for request effectively conflicting?

olehbozhok: For me, it looks like we should get role only from one token, but Mike said that we should allow getting roles from any and many tokens.

Even if the information is conflicting the `cedar-policy` will correctly evaluate it.  Because if none of policy is evaluated than result Deny, Permit policy evaluated is Allow and if forbid evaluated result is denied (higher priority).
So it should be OK.

olehbozhok: "why is the request still approved?" because we have policy that permit

olehbozhok: and we have no forbid policy

docs/cedarling/cedarling-policy-store.md
olehbozhok: it is OK if in different tokens we have different roles otherwise it has no sense to have mapping role from many tokens

rmarinn: > it has no sense to have mapping role from many tokens

Correct me if I'm wrong but I thought that the reason why we enable mapping from multiple tokens is because some IDPs do not provide all the tokens. for example, some IDPs might only give ID tokens. Some might also only give Userinfo Tokens.

jans-cedarling/cedarling/src/tests/mapping_entities.rs
rmarinn: > Mike said that we should allow getting roles from any and many tokens.

Can you confirm if it means allowing Cedarling to authorize for multiple roles for through one request or and not because some IDPs won't have all the tokens all the time? Because it doesn't seem right for a User to have, for example, to have a "Guest" and "Admin" role at the same time for one request.

rmarinn: > Even if the information is conflicting the cedar-policy will correctly evaluate it.

This seems wrong... why would we allow a User which isn't sure of their identity to be authorized?

> Because if none of policy is evaluated than result Deny

This is right because we should Deny by default.

olehbozhok: "This is right because we should Deny by default." It is how cedar-policy work, We don't need to do anything special for this

docs/cedarling/cedarling-policy-store.md
olehbozhok: At first, I have implemented role mapping only from one token. @SafinWasi and @nynymike ask me to extend this to get roles from many tokens.

jans-cedarling/cedarling/src/tests/mapping_entities.rs
olehbozhok: `This seems wrong... why would we allow a User which isn't sure of their identity to be authorized?` we write `cedar-policy` rules (policies) to check if we permit or forbid.

olehbozhok: Also, a lot of systems allows users to have many roles simultaneously.

rmarinn: > Also, a lot of systems allows users to have many roles simultaneously.

yes, but not from different sources. why not make the "roles" claim for a token a list that contains all the role for that request instead of assigning different roles per token?

docs/cedarling/cedarling-policy-store.md
nynymike: I think we should do a UNION of all the roles found for signed tokens that we trust.  This is similar to the question of which tokens take precedence, for example if `surname` is found in both id_token and userinfo token with different values. Perhaps we can expose this as a bootstrap property: `CEDARLING_MAPPING_COLLISSION_HANDLING` = `join`, `replace`

nynymike: This is probably a corner case. In Jans Auth Server for example, the Userinfo and id_token claims should be the same (unless there is an interception script).

jans-cedarling/cedarling/src/tests/mapping_entities.rs
nynymike: I think this is a corner case that doesn't exist much in practice. At the end of the day, the Cedarling is open source and people can fork it for their esoteric use cases. One list of roles for a person is enough, e.g. we don't need the roles scoped to a token.

nynymike (dismissed):

rmarinn (dismissed):

nynymike (approved):

rmarinn (approved):

PR #10542, opened 2025-01-02
feat(jans-cedarling): add WASM bindings for Cedarling

### LINKED ISSUE #10013: feat(jans-cedarling): add WASM bindings for Cedarling
Feature Request: WASM Bindings for Cedarling

This feature request focuses on implementing WebAssembly (WASM) bindings for the Cedarling project, allowing it to run in a web environment.

The following steps are necessary to be able to compile Cedarling to WASM:

- [ ] Replace the `uuid7` crate with `uuid` configured with the `js` feature enabled for better WASM compatibility.
- [x] Refactor to use `reqwest::Client` in place of `reqwest::blocking::Client` to avoid blocking I/O calls.
- [x] Make `Cedarling::authz` asynchronous to support non-blocking initialization.
- [ ] Implement lazy loading for JWKS fetching to allow non-async initialization and improve startup efficiency in WASM environments.
- [ ] Include usage examples to demonstrate Cedarling’s WASM API in a browser environment.

olehbozhok:
### Prepare

-------------------

#### Target issue
  
  [link](https://github.com/JanssenProject/jans/issues/10013)
  
  
closes #10013

#### Implementation Details
  
- Add WASM binding 
- update memory log interface to return JSON and implement to WASM binding
- fix generation UUD7 for logging, created own function that use time from `chrono`, std time cause panic
- add unit test to work in browser
- implement simple html page that execute `Cedarling` in browser

#### Problems that should be fixed in future
- Some JSON logs entities serialize to `{}` empty JSON struct. It reproduces only in WASM. 

-------------------


nynymike (dismissed):

jans-cedarling/bindings/cedarling_python/pyproject.toml
```diff
@@ -11,7 +11,6 @@ classifiers = [
     "Programming Language :: Python :: Implementation :: CPython",
     "Programming Language :: Python :: Implementation :: PyPy",
 ]
```
rmarinn: There was a discussion here to make it dynamic here instead of static. so remove the
```toml
version = "0.0.0"
```
then bring back the
```toml
dynamic = ["version"]
```

jans-cedarling/bindings/cedarling_wasm/index.html
```diff
@@ -0,0 +1,412 @@
...
+    <script type="module">
+        import initWasm, { init, Cedarling } from "./pkg/cedarling_wasm.js";
+        async function main() {
+            await initWasm(); // Initialize the WebAssembly module
+            window.Cedarling = Cedarling;
+            window.cedarling_init = init;
+        }
+        main().catch(console.error);
+
+    </script>
```
rmarinn: Could you move the code only relating to initializing and using Cedarling to it's own file so it would be easier for users to understand the example usage?

jans-cedarling/bindings/cedarling_wasm/index.html
```diff
@@ -0,0 +1,412 @@
...
+    <!-- jQuery (necessary for Bootstrap's JavaScript plugins) -->
+    <script src="https://code.jquery.com/jquery-1.12.4.min.js"
+        integrity="sha384-nvAa0+6Qg9clwYCGGPpDQLVpLNn0fRaROjHqs13t4Ggj3Ez50XnGQqc/r8MhnRDZ"
+        crossorigin="anonymous"></script>
+    <!-- Include all compiled plugins (below), or include individual files as needed -->
+    <script src="https://cdn.jsdelivr.net/npm/bootstrap@3.4.1/dist/js/bootstrap.min.js"
+        integrity="sha384-aJ21OjlMXNL5UyIl/XNwTMqvzeRMZH2w8c5cRVpzpU8Y5bApTppSuUkhZXN0VxHd"
+        crossorigin="anonymous"></script>
+
+    <script>
```
rmarinn: Could you also move the contents of this whole script tag to it's own `.js` file? I think the constants could also be moved to it's own file so we can isolate the usage example on it's own. It's kinda difficult to understand what's going on if everything is in one big file.

jans-cedarling/bindings/cedarling_wasm/index.html
```diff
@@ -0,0 +1,412 @@
...
+
+        // // Payload of userinfo_token:
+        // // {
```
rmarinn: minor issue but the comments here is double-double slashes (`// //`). we can also use multi-line comments in JS.

jans-cedarling/cedarling/Cargo.toml
```diff
@@ -2,10 +2,19 @@
 name = "cedarling"
 version = "0.0.0-nightly"
 edition = "2021"
+description = "WASM binding to Cedarling"
```
rmarinn: This `Cargo.toml` file isn't for the WASM binding.

jans-cedarling/cedarling/Cargo.toml
```diff
@@ -2,10 +2,19 @@
...
+license = "Apache-2.0"
+
+[features]
+# blocking with default is used only for local testing, next line should be commented
+# default = ["blocking"]
```
rmarinn: Could you just explain what this means since I don't understand the comment. It sounds like if we're using local testing, we should uncomment this -- but i found that the tests run fine even if it's commented.

jans-cedarling/cedarling/src/http/mod.rs
```diff
@@ -1,40 +1,78 @@
...
-mod wasm;
-
-use std::time::Duration;
-
+/*
+* This software is available under the Apache-2.0 license.
+* See https://www.apache.org/licenses/LICENSE-2.0.txt for full text.
+*
+* Copyright (c) 2024, Gluu, Inc.
+cfg*/
```
rmarinn: there's a `"cfg"` at the end here

jans-cedarling/cedarling/Cargo.toml
```diff
@@ -2,10 +2,19 @@
 name = "cedarling"
 version = "0.0.0-nightly"
```
rmarinn, line 3: it's probably pretty late to ask this but shouldn't we also bumping the version here?

jans-cedarling/bindings/cedarling_wasm/index.html
```diff
@@ -0,0 +1,412 @@
...
+                let bootsrap_config_str = bootstrap_config_textarea.val();
+                let bootsrap_config_json = JSON.parse(bootsrap_config_str);
+
+                // init cedarling
+                const cedar_instance = await window.cedarling_init(bootsrap_config_json);
+                window.cedar_instance = cedar_instance;
+
+                let logs = cedar_instance.pop_logs();
+                if (logs.length != 0) {
+                    let pretty_logs = logs.map(log => JSON.stringify(Object.fromEntries(log), null, 2));
```
rmarinn: from the JSON.stringify() [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify):
> Only [enumerable own properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties) are visited. This means [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set), etc. will become "{}".

You can iterate through each item in the map then serialize them one-by-one. so the nested objects will be printed as well.
```suggestion
                    function serializeMap(value) {
                        if (value instanceof Map) {
                            // Convert the Map into an object where keys and values are recursively processed
                            return Object.fromEntries(
                                Array.from(value.entries()).map(([key, val]) => [key, serializeMap(val)])
                            );
                        } else if (Array.isArray(value)) {
                            // Recursively process arrays
                            return value.map(item => serializeMap(item));
                        } else if (value && typeof value === 'object') {
                            // Recursively process plain objects
                            return Object.entries(value).reduce((acc, [key, val]) => {
                                acc[key] = serializeMap(val);
                                return acc;
                            }, {});
                        }
                        // Return primitive values as-is
                        return value;
                    }

                    let pretty_logs = logs.map(log => {
                        if (log instanceof Map) {
                            // Serialize the Map into an object representation
                            return JSON.stringify(serializeMap(log), null, 2);
                        }
                        return JSON.stringify(log, null, 2); // Fallback for non-Map logs
                    });
```

rmarinn (changes requested):
Works well but the example is a bit messy to present to someone that will want try out Cedarling as-is.

Furthermore the docs need to be updated, especially since Cedarling is now async by default.

jans-cedarling/cedarling/Cargo.toml
olehbozhok: Looks like better to delete, to not confuse.
I wanted to say, `blocking` feature as default can be used in some local development or testing, but should be commented by default.

jans-cedarling/cedarling/Cargo.toml
olehbozhok: I have no idea...

olehbozhok: https://github.com/JanssenProject/jans/blob/main/jans-cedarling/cedarling/Cargo.toml
In main branch it is the same, so I assume if we need update version it will be changed in other PR

jans-cedarling/cedarling/Cargo.toml
olehbozhok: Updated

jans-cedarling/bindings/cedarling_python/pyproject.toml
olehbozhok: resolved

jans-cedarling/bindings/cedarling_wasm/index.html
olehbozhok: I think it will be better to create another one file with simple example

jans-cedarling/bindings/cedarling_wasm/index.html
olehbozhok: resolved

jans-cedarling/bindings/cedarling_wasm/index.html
olehbozhok: I will create another one file with simple example. 
And I think this file shouldn't be separate it makes easier to understand that this file (app) is one whole. In modern IDE you always can open same file in split screen mode to watch to different places of same file simultaneously.

jans-cedarling/cedarling/Cargo.toml
rmarinn: I see... i think they can just import Cedarling with the blocking feature if they really want blocking, just like how `reqwest` does it.

jans-cedarling/bindings/cedarling_wasm/index.html
rmarinn: hmmm...i guess it comes down to preference. personally, it's harder for me to understand if everything is on one file since the example also has UI code which isn't related to how to use Cedarling. I like to avoid splitting my screen and scrolling too much when reading a tutorial/example.

jans-cedarling/cedarling/Cargo.toml
rmarinn: yeah, probably in another PR. but we need to discuss it.

rmarinn (dismissed):

duttarnab (approved):

jans-cedarling/bindings/cedarling_wasm/cedarling_app.html
```diff
@@ -0,0 +1,239 @@
+<!doctype html>
+<html lang="en-US">
+
+<head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>Cedarling WASM App</title>
+
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@3.4.1/dist/css/bootstrap.min.css"
```
abaghinyan: Why using an oldest version of bootstrap ? 
Is there any compatibility problem ?

jans-cedarling/bindings/cedarling_wasm/cedarling_app.html
```diff
@@ -0,0 +1,239 @@
...
+        integrity="sha384-HSMxcRTRxnN+Bdg0JdbxYKrThecOKuH5zCYotlSAcp1+c8xmyTe9GYg1l9a69psu" crossorigin="anonymous">
+
+    <style>
+        textarea {
+            width: 100%;
+        }
```
abaghinyan: Bootstrap's grid system already handles responsiveness. Normally you don't need this type of tag.

jans-cedarling/bindings/cedarling_wasm/cedarling_app.html
```diff
@@ -0,0 +1,239 @@
+<!doctype html>
```
abaghinyan, line 7: maybe you should add the copyright in html files too 
```
<!--
    This software is available under the Apache-2.0 license.
    See https://www.apache.org/licenses/LICENSE-2.0.txt for full text.

    Copyright (c) 2024, Gluu, Inc.
-->
```

jans-cedarling/bindings/cedarling_wasm/example_data.js
```diff
@@ -0,0 +1,159 @@
+const BOOTSTRAP_CONFIG = {
```
abaghinyan, line 1: maybe you should add the copyright 
```
/*
 * This software is available under the Apache-2.0 license.
 * See https://www.apache.org/licenses/LICENSE-2.0.txt for full text.
 *
 * Copyright (c) 2024, Gluu, Inc.
 */

```

jans-cedarling/bindings/cedarling_wasm/index.html
```diff
@@ -0,0 +1,49 @@
+<!doctype html>
```
abaghinyan, line 7:
```
<!--
    This software is available under the Apache-2.0 license.
    See https://www.apache.org/licenses/LICENSE-2.0.txt for full text.

    Copyright (c) 2024, Gluu, Inc.
-->
```

jans-cedarling/cedarling/src/blocking.rs
```diff
@@ -0,0 +1,56 @@
...
+#[derive(Clone)]
+pub struct Cedarling {
+    runtime: Arc<Runtime>,
+    instance: AsyncCedarling,
+}
+
+impl Cedarling {
+    /// Builder
+    pub fn new(config: &BootstrapConfig) -> Result<Cedarling, InitCedarlingError> {
+        let rt = Runtime::new().unwrap();
```
abaghinyan: To avoid the panic, maybe you should do something like 
```
let rt = Runtime::new().map_err(|e| InitCedarlingError::from(e))?;
```

jans-cedarling/sparkv/src/lib.rs
```diff
@@ -228,10 +233,14 @@ mod tests {
     #[test]
     fn test_get_item_return_none_if_expired() {
         let mut sparkv = SparKV::new();
-        _ = sparkv.set_with_ttl("kkk", "value", std::time::Duration::from_millis(50));
+        _ = sparkv.set_with_ttl(
+            "kkk",
```
abaghinyan: you should put something different then this value

jans-cedarling/bindings/cedarling_wasm/cedarling_app.html
olehbozhok: resolved

jans-cedarling/bindings/cedarling_wasm/index.html
olehbozhok: Thanks it helped a lot!

jans-cedarling/bindings/cedarling_wasm/cedarling_app.html
olehbozhok: It was cloned from one of template from bootstrap site, fixed

jans-cedarling/bindings/cedarling_wasm/cedarling_app.html
olehbozhok: Yeah, it last bootstrap support it, updated

jans-cedarling/bindings/cedarling_wasm/example_data.js
olehbozhok: resolved, thanks

jans-cedarling/bindings/cedarling_wasm/index.html
olehbozhok: resolved, thanks

jans-cedarling/sparkv/src/lib.rs
olehbozhok: Would you like to use there `key`? I don't know if it changes something dramatically.
Actually this code exist in the [original](https://github.com/uzyn/sparkv/blob/main/src/lib.rs#L217).

I made changes related to the WASM

jans-cedarling/cedarling/src/blocking.rs
olehbozhok: resolved

olehbozhok: in https://github.com/JanssenProject/jans/pull/10542/commits/f759ad85417a9026432805a8c342a9947aef8c6e

jans-cedarling/sparkv/src/lib.rs
abaghinyan: just the letters 'kkk' can be misinterpreted

olehbozhok: resolved

jans-cedarling/sparkv/src/lib.rs
```diff
@@ -228,11 +233,15 @@ mod tests {
     #[test]
     fn test_get_item_return_none_if_expired() {
         let mut sparkv = SparKV::new();
-        _ = sparkv.set_with_ttl("kkk", "value", std::time::Duration::from_millis(50));
-        assert_eq!(sparkv.get("kkk"), Some(String::from("value")));
+        _ = sparkv.set_with_ttl(
+            "key",
+            "value",
+            Duration::new(0, 10000).expect("a valid duration"),
```
djellemah: Clearly your computer is faster than mine :-|

I get failures up to 40000 nanos - same value for sleep works. But even then it's intermittent - I'm guessing because of incremental compilation.

jans-cedarling/bindings/cedarling_wasm/README.md
```diff
@@ -0,0 +1,188 @@
...
+   */
+  pop_logs(): Array<any>;
+  /**
+   * Get specific log entry.
+   * Returns `Map` with values or `null`.
+   */
+  get_log_by_id(id: string): any;
+  /**
+   * Returns a list of all log ids.
+   * Returns `Array` of `String`
```
djellemah, line 104: Is it worth saying here "array might be empty" ie if the entry has expired since the call to `get_log_ids`.

jans-cedarling/bindings/cedarling_wasm/cedarling_app.html
```diff
@@ -0,0 +1,261 @@
...
+                input bootstrap config json
+            </div>
+
+            <!-- textarea with bootstrap config -->
+            <div class="row">
+                <!-- by default textarea contains example of bootstrap config, set by js -->
+                <textarea id="bootstrap_config_textarea" rows="10" cols="100"></textarea>
+            </div>
+
+            <!-- Button to init cedarling with defined above bootsrap config JSON -->
```
djellemah: bootsrap -> bootstrap

in some other places too

jans-cedarling/sparkv/src/lib.rs
olehbozhok: Resolved, increased value

jans-cedarling/bindings/cedarling_wasm/cedarling_app.html
olehbozhok: Fixed

djellemah (approved):

djellemah (approved):

nynymike (approved):

moabu (approved):

PR #10593, opened 2025-01-13
feat(jans-cedarling): Make SparKV use generics, and update MemoryLogger to use those.

### LINKED ISSUE #10554: feat (jans-cedarling): Update sparkv to use Generics
It would be great to update the API to use generics. This would allow using `serde_json::Value` directly as a storage value, avoiding the need to deserialize from a string when retrieving the value.

djellemah:
### Prepare

Depends on #10013 

-------------------

#### Target issue
  https://github.com/JanssenProject/jans/issues/10554
   
closes #10554 

#### Implementation Details
Make SparKV use a generic to specify the type to be stored.

Update `MemoryLogger` to make use of SparKV with generics.

Provide a `drain` iterator for a fast implementation of `MemoryLogger::pop_logs`.

Update `LogStorage` and `LogWriter` to use `serde_json::Value` internally.

-------------------


nynymike (approved):
Cool!

jans-cedarling/cedarling/src/log/memory_logger.rs
```diff
@@ -48,15 +46,21 @@ impl LogWriter for MemoryLogger {
             return;
         }
 
-        let json_string = serde_json::json!(entry).to_string();
+        let json = match serde_json::to_value(&entry) {
```
olehbozhok, line 87: We can use here `let json_string = serde_json::json!(entry);`

olehbozhok (changes requested):
I would like you to use `cargo fmt` formatter
```
cargo fmt -p sparkv
```

jans-cedarling/cedarling/src/log/memory_logger.rs
djellemah: The `serde_json::json!` macro expands to `::serde_json::to_value(&entry).unwrap()` in this case.

Also, it raises the question: how should the `return` case be handled?

olehbozhok: Trait `Loggable` require implementation `serde::Serialize`.
As I understand it means that we will never get error on calling `serde_json::to_value`. So we no need to cover error case

djellemah: `serde_json::to_value` calls `Serialize::serialize`
```
pub trait Serialize {
    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error>
}
```

Can we guarantee that every implementation (currently only `LogEntry` I think) of `log::interface::Loggable::serialize` will always return `Ok(...)`?

rmarinn: I don't think we can guarantee that every implementation will alwayrs return `Ok` since we might encounter an implementation that returns an error in the future. 

Though if the serialization of the log fails, shouldn't Cedarling just stop then print an error message since I don't think we want to keep running without logging.

djellemah: I think that's an important question - if there's a log failure, should Cedarling:

-  panic with an error, ie the default `unwrap()` behaviour

- display an error and keep going. Where to display the error? Does `stderr` even make sense in wasm or android?

olehbozhok: OK, I agree that If some type have manual implementation of `Serialize ` it can throw error.
If memory logger can't handle error it would be great to write error to `stderror`. I believe that on android on debugging developer can catch it. On WASM we need to write to browser console (You can find example in the `cedarling\src\log\stdout_logger\wasm_logger.rs`).

djellemah:
> I would like you to use `cargo fmt` formatter
> 
> ```
> cargo fmt -p sparkv
> ```

`cargo fmt` changes in #927ffc #5cad82

jans-cedarling/cedarling/src/log/memory_logger.rs
djellemah: changes for fallback logger to stdout/console are in #edc088e0

It re-uses the existing code from `LogStrategy`. Which in quite a bit of code. But the alternative is refactoring `wasm_logger::StdOutLogger`.

olehbozhok (approved):
I think `fallback` logger no need initialize every time. But in current context it doesn't really matter.

nynymike (approved):

Issue #10705, opened 2025-01-20
feat(jans-cedarling): improve getting logs from memory logger


olehbozhok:
**Is your feature request related to a problem? Please describe.**
To get single entry from memory logger we need to know ID of entry.
Currently, it is not possible without getting all logs.

Actually we need the way to get decision log exactly.

**Describe the solution you'd like**
We need to create `request_id` per one `authorize` request, propagate it to each log entry. And return it in the `AuthorizeResult` structure.

Also, we need to improve `sparkv`, add index with additional ID for entity and add `tag` for entity.

It will allow go get entities per request and filter per tag. 

**Describe alternatives you've considered**
to result structure add ID of log entry that show [decision-log](https://github.com/JanssenProject/jans/wiki/Cedarling-Nativity-Plan#decision-log)

**Additional context**

- generate request ID per single execution authorize call. And pass it to every log entry.
- currently request_id generate per log entry (because only one log entry use it)
- add to each log entry id and request_id (if was missed)
add to sparkv possibility to get log entries per request_id and request_id + some tag (in result will be id of log entry or list of ids)
add to result structure request_id
and eventually use updated api of memory logger that allows you to extract logs selectively using request_id or request_id + some tag

Issue #10591, opened 2025-01-10
feat(jans-cedarling): Support custom tokens and automatically populate principal entity attrs with token entity refs


rmarinn:
## **Is your feature request related to a problem? Please describe.**

When creating policies, users might need access to token entities. For instance, a token entity could be referenced in the attributes of a principal entity. Here's an example of how this might look in a schema:

```cedarschema
namespace Jans {
    entity Access_token = {
        jti: String,
    };
    entity Custom_token = {
        jti: String,
    };
    entity Workload = {
        client_id: String,
        access_token: Access_token,
        custom_token: Custom_token,
    };
}
```

To streamline this process, we aim to automate the creation of these token entities and ensure they are properly referenced within principal entities. However, this requires a mechanism to link token entities to their corresponding principals during entity creation.

Furthermore, we want this solution to support any custom tokens that the user might want.

### Supporting custom tokens

Currently, input tokens are passed into Cedarling using a map:

```js
input = { 
           "tokens": {
               "access_token": "...", 
               "id_token": "...", 
               "userinfo_token": "...", 
               "custom_token": "..." 
            },
           "resource": {"id": "12345", "type": "Ticket", "creator": "mike@gluu.org", "organization": "gluu"},
           "action": "View",
           "context": {
                       "ip_address": "54.9.21.201",
                       "network_type": "VPN",
                       "user_agent": "Chrome 125.0.6422.77 (Official Build) (arm64)",
                       "time": "1719266610.98636",
                      }
         }

decision_result = authz(input)
```

#### Token Name -> Entity Name mapping

We know that the `access_token`, `id_token`, and `userinfo_token` are token entities because of the `CEDARLING_MAPPING_ID_TOKEN`, `CEDARLING_MAPPING_ACCESS_TOKEN`, `CEDARLING_MAPPING_USERINFO_TOKEN` bootstrap entities. However, we might need more than these to support custom tokens.

Instead of separate boostrap property for each token, we can use a single object to hold all token to entity name mapping.

```js
CEDARLING_MAPPING_TOKEN_ENTITIES = {
    "access_token": "Jans::Access_token",
    "id_token": "Jans::id_token",
    "userinfo_token": "Jans::Userinfo_token",
    "custom_token": "SomeCompany::Custom_token"
}
```

#### Token Entity -> Principal Entity mapping

[...]

PR #10721, opened 2025-01-23
refactor(jans-cedarling): handle recoverable errors gracefully in authorization

### LINKED ISSUE #10589: refactor(jans-cedarling): handle recoverable errors gracefully in authorization
## **Is your feature request related to a problem? Please describe.**

Currently, Cedarling returns an error for scenarios like failing to create certain entities or parsing a request, even if these errors are recoverable. While this approach provides clarity in Rust (via explicit error handling), it complicates error management in bindings for other languages where error handling may not be as robust or idiomatic.

## **Describe the solution you'd like**

Update the `authorize` function to handle recoverable errors gracefully. Instead of returning an error, the function should:

1. Return a `DENY` decision in cases where recoverable errors occur.
2. Provide detailed feedback about the decision through structured fields:
   - **Principal Name**: Indicates the entity being authorized.
   - **Decision Result**: `ALLOW` or `DENY`.
   - **Warnings**: A list of messages for recoverable issues, e.g., failure to create a secondary entity like a `Role` that doesn’t block the primary authorization.
   - **Errors**: A list of critical issues that directly result in a `DENY` decision, e.g., failing to create the primary `User` entity.

An example for a serialized `ALLOW` result might be:
```json
{
    "results": [{"principal": "Jans::Workload", "decision": "ALLOW"}],
    "decision": "ALLOW",
}
```
Then for a `DENY` result:
```json
{
    "results": [{"principal": "Jans::User", "decision": "DENY"}],
    "decision": "DENY",
    "errors":  ["could not create User entity: could not get attribute value from payload: type mismatch for key 'email' expected: 'string' but found: 'number'"],
}
```
And for an `ALLOW` but with a warning:
```json
{
    "results": [{"principal": "Jans::User", "decision": "DENY"}],
    "decision": "DENY",
    "warnings":  ["could not create Role entity: could not get attribute value from payload: type mismatch for key 'role' expected: 'string' but found: 'number'"],    
}
```

### Handling Principal Combinations

To be able, to handle returning a decision result that is based on the combination of the authz result of different principals, we also return each of their results.

```json
{
    "decisions": [
        {"principal": "Jans::User", "decision": "DENY"}, 
        {"principal": "Jans::Workload", "decision": "ALLOW"}
    ],
    "decision": "DENY",
    "principal_operator": { /* JSON logic here */ },
    "warnings":  ["could not create Role entity: could not get attribute value from payload: type mismatch for key 'role' expected: 'string' but found: 'number'"],    
}
```

## **Describe alternatives you've considered**

~~An alternative approach would be to retain error returns for critical failures while ensuring that recoverable issues generate structured warnings or messages. However, Cedarling currently does not encounter critical errors that justify halting entirely.~~

EDIT: We now identified a potentially non-recoverable error (the logger failing).

## **Additional context**

Here’s the updated function signature stays the same in Rust but the Error should now only be non-recoverable errors.

```rs
impl Cedarling {
    pub fn authorize(&self, request: Request) -> Result<AuthorizeResult, Error> {
        // ...
    }
} 
```

See #10590 for the proposed improvement on the `"principal_operator"`.

rmarinn:
### Prepare

-------------------

### Description

This PR implements gracefully handling recoverable errors -- specifically, "bad input" errors where the issue is either a malformed or invalid input. In the case of a bad input, the call to `Cedarling::authorize` will not no longer return an error but will return an `Ok` with a `DENY` decision together with a reason.

#### Target issue
  
closes #10589 

#### Implementation Details

1. All the errors from the `AuthorizeError` enum that are just bad inputs and recoverable has been moved to a `BadInputErrorEnum`.

2. `AuthorizeResult` has been updated to this:

```rs
#[derive(Debug, Serialize)]
pub struct AuthorizeResult {
    #[serde(serialize_with = "serialize_opt_response")]
    pub workload: Option<cedar_policy::Response>,
    #[serde(serialize_with = "serialize_opt_response")]
    pub user: Option<cedar_policy::Response>,
    #[serde(
        serialize_with = "serialize_reason_input",
        skip_serializing_if = "Option::is_none"
    )]
    pub reason_input: Option<BadInputError>,
    pub decision: bool,
}
```

which looks like this when serialized to json: 

```json
{
  "workload": null,
  "user": null,
  "reason_input": "Invalid Access token: Failed to decode JWT from Base64: Invalid last symbol 48, offset 749.",
  "decision": false
}
```

-------------------


rmarinn:
try to merge #10591 before this one

jans-cedarling/cedarling/src/authz/mod.rs
```diff
@@ -268,93 +294,18 @@ impl Authz {
...
-                context: request.context.clone(),
-                resource: resource_uid.to_string(),
-                entities: entities_json,
-                person_authorize_info: user_authz_info,
-                workload_authorize_info: workload_authz_info,
-                authorized: result.decision,
-            })
-            .set_message("Result of authorize.".to_string()),
-        );
+        self.log_authz(LogAuthzArgs {
```
olehbozhok, line 297: Please, do not touch logging part, to avoid conflicts in the https://github.com/JanssenProject/jans/issues/10705

olehbozhok: If you wish to make this update, how about extract this changes to another branch? And I will merge it to my branch

jans-cedarling/cedarling/src/authz/mod.rs
```diff
@@ -472,6 +423,14 @@ impl AuthorizeEntitiesData {
...
 #[derive(thiserror::Error, Debug)]
 pub enum AuthorizeError {
+    /// Error encountered while logging the request
+    #[error("failed to log authorization request: {0}")]
+    Logging(#[from] LoggingError),
+}
+
+/// Error returned because of invalid or malformed inputs to [`Authz::authorize`]
+#[derive(thiserror::Error, Debug)]
+pub enum BadInputError {
```
olehbozhok, line 433: Does all this error can happen only because of bad input only?
If we can't check jwt because of problems on issuer page (can't load keys), it is also bad input?

jans-cedarling/cedarling/src/log/log_entry.rs
```diff
@@ -6,17 +6,15 @@
...
-use std::hash::Hash;
-
-use uuid7::Uuid;
-
 use super::LogLevel;
 use super::interface::Loggable;
 use crate::bootstrap_config::AuthorizationConfig;
 use crate::common::app_types::{self, ApplicationName};
 use crate::common::policy_store::PoliciesContainer;
+use std::collections::{HashMap, HashSet};
```
olehbozhok, line 14: You moved it manually or it is your IDE? I like when imports from std is separated =)

jans-cedarling/cedarling/src/authz/authorize_result.rs
```diff
@@ -5,23 +5,29 @@
...
     #[serde(serialize_with = "serialize_opt_response")]
-    pub person: Option<cedar_policy::Response>,
+    pub user: Option<cedar_policy::Response>,
+
+    /// Reasons why the authorization failed due to malformed inputs
+    #[serde(
+        serialize_with = "serialize_reason_input",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub reason_input: Option<BadInputError>,
```
olehbozhok, line 30: I think It should be `Vec` because it can be more that one `recoverable errors`.

And in the issue is said that for ` "decision": "DENY"` will be key with `"errors"` and for `"decision": "DENY"` will be key with `"warnings"`. 

It is planning?

jans-cedarling/cedarling/src/tests/mapping_entities.rs
```diff
@@ -182,13 +182,18 @@ async fn test_failed_user_mapping() {
 
     let request = REQUEST.clone();
 
-    let err = cedarling
+    let result = cedarling
         .authorize(request)
         .await
-        .expect_err("request should be parsed with mapping error");
+        .expect("request should be parsed succesfully");
+    assert_eq!(result.decision, false, "decision should be deny");
```
olehbozhok, line 189: Mistype
```suggestion
        .expect("request should be parsed successfully");
```

olehbozhok (dismissed):

jans-cedarling/cedarling/src/authz/mod.rs
rmarinn: The changes does not only involve moving the logging to another file, some of the inputs change as well. Thus, if you want to move the changes, you might as well merge this whole branch onto your branch. 

There will be conflicts no matter what, and we'll just have to deal with it.

Another option is to just merge main into your branch if this merges first.

Or I could also just wait for your PR then fix the conflicts if you want.

jans-cedarling/cedarling/src/log/log_entry.rs
rmarinn: I moved them together to consume less space.

jans-cedarling/cedarling/src/authz/mod.rs
rmarinn: > If we can't check jwt because of problems on issuer page (can't load keys), it is also bad input?

Getting the keys happens on startup and in the background/SSE (not implemented yet) so JWT validation wont fail because of an HTTP request. However, if a decoding key cannot be found to decode the JWT or if the JWT is expired, etc., it's under the BadInputError.

jans-cedarling/cedarling/src/authz/authorize_result.rs
rmarinn: > I think It should be `Vec` because it can be more that one recoverable errors.

We return immediately instead of going through the request the in the first recoverable error since there's no point continuing if for example the access token is malformed. In that context, is does `Vec` still make sense since i think it will always only have 1 item.

> And in the issue is said that for  "decision": "DENY" will be key with "errors" and for "decision": "DENY" will be key with "warnings".

Yeah, i just changed it from `"warnings"` to `"reason_input"` after looking at [authzen](https://openid.github.io/authzen/#name-sample-response-with-additi). It looks like they prefer using `"reason_*"` for adding additional context.

jans-cedarling/cedarling/src/log/log_entry.rs
rmarinn: I prefer moving then together then grouping them by visibility/usage. For example:

```rs
use std::collections::HashMap;
use std::str::FromStr;
use std::sync::Arc;
use trust_mode::*;

mod authorize_result;
mod build_ctx;
mod log_authz;
mod trust_mode;

pub(crate) mod entity_builder;
pub(crate) mod request;

pub use authorize_result::AuthorizeResult;
```

if we separate the std more, it would look like there's another separate group:
```rs
use std::collections::HashMap;
use std::str::FromStr;
use std::sync::Arc;

use trust_mode::*;

mod authorize_result;
mod build_ctx;
mod log_authz;
mod trust_mode;

pub(crate) mod entity_builder;
pub(crate) mod request;

pub use authorize_result::AuthorizeResult;
```
But then again, we don't even have a style-guide so i feel like whatever works.

jans-cedarling/cedarling/src/tests/mapping_entities.rs
rmarinn: fixed here: [7599ac7](https://github.com/JanssenProject/jans/pull/10721/commits/7599ac7014367887c03983380678c9f930381cfe)

jans-cedarling/cedarling/src/authz/authorize_result.rs
olehbozhok: Hmm, the task `handle recoverable errors gracefully in authorization`. As I understand it means that if we have error on build `role` (for example) or other entity, and we can continue to execute authorize request we should continue. And this kind of errors should be collected in the vector (and be logged).

olehbozhok: @nynymike what do you think?

nynymike: I agree with your statement, but I think the scope of the issue was beyond authz. I don't want the Cedarling to crash ever... it should keep running. Your web server never crashes... A crash should only happen if something really catastrophic happens. Right now it seems like there are too many reasons the PDP might just crash.
That was my goal here...

olehbozhok: @SafinWasi does `flask-sidecar` crash on bad input? Do you use `try ... expect ... ` statement like [here](https://github.com/JanssenProject/jans/blob/main/jans-cedarling/bindings/cedarling_python/tests/test_authorize.py#L201)?

SafinWasi: The sidecar does not crash on invalid input. It can catch cedarling exceptions and puts them in the response. 
https://github.com/JanssenProject/jans/blob/2c0cc6238544f32f022e7cd81cee927ad5c2230b/jans-cedarling/flask-sidecar/main/base/cedarling/cedarling.py#L121

rmarinn:
This will no longer be applied. We will be sticking to returning an error and using a try-catch to handle errors.

PR #10772, opened 2025-01-30
feat(jans-cedarling): Improve log searching and retrieval

### LINKED ISSUE #10705: feat(jans-cedarling): improve getting logs from memory logger

olehbozhok:
### Prepare

-------------------

#### Target issue
  
[link](https://github.com/JanssenProject/jans/issues/10705)
  
  
  
closes #10705

#### Implementation Details
  extended API of memory logger

    /// Get logs by tag, like `log_kind` or `log level`.
    /// Tag can be `log_kind`, `log_level`.
    fn get_logs_by_tag(&self, tag: &str) -> Vec<serde_json::Value>;

    /// Get logs by request_id.
    /// Return log entries that match the given request_id.
    fn get_logs_by_request_id(&self, request_id: &str) -> Vec<serde_json::Value>;

    /// Get log by request_id and tag, like composite key `request_id` + `log_kind`.
    /// Tag can be `log_kind`, `log_level`.
    /// Return log entries that match the given request_id and tag.
    fn get_logs_by_request_id_and_tag(&self, request_id: &str, tag: &str)
    -> Vec<serde_json::Value>;

-------------------


jans-cedarling/cedarling/src/log/memory_logger.rs
```diff
@@ -131,15 +144,50 @@ impl LogStorage for MemoryLogger {
...
+    ) -> Vec<serde_json::Value> {
+        let key = composite_key(request_id, tag);
+
+        self.storage
+            .lock()
+            .expect(STORAGE_MUTEX_EXPECT_MESSAGE)
+            .get_by_index_key(&key)
+            .map(|v| v.to_owned())
+            .collect()
+    }
```
rmarinn, line 179: shouldn't we return errors here instead of panicking?

jans-cedarling/cedarling/src/log/memory_logger.rs
```diff
@@ -94,14 +104,17 @@ impl LogWriter for MemoryLogger {
             },
         };
 
-        let set_result = self
-            .storage
-            .lock()
-            .expect(STORAGE_MUTEX_EXPECT_MESSAGE)
-            .set(&entry.get_request_id().to_string(), json);
+        let mut storage = self.storage.lock().expect(STORAGE_MUTEX_EXPECT_MESSAGE);
```
rmarinn, line 107: isn't it better if we return an error instead of calling `.expect()` in case the user wants some to implement a tear-down sequence in their code in case of a poisoned lock?

jans-cedarling/sparkv/src/index.rs
```diff
@@ -0,0 +1,91 @@
...
+
+/// HashMapIndex allows to group keys (to origin value) by index key.
+/// And get it efficiently.
+pub(crate) struct HashMapIndex {
+    buckets: HashMap<IndexKey, HashSet<ValueKey>>,
+    // is used to track what `IndexKey` is used for `ValueKey`
+    index_tracker: HashMap<ValueKey, HashSet<IndexKey>>,
+}
+
+impl HashMapIndex {
```
rmarinn, line 24: Are you sure you don't want to add tests for this implementation to ensure correctness?

olehbozhok: hmm, it is covered by tests in memory logger, I will think about that

jans-cedarling/cedarling/src/log/memory_logger.rs
olehbozhok: From my point of view poisoned lock is not recoverable state of program, so returning error has no sense, because this error will newer be fixed. And we need reload program to fix.

Also in our case storage `lock` can newer be poisoned. 
I think using `panicking` instead of return error **(in this case)** is a good tradeoff. Because It will newer happen.

jans-cedarling/cedarling/src/log/memory_logger.rs
olehbozhok: Could you describe how lock can be poisoned (in our case)?

jans-cedarling/sparkv/src/index.rs
olehbozhok: added in https://github.com/JanssenProject/jans/pull/10772/commits/bec069ceac47b5c191aa129ae65b1df341d4876d

SafinWasi (dismissed):
Python side looks good to me

jans-cedarling/cedarling/src/log/memory_logger.rs
rmarinn: > From my point of view poisoned lock is not recoverable state of program, so returning error has no sense

My concern if the user of the Cedarling library wants to gracefully handle the situation. For example, what if the system they're building shouldn't shut down unexpectedly without logging the current state and the error? 

> Also in our case storage lock can newer be poisoned.

Im not sure about that since we're writing bindings to other languages that do not implement Rust's safety guarantees.

But if you think that will still never happen, then go ahead.

jans-cedarling/cedarling/src/log/memory_logger.rs
rmarinn: In the bindings code for other languages, code might panic in undefined ways.

It probably won't happen in the rust code but I don't think we can guarantee that in the bindings code. I think we already saw that problem in the Kotlin bindings where the control over memory is limited.

jans-cedarling/cedarling/src/log/memory_logger.rs
olehbozhok: Hmm, in case if `target` does not support `atomic` it will panic anyway and returning `PoisonError` does not help. And in our situation poisoning is impossible. Because we don't use some tricky recursive function calling that tries to call same `lock`. So I'm pretty sure about that.

jans-cedarling/cedarling/src/log/memory_logger.rs
olehbozhok: Responded here
https://github.com/JanssenProject/jans/pull/10772#discussion_r1939457790

nynymike (approved):

olehbozhok:
Sorry, I slightly messed up the commits, but everything is OK now.
I’d appreciate it if you could approve it again.

jans-cedarling/cedarling/src/log/memory_logger.rs
rmarinn: alright, then it's ok for me

rmarinn (approved):

duttarnab (approved):

PR #10812, opened 2025-02-07
docs(jans-cedarling): shorten conditional

### LINKED ISSUE #10810: docs(jans-cedarling): update krakend readme for scope based access
-------------------

### Description

Update docs for the krakend plugin to show scope based access

-------------------

### Prepare

------------------

### Identified code changes

N/A

-------------------

N/A

-------------------

### Document the changes

- [x] Update readme
- [x] Update doc

SafinWasi:
### Prepare

-------------------

#### Target issue
  
  
  
  
closes #10810

#### Implementation Details
  

-------------------
N/A


docs/cedarling/cedarling-krakend.md
```diff
@@ -100,7 +100,7 @@ To test the plugin, you will need:
       resource is gatewayDemo::HTTP_Request
     )
     when {
-    ((principal has "access_token") && ((principal["access_token"]) has "scope")) && (((principal["access_token"])["scope"]).contains("profile"))
+        ((principal["access_token"])["scope"]).contains("profile")
```
olehbozhok:
```
You must check for presence of nested, optional attributes one layer at a time. For example, to check for the presence of principal.custom.project where both custom and project are optional, you must first check if principal has a custom attribute and then check that it principal.custom has a project attribute:
```
https://docs.cedarpolicy.com/policies/syntax-operators.html#operator-has

olehbozhok (changes requested):
Looks like you don't need to remove attribute `has` because it can cause error in evaluating this rule

olehbozhok (approved):
I don't agree to make the rule  shorten, but approve if Michael agree

nynymike:
Let's consider ergonomic syntax... is your policy more readible? Is it much faster? If so, how much? If it's not more readible, or faster, what is the advantage? Please don't just say "I don't like it"--give me a reason please.

syntrydy:
@nynymike 

1. It violates Cedar's fundamental safety principles by making assumptions about the data structure without verification. In Cedar, when we access optional attributes using bracket notation like ` principal["access_token"]` , we're essentially saying "I'm certain this attribute exists" - but we can't actually be certain. 
This is particularly dangerous in authorization policies where unexpected attribute access could lead to security vulnerabilities.

1.  When we chain attribute access like ` ((principal["access_token"])["scope"])` , we're creating a cascade of potential failure points. If either access_token or scope is missing, this could lead to a runtime error. This is especially problematic because Cedar policies need to be reliable and deterministic - a policy failure due to unsafe attribute access could result in either incorrectly granting or denying permissions, both of which are serious security concerns.

1. Cedar deliberately requires explicit handling of optional attributes using the ` has`  operator to force policy authors to think about and handle edge cases. This PR goes  against Cedar's philosophy of making attribute access safety explicit rather than implicit. This makes the policy more fragile and harder to maintain because it's not clear from reading the code whether these attributes are optional or required.

SafinWasi:
According to cedar developers, the ergonomic way to do this is to use dot notation instead of square bracket syntax. By using the [extended RHS syntax of the `has` operator](https://docs.cedarpolicy.com/policies/syntax-operators.html#operator-has), we can implicitly test the existence of both `access_token` and `scope` before the second conditional is evaluated.
```
@id("allow_one")
permit(
  principal is Jans::Workload,
  action == Jans::Action::"GET",
  resource is Jans::HTTP_Request
)
when {
    principal has access_token.scope &&
    principal.access_token.scope.contains("profile")
};
```

On cedar-policy-cli 4.3.1:
```
$ cedar validate --schema cedarling_core.cedarschema --policies policy.cedar
  ☞ policy set validation passed
  ╰─▶ no errors or warnings
```

olehbozhok (approved):

nynymike (approved):
Looks better!

PR #10833, opened 2025-02-12
refactor(jans-cedarling): embed the pdp_id and appplication_id into the logger

### LINKED ISSUE #10832: refactor(jans-cedarling): put the pdp_id and application_id into the logger

Creating log entries is a pain since we need to pass in the pdp_id and application_id into the logger for every entry.

```rs
truct LogEntry {
    pub(crate) fn new_with_data(
        pdp_id: app_types::PdpID,
        application_id: Option<app_types::ApplicationName>,
        log_type: LogType,
        request_id: Option<Uuid>,
    ) -> LogEntry {

        //...
    }
}
```

Since we only ever expect to have one pdp_id and application name for every instance of cedarling, we can put the pdp_id and application name inside the logger so that creating new log entries will be easier since we no longer need to pass those two in every time we want to create a new entry.

N/A

**Additional context**
N/A

rmarinn:
### Prepare

-------------------

### Description

Since the pdp_id and application name doesn't change per Cedarling instance, we can just embed them inside the logger so that creating new log entries will be easier since we no longer need to pass those two in every time we want to create a new entry.

#### Target issue

closes #10832 

#### Implementation Details

Initializing the logger will now need the pdp_id and application name. Every time a new log entry will be stored, the logger will add those two fields into the entry before storing the log.

-------------------


jans-cedarling/cedarling/src/lib.rs
```diff
@@ -95,28 +95,32 @@ impl Cedarling {
 
     /// Create a new instance of the Cedarling application.
     pub async fn new(config: &BootstrapConfig) -> Result<Cedarling, InitCedarlingError> {
-        let log = log::init_logger(&config.log_config);
         let pdp_id = app_types::PdpID::new();
+        let app_name = config
+            .application_name
+            .is_empty()
+            .then(|| ApplicationName(config.application_name.clone()));
```
olehbozhok, line 100: Looks like it will be cloned only if `application_name` is empty

https://doc.rust-lang.org/std/primitive.bool.html#method.then

jans-cedarling/cedarling/src/log/interface.rs
```diff
@@ -85,6 +88,9 @@ pub(crate) trait Loggable: serde::Serialize + Indexed {
             true
         }
     }
+
+    /// Serializes the entry to a JSON [`serde_json::Value`] together with the given client info
+    fn to_json_with_client_info(self, pdp_id: &PdpID, app_name: &Option<ApplicationName>) -> Value;
```
olehbozhok: Probably we can make default implementation, like for `fn can_log`.
It is up to you.

olehbozhok (changes requested):
Instead of implementing same in each logger we can do it only once in `LogStrategy`. 

Actually 
```
        pdp_id: PdpID,
        app_name: Option<ApplicationName>,
```
we need for `lock` logger.

How about next approach?
```
pub(crate) struct LogStrategy {
    log_writer: LogWriter,
    pdp_id: PdpID,
    app_name: Option<ApplicationName>,
}

pub(crate) enum LogWriter {
    Off(NopLogger),
    MemoryLogger(MemoryLogger),
    StdOut(StdOutLogger),
}
```


and add values in next part of code:
```
impl LogWriter for LogStrategy {
    fn log_any<T: Loggable>(&self, entry: T) {
    ....
```

jans-cedarling/cedarling/src/lib.rs
rmarinn: fixed here: [3a8ead3](https://github.com/JanssenProject/jans/pull/10833/commits/3a8ead3affbbfee498c364cf9aae0332ecbfee93)

jans-cedarling/cedarling/src/log/interface.rs
rmarinn: i think [97157cd](https://github.com/JanssenProject/jans/pull/10833/commits/97157cdeebfaf3c1c2c12bcfcc1514d66f0b94fc) resolves this

nynymike (approved):

jans-cedarling/cedarling/src/log/log_strategy.rs
```diff
@@ -20,25 +30,97 @@ pub(crate) enum LogStrategy {
...
+
+    #[cfg(test)]
+    pub fn logger(&self) -> &LogStrategyLogger {
+        &self.logger
+    }
+}
+
+#[derive(Serialize, PartialEq)]
+pub(crate) struct LogEntryWithClientInfo<Entry: Loggable> {
+    entry: Entry,
```
olehbozhok, line 70: Do we need to use `#[serde(flatten)]` macro for `entry`?

Also, usually generics marks as one capital character, but it would be great to mark at least as `ENTRY`, or `E`.

Also, please check how `LogEntryWithClientInfo` converts to the json value, if `entry` has field `pdp_id` or `app_name`. If `json!(...)` macro panic in this case we need gracefully handle it (maybe skip using `LogEntryWithClientInfo` in this case)

olehbozhok: I see you use safe implementation
```
   fn to_value(&self) -> serde_json::Value {
        let mut json = json!(self.entry);
        json["pdp_id"] = json!(self.pdp_id);
        if let Some(app_name) = self.app_name.as_ref() {
            json["application_id"] = json!(app_name);
        }
        json
    }
```

So all should be fine

olehbozhok (approved):

jans-cedarling/cedarling/src/log/log_strategy.rs
rmarinn: flatten implemented in: [9b503a6](https://github.com/JanssenProject/jans/pull/10833/commits/9b503a63b55e0857a290f797ae3ad2dba570a43f)
then added structure for error logs when failing to deserialize in: [6308000](https://github.com/JanssenProject/jans/pull/10833/commits/6308000fe20f256e5fa7d6868efb17c8a4d412cd)

olehbozhok:
@rmarinn What do you think about this implementation?
```
#[derive(PartialEq)]
pub(crate) struct LogEntryWithClientInfo<Entry: Loggable> {
    entry: Entry,
    pdp_id: PdpID,
    app_name: Option<ApplicationName>,
}

/// Safe implementation of serializer, newer throw error
impl<Entry: Loggable> Serialize for LogEntryWithClientInfo<Entry> {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: serde::Serializer,
    {
        let mut json_val = match serde_json::to_value(&self.entry) {
            Ok(v) => v,
            Err(err) => {
                let err_msg = format!("failed to serialize log entry to JSON: {err}");

                // Or `ErrorLogEntry`
                let value = json!({
                    "error": err_msg,
                });

                return value.serialize(serializer);
            },
        };

        json_val["pdp_id"] = serde_json::json!(self.pdp_id);
        if let Some(app_name) = self.app_name.as_ref() {
            json_val["application_id"] = json!(app_name);
        }
        json_val.serialize(serializer)
    }
}
```

rmarinn:
> @rmarinn What do you think about this implementation?
> 
> ```
> #[derive(PartialEq)]
> pub(crate) struct LogEntryWithClientInfo<Entry: Loggable> {
>     entry: Entry,
>     pdp_id: PdpID,
>     app_name: Option<ApplicationName>,
> }
> 
> /// Safe implementation of serializer, newer throw error
> impl<Entry: Loggable> Serialize for LogEntryWithClientInfo<Entry> {
>     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
>     where
>         S: serde::Serializer,
>     {
>         let mut json_val = match serde_json::to_value(&self.entry) {
>             Ok(v) => v,
>             Err(err) => {
>                 let err_msg = format!("failed to serialize log entry to JSON: {err}");
> 
>                 // Or `ErrorLogEntry`
>                 let value = json!({
>                     "error": err_msg,
>                 });
> 
>                 return value.serialize(serializer);
>             },
>         };
> 
>         json_val["pdp_id"] = serde_json::json!(self.pdp_id);
>         if let Some(app_name) = self.app_name.as_ref() {
>             json_val["application_id"] = json!(app_name);
>         }
>         json_val.serialize(serializer)
>     }
> }
> ```

I don't like that we're hiding how the error log entry is created from the caller. If we were to look at the place were `to_value` will be called we would probably have a `.expect` or `.unwrap` but the implementation of how the error log was generated will be opaque to the caller and they will have to look how `Serialize` was implemented.

rmarinn:
> I believe that we can make error fallback only once in `log_strategy.rs`, instead of implement the same in each logger, but I can be wrong.

I'd rather have some duplicate code than go for a wrong abstraction since the former is easier to fix.

PR #10908, opened 2025-02-20
fix: scope not getting added if dropdown is not used

### LINKED ISSUE #10899: fix(jans-tarp): scope not getting added if dropdown is not used
**Describe the bug**
During DCR with tarp, if the dropdown is not clicked on, the scope does not get added to the payload string.

**To Reproduce**
1. Install Tarp
2. Do DCR. In the scope field, type `openid` and hit Enter. Then type `profile` and hit enter **without clicking the drop down**.
3. Verify absence of `profile` from both payload and client in IDP

**Expected behavior**
Tarp should allow scopes to be typed in without clicking the dropdown

duttarnab:
closes #10899


rizarsyi (approved):
Looks good

syntrydy (approved):

PR #11035, opened 2025-03-12
 feat(jans-cedarling): New interface: authorizeUnverified - send Principal as Object v. JWT

### LINKED ISSUE #11000: feat(jans-cedarling): New interface: `authorizeUnverified` - send Principal as Object v. JWT
It would be nice to have a way to send the Cedar Principal as a JSON object, similar to how we send the Resource. This would be handy for developers who don't use JWTs for identity. Of course the identity can't be trusted ... it's asserted by the application. 

We create additional method to execute authorize request with new functionality:
method name `authorize_classic` - classic means that it is usual authorize method that can be executed using `cedar` CLI app.

here is example of request (for WASM)

Image

Alternatively, we can extend current structure
https://github.com/JanssenProject/jans/blob/main/jans-cedarling/bindings/cedarling_wasm/example_data.js#L123
with adding key `principal`

When `token` key is present we make evaluation using tokens.

If key `token` absent and `principal` present. We make evaluation using custom principal.

**Additional context**
NA

olehbozhok:
### Prepare

-------------------

### Description
added support principals unverified by tokens

#### Target issue
  
[link](https://github.com/JanssenProject/jans/issues/11000)
  
  
  
closes #11000

#### Implementation Details
  
- added method `authorize_unverified` that works like `authorize` but have another structure for request
```
pub struct RequestUnverified {
    /// Contains the JWTs that will be used for the AuthZ request
    pub principals: Vec<EntityData>,
    /// cedar_policy action
    pub action: String,
    /// cedar_policy resource data
    pub resource: EntityData,
    /// context to be used in cedar_policy
    pub context: serde_json::Value,
}
```

-------------------


jans-cedarling/cedarling/src/authz/authorize_result.rs
```diff
@@ -83,28 +115,70 @@ impl AuthorizeResult {
...
+            person_typename,
+            request_id,
+        )
+    }
+
+    /// Builder function for AuthorizeResult
+    pub(crate) fn new_for_many_principals(
+        principal_bool_operator: &JsonRule,
+        // key is typename
+        principal_responses: HashMap<SmolStr, cedar_policy::Response>,
```
rmarinn: It might be better to just define a type alias instead of using comments to describe the input. 

```rs
type TypeName = SmolStr;
```

Apparently [type aliases are just syntactic sugar](https://stackoverflow.com/questions/47872625/what-is-the-difference-between-a-type-alias-and-a-tuple-struct) so it's fine to have a lot of these defined on a per-file or module basis.

```suggestion
        principal_responses: HashMap<TypeName, cedar_policy::Response>,
```

jans-cedarling/cedarling/src/lib.rs
```diff
@@ -133,6 +134,15 @@ impl Cedarling {
         self.authz.authorize(request).await
     }
 
+    /// Authorize request with unverified data.
+    /// makes authorization decision based on the [`RequestUnverified`]
+    pub async fn authorize_unverified(
```
rmarinn: I'm not sure if `unverified` is the right term to use here. Technically, the inputs on the other `authorize` function is also unverified since it's Cedarling that does the verification, right?

If this interface isn't just for testing and we expect people to regularly use this, what if we just rename both function?
- `Cedarling::authorize` -> `Cedarling::authorize_jwt`
- `Cedarling::authorize` -> `Cedarling::authorize_cedar`

if people would just want to use the `authorize_cedar`, why didn't they just use the `cedar_policy` crate? 🤔

jans-cedarling/cedarling/src/authz/request.rs
```diff
@@ -14,15 +14,28 @@ pub struct Request {
...
     /// context to be used in cedar_policy
     pub context: serde_json::Value,
 }
 
-/// Cedar policy resource data
+/// Box to store authorization data, with any additional principals
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct RequestUnverified {
+    /// Contains the JWTs that will be used for the AuthZ request
+    pub principals: Vec<EntityData>,
```
rmarinn, line 77: Have you considered that approach where instead of having two structs, we would just have one `Request` struct then have an `enum` for the principals instead?

```rs
pub enum RequestPrincipal {
    Tokens(HashMap<String, String>),
    Cedar(Vec<EntityData>),
}
```

```suggestion
    /// Contains the principals that will be used for the AuthZ request
    pub principals: RequestPrincipal,
```

In the target issue, you mentioned that you opted to have two functions instead of one because of the [Interface Segregation Principal (ISP)](https://reflectoring.io/interface-segregation-principle/). From what I read, the essence of ISP is:

>  a client should not be exposed to methods it doesn’t need.

If we have two functions, and a client only want to use the JWT ***or*** the Cedar inputs but ***not both***, aren't we essentially exposing them two at least one method they don't need?

jans-cedarling/cedarling/src/authz/mod.rs
```diff
@@ -311,8 +308,171 @@ impl Authz {
                     context: request.context.clone(),
                     resource: resource_uid.to_string(),
                     entities: entities_json,
-                    person_authorize_info: user_authz_info,
-                    workload_authorize_info: workload_authz_info,
+                    authorize_info: vec![user_authz_info, workload_authz_info]
```
rmarinn: No need to initialize a `Vec` here only to turn it into an iterator then collect it onto a new `Vec`.

```suggestion
                    authorize_info: [user_authz_info, workload_authz_info]
```

jans-cedarling/cedarling/src/lib.rs
rmarinn: Though instead of using two functions, I prefer using an `enum` for the principal instead as i said in [this comment](https://github.com/JanssenProject/jans/pull/11035#discussion_r1992852258).

jans-cedarling/cedarling/src/authz/request.rs
rmarinn: I think having a single function would also simplify the error hierarchy.

```rs
pub enum AuthorizeError {
    // These two variants wraps `cedar_policy::RequestValidationError` but for `Workload` and `User` respectively
    #[error("The request for `Workload` does not conform to the schema: {0}")]
    WorkloadRequestValidation(cedar_policy::RequestValidationError),
    #[error("The request for `User` does not conform to the schema: {0}")]
    UserRequestValidation(cedar_policy::RequestValidationError),

    // This variant wraps `cedar_policy::RequestValidationError` but for any entity type
    #[error(transparent)]
    UnverifiedPrincipalRequestValidation(#[from] UnverifiedPrincipalRequestValidationError),
}

#[derive(Debug, derive_more::Error, derive_more::Display)]
#[display("The request for `{principal}` does not conform to the schema: {err}")]
pub struct UnverifiedPrincipalRequestValidationError {
    /// Principal name
    principal: String,
    /// Error value
    err: cedar_policy::RequestValidationError,
}
```

If we just have one interface, we can have a single error for building principals.

```rs
pub enum AuthorizeError {
    #[error(transparent)]
    InvalidPrincipal(InvalidPrincipal)
}

#[derive(Debug, derive_more::Error, derive_more::Display)]
#[display("the request for `{principal}` does not conform to the schema: {err}")]
pub struct InvalidPrincipal {
    /// Principal name
    principal: String,
    /// Error value
    err: cedar_policy::RequestValidationError,
}
```

olehbozhok: the naming was discussed [here](https://github.com/JanssenProject/jans/issues/11000#issuecomment-2705101286). Maybe you prefer suffix `unsigned`?

Your Idea with enum
```
pub enum RequestPrincipal {
    Tokens(HashMap<String, String>),
    Cedar(Vec<EntityData>),
}
```
is OK, but it can be almost impossible to implement on other languages.


___
about error hierarchy, yeah, we can rename `UnverifiedPrincipalRequestValidationError` to `InvalidPrincipal` and remove  `WorkloadRequestValidation` and  `UserRequestValidation`

rmarinn: > the naming was discussed here. Maybe you prefer suffix unsigned?

yeah, `authorize_unsigned` is better.

> is OK, but it can be almost impossible to implement on other languages.

hmm... that's true. i'm thinking the other languages should just send a map then our bindings will process it but yeah... having two functions would probably the simpler way to go about it.

jans-cedarling/cedarling/src/authz/authorize_result.rs
olehbozhok: resolved f4409a8f6d3f72d75f63d9937a051418ce73b75b

jans-cedarling/cedarling/src/authz/mod.rs
olehbozhok: resolved 1555d1dd3a697e045fdba68fb57179bc690b529a

jans-cedarling/cedarling/src/lib.rs
olehbozhok: discussion was finished [here](https://github.com/JanssenProject/jans/pull/11035#discussion_r1993978846)

jans-cedarling/cedarling/src/authz/request.rs
olehbozhok: method renamed 6d6e885fb6f1b2d73d605e143cddb168fc486aec

olehbozhok: Updated handling error 66127932bf33c5f657acb59185a96b69d017cbfe

nynymike:
Did we change the name from `Unverified` to `Unsigned` ?

olehbozhok:
> Did we change the name from `Unverified` to `Unsigned` ?

@nynymike yes, you gave 2 variants [here](https://github.com/JanssenProject/jans/issues/11000#issuecomment-2705101286) and for @rmarinn is more suitable `unsigned`  [link](https://github.com/JanssenProject/jans/pull/11035#discussion_r1993978846). 
I have no objections.

jans-cedarling/bindings/cedarling_python/PYTHON_TYPES.md
```diff
@@ -193,7 +216,7 @@ Attributes
 ----------  
 :param tokens: A class containing the JWTs what will be used for the request.  
 :param action: The action to be authorized.  
-:param resource: Resource data (wrapped `ResourceData` object).  
+:param resource: Resource data (wrapped `EntityData` object).
```
SafinWasi, line 219: `example.py` and the sidecar are still using `ResourceData`

SafinWasi (changes requested):
Both `example.py` and the sidecar need to be updated to use `EntityData` instead of `ResourceData`. In addition, cedarling docs need to be updated. We can add a section in `cedarling-authz.md` with info on unverified principal usage.

jans-cedarling/bindings/cedarling_python/PYTHON_TYPES.md
olehbozhok: resolved

SafinWasi (approved):

moabu (approved):

Issue #10795, opened 2025-02-05
feat(jans-edarling): implement lock server integration


rmarinn:
**Is your feature request related to a problem? Please describe.**

Cedarling needs to be able to integrate with the lock server as described in steps 10-20. In this diagram:

![Image](https://github.com/user-attachments/assets/560cd9c5-6558-466d-bd0e-651d59e72d6c) 

**Describe the solution you'd like**

The connection during the startup sequence can be implemented as follows:

1. Check if the `CEDARLING_LOCK` bootstrap property is set to `"enabled"`
2. Validate the `CEDARLING_LOCK_SSA` JWT.
3. Send a `POST` request to the IDP's `/register` endpoint
4. Recieve a 201 response together with a `client_id`
5. Get an access token via the `/token` endpoint
6. Get the policy store through the `/config` endpoint together with the access_token scoped to `"cedarling"`.
7. `POST` to the `/audit` endpoint
8. Subscribe to the SSE through the `/lock_sse`
9. Listen to SSE

**Describe alternatives you've considered**

N/A

**Additional context**

Using the Lock server is optional and could be toggled using the `CEDARLING_LOCK` bootstrap property.

olehbozhok:
@dagregi, please, verify if the subissues are implemented.

dagregi:
https://github.com/JanssenProject/jans/issues/10849 and https://github.com/JanssenProject/jans/issues/10850 are not done maybe we can do them after finishing https://github.com/JanssenProject/jans/pull/13237

PR #10564, opened 2025-01-07
ci: enhance the security of gh worflows

moabu:
### Prepare

- [ ] Read [PR guidelines](https://github.com/JanssenProject/jans/blob/main/docs/CONTRIBUTING.md#prs)
- [ ] Read [license information](https://github.com/JanssenProject/jans/blob/main/LICENSE)

-------------------

#### Target issue
  
  
  
  
closes #10563

#### Implementation Details
  

-------------------
### Test and Document the changes
- [ ] Static code analysis has been run locally and issues have been fixed
- [ ] Relevant unit and integration tests have been added/updated
- [ ] Relevant documentation has been updated if any (i.e. user guides, installation and configuration guides, technical design docs etc)

Please check the below before submitting your PR. The PR will not be merged if there are no commits that start with `docs:` to indicate documentation changes or if the below checklist is not selected.
- [x] **I confirm that there is no impact on the docs due to the code changes in this PR.**

PR #11161, opened 2025-04-03
feat(jans-cedarling): implement sending logs to the lock server

### LINKED ISSUE #10796: feat(jans-cedarling): implement lock server startup sequence
-------------------

### Description

- Development target issue
  
target issue: #10795 

- Implementation Details

This issue should implement the lock server startup sequence as described in this diagram

Image

-------------------

### Prepare

------------------

### Identified code changes

- [ ] Implement a `LockServer` struct that will hold the lock server configs and implementation.
- [ ] Step through the startup sequence to initialize the `LockServer` struct

-------------------

- [ ] Write unit test to cover added/changed code
- [ ] Update integration tests to cover added/changed code

-------------------

### Document the changes

- [ ] task for updating user guides if needed
- [ ] task for updating installation and configuration guides if needed 
- [ ] task for updating developer documentation if needed
- [ ] task for updating technical documentation if needed

rmarinn:
### Prepare

-------------------

### Description

This PR implements the following:
- DCR with the Lock server's IDP to obtain an access token. 
- Sending logs to the lock server's `/audit/log/bulk` endpoint. closes #

#### Target issue

Target issue: #10795
  
closes #10796

#### Implementation Details

For Cedarling to send logs to the Lock server, it must first dynamically register with the lock server's IDP.

The dynamic registration is done as follows:
1. Derive the `.well-known/openid-configuration` endpoint from the response of the `.well-known/lock-server-configuration` endpoint.
2. Perform DCR with an SSA JWT with the IDP to obtain an `access_token`.

After the DCR, a (green) worker thread will be spawned to periodically `POST` the logs to the `/audit/log/bulk` endpoint in batches. All `POST` requests will have a `Authorization: Bearer some.access.token` header.

-------------------


rmarinn:
This should be functionally sound but I can't figure out how we're supposed to have e2e tests for this.

Setting up the Lock Server is not exactly trivial and startup times for my fairly fast PC takes around 10 mins so I'm not sure if having a test runner that sets that up is feasible.

jans-cedarling/cedarling/src/log/lock_logger.rs
```diff
@@ -0,0 +1,332 @@
...
+        let mut auth_header =
+            HeaderValue::from_str(&format!("Bearer {}", client_creds.access_token))
+                .expect("build header");
+        auth_header.set_sensitive(true);
+        headers.insert("Authorization", auth_header);
+        let http_client = Arc::new(Client::builder().default_headers(headers).build()?);
+
+        let log_worker_tx = match (bootstrap_conf.log_interval, lock_config.audit_endpoints.log) {
+            // Case where Cedarling's config enables sending logs but the lock server
+            // does not have a log endpoint
```
olehbozhok: Maybe we need return error, to notify user?

jans-cedarling/cedarling/src/log/lock_logger.rs
```diff
@@ -0,0 +1,332 @@
...
+            (Some(_), None) => None,
+            (Some(log_interval), Some(log_endpoint)) => {
+                let (log_tx, log_rx) = mpsc::channel::<SerializedLogEntry>(100);
```
olehbozhok: What happened if the buffer full?

olehbozhok: Can you share your thoughts why you hardcoded buffer capacity to 100?

olehbozhok: OK, I see it will print to error to `stderr` but it may not work in WASM

jans-cedarling/cedarling/src/log/lock_logger.rs
```diff
@@ -0,0 +1,332 @@
...
+
+impl LogWriter for LockLogger {
+    fn log_any<T: Loggable>(&self, entry: T) {
+        // TODO: we could probably just generalize the fallback logger then log there.
+        let entry = serde_json::to_string(&entry)
+            .expect("entry should serialize successfully")
+            .into_boxed_str();
+        if let Some(log_tx) = self.log_worker_tx.as_ref() {
+            if let Err(err) = log_tx.try_send(entry) {
+                eprintln!(
```
olehbozhok: It may not print to WASM browser console..

jans-cedarling/cedarling/src/log/memory_logger/mod.rs
```diff
@@ -95,14 +97,25 @@ mod fallback {
     /// call. But this is a fallback logger, so it is not intended to be used
     /// often, and in this case correctness and non-fallibility are far more
     /// important than performance.
+    ///
+    /// # Panics
+    ///
+    /// Panics when:
+    /// - A runtime to initialize a new LogStrategy could not be built.
+    /// - A fallback logger could not be initialized.
+    // Can't we just log to StdOut directly instead of trying to initialize a StdOutLogger
```
olehbozhok: `Can't we just log to StdOut directly` - No, in WASM stdout and stderr does not work.

jans-cedarling/cedarling/src/log/lock_logger/log_worker.rs
```diff
@@ -0,0 +1,109 @@
...
+use tokio::sync::mpsc;
+use tokio::time::sleep;
+use url::Url;
+
+pub type SerializedLogEntry = Box<str>;
+
+/// Responsible for sending logs to the lock server
+pub struct LogWorker {
+    // it would be nice to store a struct here but we can't really store
+    // `VecDeque<dyn Loggable>` so we just serialize the logs before storin them in the
```
olehbozhok: `storin` - looks like a typo.

jans-cedarling/cedarling/src/log/lock_logger/log_worker.rs
```diff
@@ -0,0 +1,109 @@
...
+        let mut failed_serializations = 0;
+        let logs = self
+            .log_buffer
+            .iter()
+            .map(|entry| {
+                serde_json::from_str::<Value>(entry)
+                    .map_err(|_| failed_serializations += 1)
+                    .ok()
+            })
+            .collect::<Value>();
```
olehbozhok: Make sure that it returns json array of logs.

jans-cedarling/cedarling/src/log/lock_logger/log_worker.rs
```diff
@@ -0,0 +1,109 @@
...
+                .await
+                .and_then(|resp| resp.error_for_status());
+
+            match resp {
+                Ok(_) => {
+                    break;
+                },
+                Err(err) => {
+                    eprintln!("failed to POST logs: {err}");
+                    sleep(WORKER_HTTP_RETRY_DUR).await;
```
olehbozhok: I think in next iteration we can have more logs. So maybe should add them to send to the lock server?

jans-cedarling/cedarling/src/log/lock_logger/log_worker.rs
```diff
@@ -0,0 +1,109 @@
...
+        // log errors to stdout since there's nowhere else to log
+        if failed_serializations > 1 {
+            eprintln!("failed to serialize {failed_serializations} log entries");
+        }
+
+        loop {
+            let resp = self
+                .http_client
+                .post(self.log_endpoint.as_ref())
+                .body(logs.to_string())
```
olehbozhok: I tried to use streaming here (to avoid allocating logs to string) and enable `stream` in the `reqwest` lib, but failed. The compiler return an error related to `futures-util` lib. Probably I'm missing something.

... in golang it is very easy to implement.

jans-cedarling/cedarling/src/log/lock_logger.rs
rmarinn: > Can you share your thoughts why you hardcoded buffer capacity to 100?

Having this unbounded can silently cause the process to blow up under load or during spikes and 100 felt like plenty.

There could be a better limit and it will probably be dependent on the hardware that will run Cedarling -- right now, I'm not sure how we can figure that out.

jans-cedarling/cedarling/src/log/lock_logger/log_worker.rs
rmarinn: `.collect::<Value>()` turns it into an array of objects.

example:
```rs
fn main() {
    let log_buf = ["{ \"log_id\": 1 }", "{ \"log_id\": 2 }"];
    let serialized = log_buf
        .iter()
        .map(|s| serde_json::from_str::<Value>(s).unwrap())
        .collect::<Value>();
    println!("{:#?}", serialized);
}
```

output:
```sh
Array [
    Object {
        "log_id": Number(1),
    },
    Object {
        "log_id": Number(2),
    },
]
```

jans-cedarling/cedarling/src/log/lock_logger/log_worker.rs
rmarinn: > I tried to use streaming here (to avoid allocating logs to string)

do you mean this: [wrap_stream](https://docs.rs/reqwest/0.12.15/reqwest/struct.Body.html#method.wrap_stream)? I haven't looked into yet.

The logs have to be stored and batched first which i ran into problems with since i can't collect `dyn Loggable` with the current implementation.

```rs
pub struct LogWorker {
    // the trait `log::interface::Loggable` cannot be made into an object consider moving `serialize` to another trait...
    log_buffer: VecDeque<Box<dyn Loggable>>,
}
```

My current solution is to serialize them first before storing them:
```rs
impl LogWriter for LockLogger {
    fn log_any<T: Loggable>(&self, entry: T) {
        // TODO: we could probably just generalize the fallback logger then log there.
        let entry = serde_json::to_string(&entry)
            .expect("entry should serialize successfully")
            .into_boxed_str();
        if let Some(log_tx) = self.log_worker_tx.as_ref() {
            if let Err(err) = log_tx.try_send(entry) {
                eprintln!(
                    "failed to send log entry to LogWorker, the thread may have unexpectedly closed or is full: {err}"
                );
            }
        }
    }
}
```

jans-cedarling/cedarling/src/log/lock_logger/log_worker.rs
olehbozhok: Cool!

jans-cedarling/cedarling/src/log/lock_logger/log_worker.rs
olehbozhok: I just wanted to avoid using `logs.to_string()` in `.body()` to skip extra string allocation and use small batches when encoding Rust JSON `Value` to bytes. But it seems too complex, so let current solution be.

jans-cedarling/cedarling/src/log/lock_logger/log_worker.rs
rmarinn: what do you mean by "more logs"?

if you mean the healthchecks and telemetry, they're not included here yet.

olehbozhok: You added logs to `logs` variable.
if error happen, on next iteration we may have more logs in `log_buffer`

rmarinn: i see, you want to increase the batch size? 

that would mean that we have to re-serialize the whole batch since the `/bulk` endpoint requires a JSON body like:

```json
[
    { ... },
    { ... }
]
```

We can always send the next logs in the next batch.

Is there a reason why you're suggesting that we grow the batch size when `POST`-ing the log fails?

jans-cedarling/cedarling/src/log/lock_logger.rs
rmarinn: This runs as a background task so the only way to "return an error" is to print it out to stdout or write the output to a file.

How do you suggest we return this error?

jans-cedarling/cedarling/src/log/lock_logger/log_worker.rs
rmarinn: resolved in [156fbab](https://github.com/JanssenProject/jans/pull/11161/commits/156fbabcb03107a5eb389dad35efd619c8215a73)

jans-cedarling/cedarling/src/log/memory_logger/mod.rs
rmarinn: removed this comment in [156fbab](https://github.com/JanssenProject/jans/pull/11161/commits/156fbabcb03107a5eb389dad35efd619c8215a73)

jans-cedarling/cedarling/src/log/lock_logger.rs
rmarinn: I initialized a StdOut logger that the lock logger will as a fallback.

See [156fbab](https://github.com/JanssenProject/jans/pull/11161/commits/156fbabcb03107a5eb389dad35efd619c8215a73)

jans-cedarling/cedarling/src/log/lock_logger.rs
rmarinn: Used a StdOut logger to print out the errors for now: [156fbab](https://github.com/JanssenProject/jans/pull/11161/commits/156fbabcb03107a5eb389dad35efd619c8215a73)

jans-cedarling/cedarling/src/log/lock_logger/log_worker.rs
olehbozhok: yes, I was thinking that maybe we need serialize logs
```
        let logs = self
            .log_buffer
            .iter()
            .map(|entry| {
                serde_json::from_str::<Value>(entry)
                    .map_err(|_| failed_serializations += 1)
                    .ok()
            })
            .collect::<Value>();
```
in each loop iteration, for case if log service was unavailable for a while and after send all available logs. Because log_buffer can be full.
But I'm not sure if we really need it, maybe for now, not. It is up to you.

olehbozhok (approved):

jans-cedarling/cedarling/src/log/lock_logger/log_worker.rs
rmarinn: i think we shouldn't be growing it indefinitely either. if the log service is down for awhile and we now have a stockpile of logs, should we send all of it at once?

log_buffer won't really get filled up unless the machine runs out of memory since it's just a VecDeque but we can't be `POST`-ing enormous payloads to lock servers

olehbozhok (dismissed):

rmarinn:
The version in this branch PR should successfully be `POST`-ing logs to the lock server now, so it would be great if you can test it @SafinWasi as you already have an auth server set up.

We still have an issue with the scopes though. The [startup sequence in the wiki](https://github.com/JanssenProject/jans/wiki/Cedarling-Nativity-Plan#cedarling-startup-sequence) says that the client needs the `cedarling` scope but in practice, the `https://jans.io/oauth/lock/log.write` is required. Do we just need to update the docs or this is a configuration issue in the auth/lock server?

olehbozhok (dismissed):

SafinWasi:
I don't think we _can_ have E2E or integration tests for this. The best thing I can think of is mocking an HTTP server with the token and lock endpoints, and having cedarling call those.

rmarinn:
@SafinWasi Here's the setup i have for testing

## Required

- AS/Lock Server for version 1.4.0 (the endpoints for the lock server in v1.5.0 isn't working)
- Set the `https://jans.io/oauth/lock/lock/log.write` endpoint to a default scope. I did this using `jans tui`. I found that if you don't do this, you won't get the scope when the client performs DCR.

## Running a rust example:

This example will make logs and send it to the lock server at `demoexample.jans.io`. If you set it somewhere else, just change the `config_uri` to your `.well-known/lock-server-configuration` uri.

```sh
cargo run --example lock_integration
```

##  Bootstrap Config

If you want to set it up yourself you need the following bootstrap properties set:

```sh
CEDARLING_LOG_TYPE=std_out # this can also be `memory`, just don't turn it off
CEDARLING_LOCK=enabled
CEDARLING_LOCK_CONFIGURATION_URI=`https://demoexample.jans.io/.well-known/lock-master-configuration`
CEDARLING_LOG_INTERVAL=3 # this is in seconds
```

health and telemetry isn't implemented yet in the PR, only sending logs.

jans-cedarling/cedarling/src/log/log_strategy.rs
```diff
@@ -43,36 +44,43 @@ impl LogStrategy {
...
             LogTypeConfig::StdOut => LogStrategyLogger::StdOut(StdOutLogger::new(config.log_level)),
-            LogTypeConfig::Lock(lock_bootstrap_config) => LogStrategyLogger::Lock(
-                LockLogger::new(pdp_id, app_name.clone(), lock_bootstrap_config).await?,
-            ),
         };
         Ok(Self {
             logger,
             pdp_id,
             app_name,
+            lock_logger: RwLock::new(None),
```
olehbozhok: It looks weird. Why we need initialize `RwLock` with None value?

jans-cedarling/cedarling/src/log/log_strategy.rs
```diff
@@ -13,11 +13,13 @@ use super::nop_logger::NopLogger;
 use super::stdout_logger::StdOutLogger;
 use crate::app_types::{ApplicationName, PdpID};
 use crate::bootstrap_config::log_config::{LogConfig, LogTypeConfig};
+use serde::Serialize;
 
 pub(crate) struct LogStrategy {
     logger: LogStrategyLogger,
     pdp_id: PdpID,
     app_name: Option<ApplicationName>,
+    lock_logger: RwLock<Option<LockLogger>>,
```
olehbozhok: how about change it to `Option<RwLock<LockLogger>>` ?

olehbozhok (changes requested):

jans-cedarling/cedarling/src/log/log_strategy.rs
rmarinn: I need to pass in the Logger into `LockService`:

```rs
impl LockService {
    pub async fn new(
        pdp_id: PdpID,
        bootstrap_conf: &LockServiceConfig,
        logger: Option<Logger>,
    ) -> Result<Self, InitLockServiceError> {
        // ...
    }
}
```

Which means the `Logger` needs to be initialized first.

However, LockLogger needs to use `LockService` too:

```rs
// Implementation of LogWriter
impl LogWriter for LogStrategy {
    fn log_any<T: Loggable>(&self, entry: T) {
        let entry =
            LogEntryWithClientInfo::from_loggable(entry, self.pdp_id, self.app_name.clone());
        // Send logs to the lock service here
        if let Some(lock_service) = self
            .lock_service
            .read()
            .expect("obtain lock_logger read lock")
            .as_ref()
        {
            lock_service.log_any(entry.clone());
        }
        match &self.logger {
            LogStrategyLogger::Off(log) => log.log_any(entry),
            LogStrategyLogger::MemoryLogger(memory_logger) => memory_logger.log_any(entry),
            LogStrategyLogger::StdOut(std_out_logger) => std_out_logger.log_any(entry),
        }
    }
}
```

This means we need to assign the LockService to LockLogger after it's initialization.

But since:
```rs
/// Type alias for logger that is used in application
pub(crate) type Logger = Arc<LogStrategy>;
```

We cant have:
```rs
    pub fn set_lock_service(&mut self, lock_logger: LockService) {
         // ...
    }
```

Which is why we introduce a `RwLock` for interior mutability.

jans-cedarling/cedarling/src/log/log_strategy.rs
rmarinn: we can't do that because of this definition:

```rs
/// Type alias for logger that is used in application
pub(crate) type Logger = Arc<LogStrategy>;
```

We can't borrow the `LogStrategy` as mutable after it's initialized. I explained this more in my other [comment](https://github.com/JanssenProject/jans/pull/11161#discussion_r2050455625).

SafinWasi:
Regarding setting `https://jans.io/oauth/lock/lock/log.write` as a default scope, we can instead use a DCR interception script to add the scope if the provided SSA contains the `cedarling` claim. We use this for license SSAs.
https://docs.jans.io/v1.5.0/script-catalog/client_registration/client-registration/

jans-cedarling/cedarling/src/log/log_strategy.rs
olehbozhok: Looks like you made reference cycle here
image

We have logger, and you pass it to `lock service` and after it pass `lock service` to logger..

So looks like reference count newer can be zero, and it means it can't be dropped. We need to fix it. Maybe we can downgrade logger to `weak` when pass it to `lock service`?
`Arc::<LogStrategy>::downgrade(&logger)`

Or maybe I miss something?

olehbozhok: Added fix with `weak` reference in 7e881f20a44011032f265bca81b8e48635589772 .

rmarinn: Using `Weak` to break the reference cycle looks good to me.

rmarinn:
> Regarding setting `https://jans.io/oauth/lock/lock/log.write` as a default scope, we can instead use a DCR interception script to add the scope if the provided SSA contains the `cedarling` claim. We use this for license SSAs. https://docs.jans.io/v1.5.0/script-catalog/client_registration/client-registration/

I think this should be included in the installation of the lock server by default, no?

SafinWasi:
> > Regarding setting `https://jans.io/oauth/lock/lock/log.write` as a default scope, we can instead use a DCR interception script to add the scope if the provided SSA contains the `cedarling` claim. We use this for license SSAs. https://docs.jans.io/v1.5.0/script-catalog/client_registration/client-registration/
> 
> I think this should be included in the installation of the lock server by default, no?

Interception scripts are not default AS behavior, they provide additional business logic. In this case having the interception script on for users who don't use cedarling/lock would be problematic.

rmarinn:
> > > Regarding setting `https://jans.io/oauth/lock/lock/log.write` as a default scope, we can instead use a DCR interception script to add the scope if the provided SSA contains the `cedarling` claim. We use this for license SSAs. https://docs.jans.io/v1.5.0/script-catalog/client_registration/client-registration/
> > 
> > 
> > I think this should be included in the installation of the lock server by default, no?
> 
> Interception scripts are not default AS behavior, they provide additional business logic. In this case having the interception script on for users who don't use cedarling/lock would be problematic.

If we can't add the interception script by default, don't we have no choice but to set the default claim that Cedarling will use to be `https://jans.io/oauth/lock/lock/log.write`? 

Other options i think we can do is:
a.  just request for both the scopes at DCR to have both `"cedarling https://jans.io/oauth/lock/lock/log.write"`.
b. have a bootstrap property to edit the scope.

SafinWasi (dismissed):

jans-cedarling/cedarling/src/log/log_strategy.rs
```diff
@@ -29,42 +33,65 @@ pub(crate) enum LogStrategyLogger {
...
+        app_name: Option<ApplicationName>,
+    ) -> Result<Self, InitLockServiceError> {
+        let logger = match &config.log_type {
             LogTypeConfig::Off => LogStrategyLogger::Off(NopLogger),
             LogTypeConfig::Memory(memory_config) => LogStrategyLogger::MemoryLogger(
-                MemoryLogger::new(memory_config, config.log_level, pdp_id, app_name.clone()),
+                MemoryLogger::new(*memory_config, config.log_level, pdp_id, app_name.clone()),
             ),
             LogTypeConfig::StdOut => LogStrategyLogger::StdOut(StdOutLogger::new(config.log_level)),
-            LogTypeConfig::Lock => todo!(),
```
0xtinkle, line 39: why removed? does it need to be implemented?

jans-cedarling/cedarling/src/bootstrap_config/decode.rs
```diff
@@ -55,7 +57,6 @@ impl BootstrapConfig {
                 max_items: raw.log_max_items,
             }),
             LoggerType::StdOut => LogTypeConfig::StdOut,
-            LoggerType::Lock => LogTypeConfig::Lock,
```
0xtinkle, line 58: why removed?

jans-cedarling/cedarling/src/lock/mod.rs
```diff
@@ -4,6 +4,462 @@
 // Copyright (c) 2024, Gluu, Inc.
 
 //! # Lock Engine
-//! Lock Engine - it is enterprise feature.
```
0xtinkle, line 7: can you add proper description which explains the architecture so that anyone can understand it easily?
e.g.
//! Lock Engine provides integration with the Lock server for log aggregation and sending.
//! 
//! ## Architecture
//! - Uses background worker for log processing
//! - Collects logs from other loggers
//! - Handles DCR through register_client
//! - Sends logs to Lock server

0xtinkle:
Resolve conflicts!

jans-cedarling/cedarling/src/lock/mod.rs
```diff
@@ -4,6 +4,462 @@
 // Copyright (c) 2024, Gluu, Inc.
```
0xtinkle, line 5: Can you add specific error types?
`pub enum LockError`

jans-cedarling/cedarling/src/lock/register_client.rs
```diff
@@ -0,0 +1,128 @@
...
+    accept_invalid_certs: bool,
+) -> Result<ClientCredentials, ClientRegistrationError> {
+    let client = init_http_client(None, accept_invalid_certs)?;
+
+    #[cfg(not(test))]
+    let mut sender = Sender::new(Backoff::default_exponential());
+
+    // We implement a faster retry for the tests
+    #[cfg(test)]
+    let mut sender = {
```
0xtinkle, line 33: does it need timeout error handling for DCR?

0xtinkle:
Is it better to handle relevant configs in one place - LockConfig?
e.g. batch_size, max_attempts, retry_dur, timeout...

jans-cedarling/cedarling/src/lock/mod.rs
rmarinn: improved docs here [98a3d34](https://github.com/JanssenProject/jans/pull/11161/commits/98a3d34635ad15ad2b78d225a1793246d34c9921)

jans-cedarling/cedarling/src/lock/mod.rs
rmarinn: There's an `InitLockServiceError` enum. Currently, this is the only error type that's return when called by other code within Cedarling (so we only error on initialization).

On runtime, errors from things like sending http requests are serialized immediately by the `LockLogWorker` to a `LockLogEntry` then sent to the logger.

jans-cedarling/cedarling/src/lock/register_client.rs
rmarinn: The timeout and retry behavior is managed by the `Sender::send` function from `jans-cedarling/http_utils`.

In the file you linked, the `Sender` struct is initialized with an exponential backoff strategy using the following configuration:
```rs
    let mut sender = {
        use std::time::Duration;
        Sender::new(Backoff::new_exponential(Duration::from_millis(10), Some(3)))
    };
```
This setup starts with a 10ms delay and retries up to 3 times.

Later in the code, you can see the `Sender::send` function in use when retrieving the OpenID configuration:
```rs
    // Get openid config
    let oidc: OpenidConfig = sender
        .send(|| client.get(oidc_endpoint.as_ref()))
        .await
        .map_err(ClientRegistrationError::GetOpenidConfig)?;
```
This ensures that transient failures during the HTTP request are retried with backoff before logging the error.

rmarinn:
> Is it better to handle relevant configs in one place - LockConfig? e.g. batch_size, max_attempts, retry_dur, timeout...

The configuration values used by `LockService`, including batch_size, max_attempts, retry_dur, and timeout, are ultimately collected into the `LockConfig` struct, located in `cedarling/src/lock/lock_config.rs`.

That said, there are currently multiple representations of related config values spread across:
- `cedarling/src/bootstrap_config/log_config.rs`: via the `LockServiceConfig` struct
- `cedarling/src/bootstrap_config/raw_config.rs`: via the `BootstrapConfigRaw` struct

The current flow for loading configuration is:
- Deserialize the raw input into `BootstrapConfigRaw`.
- Convert `BootstrapConfigRaw` into `BootstrapConfig`.
- Extract and validate the relevant settings into the `LockConfig` struct.

Personally, I think it makes the most sense to keep all configs related to the Lock module within the lock module itself. However, the last time I attempted that, I was asked to move them under the `/bootstrap_config` module instead.

Refactoring the entire config loading flow would require a significant time investment. At the moment, I'm focused on other tasks, so I decided to stick with the current approach for now.

rmarinn:
> Resolve conflicts!

resolved in [eb149f8](https://github.com/JanssenProject/jans/pull/11161/commits/eb149f837e8a8ccfbb7069d62d166710ec85c62e)

rmarinn:
#10564 might have broke the ci

rmarinn:
> #10564 might have broke the ci

opened #11427 to resolve this

0xtinkle (dismissed):

rmarinn:
ci is fixed so could you guys re-approve, @0xtinkle? thanks

0xtinkle (approved):

PR #12173, opened 2025-09-17
feat(jans-cedarling): Cedarling tokens caching

### LINKED ISSUE #12167: Cedarling tokens caching
-------------------

### Description

Add caching tokens with SparKV to avoid additional overhead on decoding the same token on each request.

------------------

### Identified code changes

- [ ] Add SparKV token cache storage
- [ ] Maybe use [Cov](https://doc.rust-lang.org/std/borrow/enum.Cow.html) or [Arc](https://doc.rust-lang.org/std/sync/struct.Arc.html) to avoid additional clone overhead.
- [ ] Add CI task to check if benchmarks take more than 1m

olehbozhok:
### Prepare

-------------------

### Description
Added optimizations:
- Cache tokens to avoid deserialize and validate same token on each request
- Wrapped token strucute with `Arc` to avoid cloning the whole structure.
- Used `ahash` algorithm to hash token strings as key, (improved perfomance to seach over Btree in SparKV)

#### Target issue
  
  [link](https://github.com/JanssenProject/jans/issues/12167)
  
  
closes #12167

-------------------


mo-auto:
### :tada: **Snyk checks have passed. No issues have been found so far.**

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/janssenproject/pr-checks/de6c9706-b1db-4d65-a3d6-16ff2a148211?refs=4b7cfce64df5ca53bd3558e0c89f9c8e63a70402&source=prChecksComment)

jans-cedarling/cedarling/examples/profiling.rs
```diff
@@ -23,8 +23,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         .build()
         .unwrap();
 
-    // calls Cedarling::authorize
-    call_authorize(&cedarling, &access_token, &id_token, &userinfo_token).await;
+    for _ in 0..1000 {
```
0xtinkle, line 26: Not sure, this is intended because this code is used as an example in doc.

jans-cedarling/cedarling/src/jwt/mod.rs
```diff
@@ -166,53 +174,101 @@ impl JwtService {
...
-                Err(err) => {
-                    if matches!(err, ValidateJwtError::MissingValidator(_)) {
-                        self.logger
-                            .log_any(JwtLogEntry::new(err.to_string(), Some(LogLevel::WARN)));
-                        continue;
-                    } else {
-                        return Err(JwtProcessingError::ValidateJwt(token_name.clone(), err));
-                    }
-                },
+            let token = if let Some(validated_token) = self.find_token_in_cache(jwt) {
```
0xtinkle, line 198: Should we check the exp of the token here?

jans-cedarling/cedarling/src/jwt/mod.rs
```diff
@@ -106,6 +113,7 @@ pub struct JwtService {
     key_service: Arc<KeyService>,
     issuer_configs: HashMap<IssClaim, IssuerConfig>,
     logger: Option<Logger>,
+    token_cache: Arc<Mutex<SparKV<Arc<Token>>>>,
```
haileyesus2433: Using Mutex creates a significant concurrency bottleneck. All cache operations (both reads and writes) are serialized, meaning one slow operation blocks all concurrent requests. In a high-traffic scenario, cache reads will queue behind cache writes, eliminating the performance benefits of caching. i think it will be better if we use
`token_cache: Arc<RwLock<SparKV<Arc<Token>>>>`

jans-cedarling/cedarling/src/jwt/mod.rs
```diff
@@ -332,6 +387,16 @@ fn fix_aud_claim_value_to_array(claims: TokenClaims) -> TokenClaims {
     claims
 }
 
+/// Hash a string using `ahash` and return the hash value as a string
+/// This is used to create a key for caching tokens
+/// The hash value is used instead of the original string to have shorter keys for SparKV which utilizes BTree.
+fn hash_str(s: &str) -> String {
```
haileyesus2433, line 419: [Hash Collision Security Vulnerability](https://www.startupdefense.io/cyberattacks/hash-collision-attack)
There is No collision detection Different JWT strings with the same hash will return wrong tokens also we might face  DoS vulnerability since `AHasher::default()` uses compile-time constant, enabling collision attacks [resource](https://docs.rs/ahash/latest/ahash/struct.AHasher.html)
what do u think about this @olehbozhok 
```rust
fn hash_str(s: &str) -> String {
    use std::sync::LazyLock;
    static HASHER_KEYS: LazyLock<(u64, u64)> = LazyLock::new(|| {
        (rand::random(), rand::random())
    });
    
    let mut hasher = ahash::AHasher::new_with_keys(HASHER_KEYS.0, HASHER_KEYS.1);
    s.hash(&mut hasher);
    hasher.finish().to_string()
}
```

jans-cedarling/cedarling/src/jwt/mod.rs
```diff
@@ -166,53 +174,101 @@ impl JwtService {
...
-                token_name.to_string(),
-                Token::new(token_name, claims, validated_jwt.trusted_iss),
-            );
+            validated_tokens.insert(token_name.to_string(), token);
         }
 
         Ok(validated_tokens)
     }
 
+    fn find_token_in_cache(&self, jwt: &str) -> Option<Arc<Token>> {
```
haileyesus2433, line 240: Memory Leak in Read-Heavy Workloads
Expired entries are only cleared during writes (`clear_expired_if_auto()` called in `set_with_ttl` in `sparkv`). In read-only workloads, memory grows unbounded. Adding cleanup to read path should fix it

jans-cedarling/cedarling/src/jwt/mod.rs
```diff
@@ -166,53 +174,101 @@ impl JwtService {
...
+                let now = Utc::now();
+                let duration = exp - now.timestamp();
+                if duration > 0 {
+                    Some(duration as u64)
+                } else {
+                    None
+                }
+            });
+
+        if let Some(duration) = cache_duration_opt {
```
haileyesus2433, line 282: Caching already-expired tokens wastes memory and creates confusing behavior please add early return so that we dont cache expired tokens

jans-cedarling/cedarling/src/jwt/mod.rs
```diff
@@ -166,53 +174,101 @@ impl JwtService {
...
+
+    fn save_token_in_cache(&self, jwt: &str, token: Arc<Token>) {
+        let key = hash_str(jwt);
+
+        let cache_duration_opt = token
+            .claims
+            .get_claim("exp")
+            .and_then(|exp| exp.value().as_i64())
+            .and_then(|exp| {
+                let now = Utc::now();
```
haileyesus2433: Repeated Timestamp Calls `Utc::now()` is called once per token in the loop, creating unnecessary system calls.
what is the effect if we Cache timestamp at start of `validate_tokens()`

jans-cedarling/cedarling/examples/profiling.rs
olehbozhok: Yes, it was intended, because without it the report is empty

jans-cedarling/cedarling/src/jwt/mod.rs
olehbozhok: Yes, you are right I missed it.

jans-cedarling/cedarling/src/jwt/mod.rs
olehbozhok: `In read-only workloads, memory grows unbounded` - why?

jans-cedarling/cedarling/src/jwt/mod.rs
olehbozhok: Valid point, thanks!

jans-cedarling/cedarling/src/jwt/mod.rs
olehbozhok: Added early clearing expired tokens from cache.

jans-cedarling/cedarling/src/jwt/mod.rs
olehbozhok: In SparKV default expiration time - 5 minutes. So it shouldn't be a problem. And added early cleaning tokens from cache (before tokens iteration).

jans-cedarling/cedarling/src/jwt/mod.rs
olehbozhok: Added calling `clear_expired()` before token iteration.

jans-cedarling/cedarling/src/jwt/mod.rs
olehbozhok: I enabled the ahash feature `compile-time-rng` because we need stable keys generation. I like your idea and just found the `runtime-rng` feature. I digged deeper and based on source code, it should work the same and it is enabled by default.

olehbozhok: updated (used default `ahash` features)

0xtinkle (dismissed):

jans-cedarling/cedarling/src/jwt/mod.rs
haileyesus2433: when i saw `clear_expired_if_auto()` only called during writes "Read-only workloads = no cleanup = unbounded memory growth" but i missed the Default TTL of 5 minutes sorry for my mistake

haileyesus2433 (dismissed):

moabu (approved):

PR #13538, opened 2026-03-19
feat(jans-cedarling): Remove authorization method based on user principals

haileyesus2433:
### Prepare

- [x] Read [PR guidelines](https://github.com/JanssenProject/jans/blob/main/docs/CONTRIBUTING.md#prs)
- [x] Read [license information](https://github.com/JanssenProject/jans/blob/main/LICENSE)

-------------------

#### Target issue
  
  
  
  
closes #13381

#### Implementation Details
  

-------------------
### Test and Document the changes
- [x] Static code analysis has been run locally and issues have been fixed
- [x] Relevant unit and integration tests have been added/updated
- [x] Relevant documentation has been updated if any (i.e. user guides, installation and configuration guides, technical design docs etc)

Please check the below before submitting your PR. The PR will not be merged if there are no commits that start with `docs:` to indicate documentation changes or if the below checklist is not selected.
- [x] **I confirm that there is no impact on the docs due to the code changes in this PR.**

PR #13542, opened 2026-03-20
feat(jans-cedarling): add  C bindings support

### LINKED ISSUE #11754: feat(jans-cedarling): Add C Bindings Support
The cedarling currently can’t be used directly from C. Developers who use C can’t easily access the functionalities provided.

Provide C bindings by exposing a C-compatible API.

N/A

**Additional context**
N/A

haileyesus2433:
### Prepare

-------------------

#### Target issue
  
  
  
  
closes #11754

#### Implementation Details
  

-------------------

## Summary by CodeRabbit

* **New Features**
  * C bindings packaged as shared/static libraries exposing init, instance lifecycle, authorization (unsigned & multi‑issuer), context data API, logging, trusted‑issuer inspection, error/ memory helpers, and example programs.

* **Documentation**
  * New C getting‑started tutorial, integrated C snippets in the main guide, and a dedicated C bindings README.

* **Tests**
  * End‑to‑end C test suite, example Makefiles, and runnable C examples.

* **Chores**
  * CI updated to add C test job and minor CI formatting fix.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/6103c486-8eed-4220-82bf-0e357cd26fe1?refs=c7910e66ef5deffcb39c89050c982891f9f6c5cb&source=prChecksComment) |

haileyesus2433:
the failing C tests will be resolved once https://github.com/JanssenProject/jans/pull/13538 is merged its currently blocked by that

tareknaser:
@haileyesus2433  
I have some review comments for this PR. I’m finalizing a few more and will submit my review in the next couple of days

jans-cedarling/bindings/cedarling_c/tests/test_cedarling.c
```diff
@@ -0,0 +1,499 @@
...
+        printf(" Authorization result: %.200s...\n", result_str);
+        
+        // Verify the response structure contains expected fields
+        // Note: The overall decision depends on principal_bool_operator config (not exposed to C)
+        // Default operator requires Jans::Workload AND Jans::User, so decision will be false
+        // when using TestPrincipal1. We verify the individual principal got "allow".
+        TEST_ASSERT(strstr(result_str, "\"decision\":\"allow\"") != NULL,
+                   "Individual principal decision should be allow");
+        TEST_ASSERT(strstr(result_str, "principals") != NULL,
+                   "Response should contain principals field");
```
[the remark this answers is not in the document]
haileyesus2433: Addressed in commits https://github.com/JanssenProject/jans/commit/0f1a2e42437bdf2647822c9987c4abd95e443d9e to https://github.com/JanssenProject/jans/commit/28d3367625c4a5397e6c364c4f284c8645a1b1c8

jans-cedarling/bindings/cedarling_c/examples/makefile
```diff
@@ -0,0 +1,51 @@
...
+INCLUDES = -I../include
+
+# Detect OS for library linking
+UNAME_S := $(shell uname -s)
+ifeq ($(UNAME_S),Linux)
+    LDFLAGS += -ldl -lpthread -lm
+endif
+ifeq ($(UNAME_S),Darwin)
+    LDFLAGS += -ldl -lpthread -lm
+endif
```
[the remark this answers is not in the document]
haileyesus2433: Addressed in commits https://github.com/JanssenProject/jans/commit/0f1a2e42437bdf2647822c9987c4abd95e443d9e to https://github.com/JanssenProject/jans/commit/28d3367625c4a5397e6c364c4f284c8645a1b1c8

.github/workflows/test-cedarling.yml
```diff
@@ -228,4 +228,37 @@ jobs:
...
+      run: |
+        cargo build --release -p cedarling_c
+        test -f target/include/cedarling_c.h || (echo "header missing" && exit 1)
+    - name: Build and run C tests
+      working-directory: jans-cedarling/bindings/cedarling_c
+      run: |
+        export LD_LIBRARY_PATH="../../target/release:${LD_LIBRARY_PATH}"
+        cd tests
+        make
+        ./test_cedarling
```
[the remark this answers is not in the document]
haileyesus2433: this is fine since the test is working and passing in the [CI](https://github.com/JanssenProject/jans/actions/runs/23653215151/job/68904755542)

docs/cedarling/tutorials/c.md
```diff
@@ -0,0 +1,536 @@
...
+
+    ```sh
+    cbindgen --crate cedarling_c --output target/include/cedarling_c.h
+    ```
+
+5. Copy the built artifacts to your project:
+
+    - **Linux**: `cp target/release/libcedarling_c.so .`
+    - **macOS**: `cp target/release/libcedarling_c.dylib .`
+    - **Windows**: `cp target/release/cedarling_c.dll .` and `cp target/release/cedarling_c.dll.lib cedarling_c.lib`
```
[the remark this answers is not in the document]
haileyesus2433: this is fine

jans-cedarling/bindings/cedarling_c/src/types.rs
```diff
@@ -0,0 +1,214 @@
...
+                        count: 0,
+                    };
+                },
+            };
+        }
+        let items_ptr = c_strings.as_mut_ptr();
+        std::mem::forget(c_strings); // Prevent Deallocation
+        Self {
+            items: items_ptr,
+            count: strings.len(),
```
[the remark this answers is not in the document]
haileyesus2433: this is fine

jans-cedarling/bindings/cedarling_c/src/lib.rs
```diff
@@ -0,0 +1,667 @@
...
+        return CedarlingErrorCode::InvalidArgument as c_int;
+    }
+
+    let logs = pop_logs(instance_id);
+
+    unsafe { *result = logs };
+    if !get_last_error().is_null() {
+        return CedarlingErrorCode::InstanceNotFound as c_int;
+    }
+    CedarlingErrorCode::Success as c_int
```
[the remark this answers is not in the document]
haileyesus2433: Addressed in commit https://github.com/JanssenProject/jans/commit/07d0944e8819fbb44fb5e8587d70b80371740c03

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
```diff
@@ -0,0 +1,889 @@
...
+        Err(e) => {
+            let error_msg = format!("Failed to create Cedarling instance: {}", e);
+            set_last_error(&error_msg);
+
+            return CedarlingInstanceResult::error(CedarlingErrorCode::Internal, &error_msg);
+        },
+    };
+
+    let instance_id = runtime.add_instance(instance);
+    CedarlingInstanceResult::success(instance_id)
```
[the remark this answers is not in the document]
haileyesus2433: Addressed in commit https://github.com/JanssenProject/jans/commit/07d0944e8819fbb44fb5e8587d70b80371740c03

jans-cedarling/bindings/cedarling_c/src/lib.rs
```diff
@@ -0,0 +1,720 @@
...
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn cedarling_new(
+    config_json: *const c_char,
+    result: *mut CedarlingInstanceResult,
+) -> c_int {
+    clear_last_error();
+    if config_json.is_null() || result.is_null() {
+        set_last_error("null config_json or result pointer");
+        return CedarlingErrorCode::InvalidArgument as c_int;
+    }
```
tareknaser: This combined null check returns early without populating `*result`. If only `config_json` is null but `result` is a valid pointer, the result struct stays uninitialized. A C caller doing `cedarling_free_instance_result(&result)` after this would try to free garbage pointers.

I wrote a test to verify this: basically if you do `memset(&result, 0, sizeof(result))` then `cedarling_new(NULL, &result)` returns error code 1 but `result.ERROR_CODE` stays 0 (never set)

The same pattern is there for `cedarling_authorize_unsigned`, `cedarling_authorize_multi_issuer`, `cedarling_get_log_by_id`, `cedarling_get_logs_by_tag`, `cedarling_get_logs_by_request_id`, `cedarling_get_logs_by_request_id_and_tag`, `cedarling_context_push`, `cedarling_context_get`, `cedarling_context_remove`.

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
```diff
@@ -0,0 +1,983 @@
...
+        Ok(instance) => instance,
+        Err(e) => {
+            let error_msg = format!("Failed to create Cedarling instance: {}", e);
+            set_last_error(&error_msg);
+
+            return CedarlingInstanceResult::error(CedarlingErrorCode::Internal, &error_msg);
+        },
+    };
+
+    let instance_id = runtime.add_instance(instance);
```
tareknaser: `add_instance` returns 0 when the mutex lock fails. `create_instance` checks for this and returns an error (See below). `create_instance_with_env` skips that check and returns `CedarlingInstanceResult::success(0)`. The caller thinks creation succeeded but every call after this one with ID 0 will fail with "instance not found."

```rust
let instance_id = runtime.add_instance(instance);
if instance_id == 0 {
    return CedarlingInstanceResult::error(
        CedarlingErrorCode::Internal,
        "Internal lock failure while adding instance",
    );
}
```

jans-cedarling/bindings/cedarling_c/src/lib.rs
```diff
@@ -0,0 +1,720 @@
...
+        clear_last_error();
+        set_last_error("null result pointer");
+        return CedarlingErrorCode::InvalidArgument as c_int;
+    }
+
+    let logs = pop_logs(instance_id);
+
+    unsafe { *result = logs };
+    if !get_last_error().is_null() {
+        return CedarlingErrorCode::InstanceNotFound as c_int;
```
tareknaser: `get_last_error()` can be not null for multiple reasons: instance not found, interior NUL byte in a log string, mutex lock failure. The caller always gets error code 2 (`INSTANCE_NOT_FOUND`) no matter what actually went wrong.

The internal functions already set a specific error. We need to propagate the actual error code instead of hardcoding it.

jans-cedarling/bindings/cedarling_c/src/lib.rs
```diff
@@ -0,0 +1,720 @@
...
+#[unsafe(no_mangle)]
+pub extern "C" fn cedarling_init() -> c_int {
+    // Initialize any global state if needed
+    0 // Success
+}
+/// # Safety
+/// Create a new Cedarling instance
+///
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn cedarling_new(
```
tareknaser, line 79: There are no direct panic sources like `.expect()` and `.lock().unwrap()` which is good (I had a pending comment about this but it was fixed yesterday) but panics from deeper in cedarling core library or from allocation failures can still propagate across the FFI boundary. This is UB.

I suggest adding an `ffi_guard!` macro (similar to what rustls-ffi does) that wraps every `extern "C"` function body in `catch_unwind`.

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
```diff
@@ -0,0 +1,983 @@
...
+        Err(code) => {
+            return CedarlingResult::error(code, "Internal lock failure");
+        },
+    };
+
+    let log = instance.get_log_by_id(log_id);
+
+    let log_string = log.iter().map(|l| l.to_string()).collect::<String>();
+
+    CedarlingResult::success(log_string)
```
tareknaser: When `get_log_by_id` returns `None`, this produces an empty string and returns success. This basically means that the caller can't tell "not found" from "empty."

> `None.iter()` yields zero elements, `collect::<String()` returns `""` and `CedarlingResult::success("")` returns error code 0 with a valid but empty DATA pointer.

We can update this to return an error with `InstanceNotFound` or `KeyNotFound` when the log is `None`.

jans-cedarling/bindings/cedarling_c/src/types.rs
```diff
@@ -0,0 +1,214 @@
...
+///
+/// Pointer is valid only until the next Cedarling library call, because most
+/// operations clear and may replace thread-local error storage at entry.
+/// C callers should copy this string immediately if it must outlive the next
+/// Cedarling API call.
+///
+/// Internally this returns `*mut c_char` for convenience with existing call sites,
+/// while the public C API exposes `*const c_char` for const-safety.
+pub fn get_last_error() -> *mut c_char {
+    LAST_ERROR.with(|last_error| {
```
tareknaser: This returns a pointer into thread-local storage. Any later cedarling call drops the CString which leaves the pointer dangling.

The doc comment at line 189 does mention this which is good but in practice, I assume C callers will do:

```c
const char* err = cedarling_get_last_error();
cedarling_new("{bad}", &result);  // drops old CString
printf("%s\n", err);              // use-after-free
```

Let's update this so we return a freshly allocated copy that the caller must free (This is the very first solution that comes to mind. maybe you have a better idea using lifetime constraints?)

jans-cedarling/bindings/cedarling_c/src/lib.rs
```diff
@@ -0,0 +1,720 @@
...
+/// This function only clears thread-local last-error state and performs no unsafe memory access.
+/// It is thread-safe and idempotent: it can be called multiple times with the same effect.
+/// There are no preconditions; call it at process shutdown or whenever you want to reset error state.
+///
+/// # Postconditions
+/// The calling thread's `cedarling_get_last_error()` value is reset to null until a new error is set.
+#[unsafe(no_mangle)]
+pub extern "C" fn cedarling_cleanup() {
+    // Perform any necessary cleanup
+    clear_last_error();
```
tareknaser: I think the function name may be a bit misleading.

A C developer calling `cedarling_cleanup()` at process exit would reasonably expect it to shut down all instances and release resources but it only clears the error state. I understand that the doc comment explain this but I think it's still confusing from an API design perspective.

jans-cedarling/bindings/cedarling_c/src/lib.rs
```diff
@@ -0,0 +1,720 @@
...
+use c_interface::*;
+use types::*;
+/// # Safety
+/// Intialize the Cedarling library
+/// This Function should be called before any other functions
+#[unsafe(no_mangle)]
+pub extern "C" fn cedarling_init() -> c_int {
+    // Initialize any global state if needed
+    0 // Success
+}
```
tareknaser, line 70: I am not sure what the intent of this function is but it currently does nothing and always returns success. The Tokio runtime is initialized lazily on first use so calling `cedarling_init()` has no effect.

If runtime creation fails for some reason, the error shows up during `cedarling_new` instead of at init time which I think makes it harder to diagnose startup failures.

Maybe we can force initialize the runtime here: `let _ = &*CEDARLING_RUNTIME;` and wrap it in `catch_unwind` to handle potential panics.

jans-cedarling/bindings/cedarling_c/src/types.rs
```diff
@@ -0,0 +1,214 @@
...
+    pub fn new(strings: Vec<String>) -> Self {
+        if strings.is_empty() {
+            return CedarlingStringArray {
+                items: ptr::null_mut(),
+                count: 0,
+            };
+        }
+        let mut c_strings: Vec<*mut c_char> = Vec::with_capacity(strings.len());
+
+        for s in strings.clone() {
```
tareknaser: Seems to be an unnecessary `Vec::clone()`. this clones all strings and the vec itself just to iterate. The original `strings` is only used for `.len()` afterward.

Let's save the length first, then consume the Vec directly:

```rust
let count = strings.len();
for s in strings {
    // ...
}
// use count instead of strings.len()
```

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
```diff
@@ -0,0 +1,983 @@
...
+        },
+        Err(code) => {
+            return code;
+        },
+    };
+
+    runtime.runtime.block_on(instance.shut_down());
+
+    CedarlingErrorCode::Success
+}
```
tareknaser: After `cedarling_shutdown(id)`, the instance stays in the HashMap and keeps working. Authorization calls succeed even after shutdown.

I wrote a simple test to verify this at runtime: `cedarling_authorize_unsigned` returns success after `cedarling_shutdown`. Double shutdown is also silently accepted. 

Let me know if you want me to share the test code (I don't want to leave long code snippets in the comments).

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
```diff
@@ -0,0 +1,983 @@
...
+    let value: serde_json::Value = match serde_json::from_str(value_json) {
+        Ok(v) => v,
+        Err(e) => {
+            let error_msg = format!("Failed to parse value JSON: {}", e);
+            set_last_error(&error_msg);
+            return CedarlingResult::error(CedarlingErrorCode::JsonError, &error_msg);
+        },
+    };
+
+    match instance.push_data_ctx(key, value, None) {
```
tareknaser: C callers have no way to set TTL on context data. The Python, Go, and UniFFI bindings all expose a TTL parameter. The C binding hardcodes it to `None`.

Is there a reason for this I might be missing?

jans-cedarling/bindings/cedarling_c/tests/test_cedarling.c
```diff
@@ -0,0 +1,502 @@
+/*
```
tareknaser, line 1: Some functions have no tests like `cedarling_context_push`, `cedarling_context_get`, `cedarling_context_remove`, `cedarling_context_clear`, `cedarling_context_list`, `cedarling_context_stats`. 

These are some exported functions with no coverage at all.

jans-cedarling/bindings/cedarling_c/src/lib.rs
```diff
@@ -0,0 +1,720 @@
...
+        array_ref.items = ptr::null_mut();
+        array_ref.count = 0;
+    }
+}
+
+/// # Safety
+/// Free a CedarlingResult structure
+///
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn cedarling_free_result(result: *mut CedarlingResult) {
```
tareknaser, line 1073: Every `# Safety` section says what the function does not what the caller must guarantee. Rust convention is for `# Safety` to list preconditions the caller must uphold.
This also shows up in the generated C header as is which looks odd to C developers.

jans-cedarling/bindings/cedarling_c/src/lib.rs
```diff
@@ -0,0 +1,720 @@
...
+}
+
+/// # Safety
+/// Check whether a trusted issuer was loaded by issuer identifier
+///
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn cedarling_is_trusted_issuer_loaded_by_name(
+    instance_id: u64,
+    issuer_id: *const c_char,
+) -> bool {
```
tareknaser: These return `false` for both "issuer not loaded" and "instance doesn't exist" and "null pointer passed." The caller has to check `cedarling_get_last_error()` after every call to tell the difference.

I also wrote a test to verify this: `cedarling_is_trusted_issuer_loaded_by_name(99999, "x")` returns `false` with `get_last_error()` set to "Instance not found." Same ambiguity for `cedarling_is_trusted_issuer_loaded_by_iss`.

Let's change to `c_int` return with error codes or add a `bool* out_result` parameter.

jans-cedarling/bindings/cedarling_c/src/lib.rs
```diff
@@ -0,0 +1,720 @@
...
+        },
+    };
+    is_trusted_issuer_loaded_by_iss(instance_id, &iss_claim_str)
+}
+
+/// # Safety
+/// Get total number of trusted issuers discovered
+///
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn cedarling_total_issuers(instance_id: u64) -> usize {
```
tareknaser: Same comment.

jans-cedarling/bindings/cedarling_c/src/lib.rs
```diff
@@ -0,0 +1,720 @@
...
+    total_issuers(instance_id)
+}
+
+/// # Safety
+/// Get number of trusted issuers loaded successfully
+///
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn cedarling_loaded_trusted_issuers_count(instance_id: u64) -> usize {
+    loaded_trusted_issuers_count(instance_id)
+}
```
tareknaser: Same comment. These functions return 0 for both errors and zero count

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
```diff
@@ -0,0 +1,983 @@
...
+
+fn get_instance_id() -> u64 {
+    static INSTANCE_ID: AtomicU64 = AtomicU64::new(1);
+    INSTANCE_ID.fetch_add(1, Ordering::SeqCst)
+}
+
+/// Runtime manager for Cedarling instances
+struct CedarlingRuntime {
+    runtime: Arc<Runtime>,
+    instances: Mutex<HashMap<u64, Arc<base::Cedarling>>>,
```
tareknaser: The C binding wraps the async `Cedarling` and calls `block_on()` for every operation. But `blocking::Cedarling` already exists in the core library and calls the authorization engine directly without any async overhead.

`blocking::Cedarling::authorize_unsigned` at `cedarling/src/blocking.rs:57` calls `self.instance.authz.authorize_unsigned(&request)` directly. No `block_on` or Tokio scheduling. The async versions are marked `#[allow(clippy::unused_async)]` because they don't actually await anything.

Let's replace `Arc<base::Cedarling` with `blocking::Cedarling` to eliminate the `block_on` overhead on the hot authorization path (every query in the Postgres extension) and keep `block_on` only for instance creation and shutdown where it's actually needed.

jans-cedarling/bindings/cedarling_c/src/types.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// Memory allocation error
+    MemoryError = 6,
+    /// Internal error
+    Internal = 7,
+}
+
+/// Result structure for operations that return data
+#[repr(C)]
+pub struct CedarlingResult {
+    /// Error code (0=Success))
```
tareknaser: Just a typo
```suggestion
    /// Error code (0=Success)
```

jans-cedarling/bindings/cedarling_c/src/types.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// Array of string pointers
+    pub items: *mut *mut c_char,
+    /// Number of items in the array
+    pub count: usize,
+}
+
+/// Structure for instance creation results
+#[repr(C)]
+pub struct CedarlingInstanceResult {
+    /// Error code (0=Success))
```
tareknaser:
```suggestion
    /// Error code (0=Success)
```

jans-cedarling/bindings/cedarling_c/src/types.rs
```diff
@@ -0,0 +1,214 @@
...
+pub struct CedarlingResult {
+    /// Error code (0=Success))
+    pub error_code: CedarlingErrorCode,
+    /// Data pointer (null if error)
+    pub data: *mut c_char,
+    /// Error message (null if success)
+    pub error_message: *mut c_char,
+}
+
+///Structure for returning arrays of strings
```
tareknaser:
```suggestion
/// Structure for returning arrays of strings
```

jans-cedarling/bindings/cedarling_c/src/lib.rs
```diff
@@ -0,0 +1,720 @@
...
+use std::os::raw::{c_char, c_int};
+use std::ptr;
+
+mod c_interface;
+mod types;
+
+use c_interface::*;
+use types::*;
+/// # Safety
+/// Intialize the Cedarling library
```
tareknaser:
```suggestion
/// Initialize the Cedarling library
```

jans-cedarling/bindings/cedarling_c/src/types.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// Instance not found
+    InstanceNotFound = 2,
+    /// JSON parsing error
+    JsonError = 3,
+    /// Authorization error
+    AuthorizationError = 4,
+    /// Configuration error
+    ConfigurationError = 5,
+    /// Memory allocation error
+    MemoryError = 6,
```
tareknaser: `MemoryError` seems to be never actually used in the code

jans-cedarling/bindings/cedarling_c/src/types.rs
```diff
@@ -0,0 +1,214 @@
...
+    }
+}
+
+#[allow(unused)]
+pub fn string_to_c_string(s: &str) -> *mut c_char {
+    match CString::new(s) {
+        Ok(c_string) => c_string.into_raw(),
+        Err(_) => ptr::null_mut(),
+    }
+}
```
tareknaser: This is basically deadcode. Not sure why we are keeping it

jans-cedarling/bindings/cedarling_c/cbindgen.toml
```diff
@@ -0,0 +1,30 @@
...
+style = "both"
+
+[export]
+include=["CedarlingErrorCode","CedarlingStringArray","CedarlingResult","CedarlingInstanceResult"]
+exclude=[]
+
+[fn]
+postfix = ""
+args="horizontal"
+rename_args="GeckoCase"
```
tareknaser: GeckoCase is a just a mozilla convention that's unfamiliar to most C developers. Let's change to `SnakeCase`

jans-cedarling/bindings/cedarling_c/cbindgen.toml
```diff
@@ -0,0 +1,30 @@
...
+
+
+[struct]
+rename_fields = "ScreamingSnakeCase"
```
tareknaser: `ScreamingSnakeCase` for struct fields is typically reserved for macros and constants in C. We can also use `snake_case` here

jans-cedarling/bindings/cedarling_c/cbindgen.toml
```diff
@@ -0,0 +1,30 @@
+language="C"
```
tareknaser, line 1: The generated `cedarling_c.h` has no `#ifndef`/`#define` include guard. Let's add it here

```toml
include_guard = "CEDARLING_C_H"
```

jans-cedarling/bindings/cedarling_c/src/lib.rs
```diff
@@ -0,0 +1,720 @@
...
+    let instance_result = create_instance_with_env(config_str.as_deref());
+    unsafe { *result = instance_result };
+    unsafe { (*result).error_code as c_int }
+}
+/// # Safety
+/// Drop a cedarling instance
+///
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn cedarling_drop(instance_id: u64) {
+    drop_instance(instance_id);
```
tareknaser: `drop_instance` returns a `bool` but the return value is discarded but `cedarling_shutdown` returns `c_int` for the same kind of operation which is inconsistent.

Let's return `c_int` like shutdown does

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
```diff
@@ -0,0 +1,983 @@
...
+        .map(|rt| CedarlingRuntime {
+            runtime: Arc::new(rt),
+            instances: Mutex::new(HashMap::new()),
+        })
+        .map_err(|e| format!("Internal runtime initialization failure: {}", e))
+});
+
+fn get_instance_id() -> u64 {
+    static INSTANCE_ID: AtomicU64 = AtomicU64::new(1);
+    INSTANCE_ID.fetch_add(1, Ordering::SeqCst)
```
tareknaser: `Ordering::SeqCst` is probably stronger than needed for the ID counter
For a monotonic counter that just needs uniqueness, `Relaxed` is sufficient. I think that `SeqCst` adds unnecessary global ordering overhead.

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
olehbozhok: I agree

jans-cedarling/bindings/cedarling_c/tests/test_cedarling.c
```diff
@@ -0,0 +1,674 @@
...
+    TEST_ASSERT(ret!=0,"Reject NULL result pointer");
+
+    // Test invalid instance ID
+    ret=cedarling_authorize_unsigned(99999,valid_request,&auth_result);
+    TEST_ASSERT(ret!=0,"Reject invalid instance ID");
+    cedarling_free_result(&auth_result);
+
+    // Clean up
+    cedarling_drop(instance_id);
+}
```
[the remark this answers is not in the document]
haileyesus2433: Adding comprehensive authorization tests for cedarling_authorize_multi_issuer requires valid JWT tokens and a matching policy store configuration. The existing test_cedarling.c doesn't have this infrastructure. Writing such a test would require:
Generating or embedding valid JWTs with proper signatures
A policy store that recognizes those issuers
Network/JWKS infrastructure or mock setup
This is an integration test concern, not something for a unit test file that currently focuses on API lifecycle and error handling. The authorize_multi_issuer.c example already demonstrates the call pattern. Adding a brittle test with hardcoded JWTs would be worse than no test.

jans-cedarling/bindings/cedarling_c/src/types.rs
```diff
@@ -0,0 +1,202 @@
...
+    if c_str.is_null() {
+        return Err(CedarlingErrorCode::InvalidArgument);
+    }
+    unsafe {
+        match CStr::from_ptr(c_str).to_str() {
+            Ok(s) => Ok(s.to_string()),
+            Err(_) => Err(CedarlingErrorCode::InvalidArgument),
+        }
+    }
+}
```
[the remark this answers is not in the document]
haileyesus2433: The function already contains the unsafe block internally and performs a null check before the unsafe dereference. It's pub (not pub(crate)) but it's in a crate that doesn't expose a Rust API — it's an internal helper within the C bindings crate. Making it pub unsafe fn would force every call site (30+ places in c_interface.rs) to wrap each call in unsafe { }, adding noise with no safety benefit since the callers already operate in an FFI context. The null-check guard makes the function safe to call. This is a common Rust FFI pattern.

jans-cedarling/bindings/cedarling_c/src/lib.rs
```diff
@@ -0,0 +1,1119 @@
...
+///
+/// # Postconditions
+/// No instances remain registered; the calling thread's `cedarling_get_last_error()` reads as
+/// null until a new error is set.
+#[unsafe(no_mangle)]
+pub extern "C" fn cedarling_cleanup() {
+    ffi_guard_void!({
+        let _ = cleanup_runtime();
+        // Keep historical behavior: reset caller-visible last error state.
+        clear_last_error();
```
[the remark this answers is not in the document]
haileyesus2433: this is by design. cedarling_cleanup is the final teardown call. Its documented postcondition explicitly states: "No instances remain registered; the calling thread's cedarling_get_last_error() reads as null until a new error is set." This is the expected behavior — after cleanup, the slate is wiped clean. If cleanup_runtime() fails because the mutex is poisoned, the program is already in a bad state. Trying to preserve that error for a caller who is tearing down the library anyway adds complexity for no practical value. The let _ = deliberately discards the error.

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
```diff
@@ -0,0 +1,804 @@
...
+    match instance.get_data_ctx(key) {
+        Ok(Some(value)) => match serde_json::to_string(&value) {
+            Ok(json) => CedarlingResult::success(json),
+            Err(e) => {
+                let error_msg = format!("Failed to serialize value: {}", e);
+                set_last_error(&error_msg);
+                CedarlingResult::error(CedarlingErrorCode::Internal, &error_msg)
+            },
+        },
+        Ok(None) => CedarlingResult::success("null".to_string()),
```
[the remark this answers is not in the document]
haileyesus2433: This is a theoretical concern. In practice, context_push stores arbitrary JSON values provided by the C caller. If a caller stores "null" as JSON, they chose to do so. The Ok(None) case means the key doesn't exist in the HashMap; Ok(Some(Value::Null)) means the key exists with value null. The AI wants these differentiated.
However, this is overkill for the C API. The current behavior (Ok(None) returns success with "null") is the same pattern used in the Python and Go bindings. Changing this would:
Break consistency with other bindings.
Add complexity to the C caller's logic for an edge case.

jans-cedarling/bindings/cedarling_c/examples/basic_example.c
```diff
@@ -0,0 +1,193 @@
...
+
+    // Drop the instance
+    printf("8. Dropping instance...\n");
+    cedarling_drop(instance_id);
+    printf("   Instance dropped\n\n");
+
+    // Cleanup
+    printf("9. Cleaning up...\n");
+    cedarling_cleanup();
+    printf("   Cleanup completed\n\n");
```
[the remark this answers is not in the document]
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/fbfc0e9a80a328c447ae18e525590a2d19cc42a7

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
```diff
@@ -0,0 +1,804 @@
...
+    };
+
+    instance.shut_down();
+
+    // Remove from the runtime map so further API calls fail with InstanceNotFound.
+    match runtime.drop_instance(instance_id) {
+        CedarlingErrorCode::Success => CedarlingErrorCode::Success,
+        CedarlingErrorCode::InstanceNotFound => CedarlingErrorCode::Success,
+        code => code,
+    }
```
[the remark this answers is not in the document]
haileyesus2433: resolved in https://github.com/JanssenProject/jans/commit/fbfc0e9a80a328c447ae18e525590a2d19cc42a7

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/fb111eafa75df73898d0427ce01852231664ba7e

jans-cedarling/bindings/cedarling_c/src/types.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/fb111eafa75df73898d0427ce01852231664ba7e

jans-cedarling/bindings/cedarling_c/src/types.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/fb111eafa75df73898d0427ce01852231664ba7e

jans-cedarling/bindings/cedarling_c/src/types.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/fb111eafa75df73898d0427ce01852231664ba7e

jans-cedarling/bindings/cedarling_c/src/lib.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/fb111eafa75df73898d0427ce01852231664ba7e

jans-cedarling/bindings/cedarling_c/cbindgen.toml
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/82b5fcf0e6bf37de57badc7b60c63439b6ef3ab6

jans-cedarling/bindings/cedarling_c/cbindgen.toml
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/82b5fcf0e6bf37de57badc7b60c63439b6ef3ab6

jans-cedarling/bindings/cedarling_c/cbindgen.toml
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/82b5fcf0e6bf37de57badc7b60c63439b6ef3ab6

jans-cedarling/bindings/cedarling_c/src/lib.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/8bafab32814caf07278d0478f9e3891b4c1b3262

jans-cedarling/bindings/cedarling_c/src/lib.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/5a8a7718192060efec6eba9022895d8d951c0933

jans-cedarling/bindings/cedarling_c/src/types.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/fb111eafa75df73898d0427ce01852231664ba7e

jans-cedarling/bindings/cedarling_c/src/lib.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/838fc46d43c1e013f9e28949f3ac51847b755727

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/ebd28e8e20d21402ee754a20844f23c5b012034b

jans-cedarling/bindings/cedarling_c/src/lib.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/ed44745c3782a0e31cc213da9f84762fab50e627

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/56099686943f0af119fa1dd059c53d892f1c6d60

jans-cedarling/bindings/cedarling_c/src/lib.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/7c6422a79291c82396dc76a3d8e60cf649cca1c5

jans-cedarling/bindings/cedarling_c/src/types.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/fb111eafa75df73898d0427ce01852231664ba7e

jans-cedarling/bindings/cedarling_c/src/types.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/commit/fb111eafa75df73898d0427ce01852231664ba7e

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/d577b8ff5f6d9a04514aefff1b4275fd3e43e67c

jans-cedarling/bindings/cedarling_c/src/types.rs
haileyesus2433: lifetime would complicate things and ur solution seems enough, its resolved in here https://github.com/JanssenProject/jans/pull/13542/commits/1d5fcc7b5ade81353bc2c9bfefc7c50aa2dd7633

jans-cedarling/bindings/cedarling_c/src/lib.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/e09857e1e6d737fccc3efc74311eda7d58a4924f

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/9f5704e5ac92eb021bda191a3a2dd2a4a99d9242

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/ebbe2cbb088276d70255c5914f049566cd21f97d

jans-cedarling/bindings/cedarling_c/tests/test_cedarling.c
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/5146201300843d10594c056c9b2d37a94e87209e

jans-cedarling/bindings/cedarling_c/src/lib.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/e4f410ee7a9c9be37e7c4b2c16c590ac8a8e65db

jans-cedarling/bindings/cedarling_c/src/lib.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/28cf253a2f027ec1a12ac3a256fe5a4fe6af8275

jans-cedarling/bindings/cedarling_c/src/lib.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/28cf253a2f027ec1a12ac3a256fe5a4fe6af8275

jans-cedarling/bindings/cedarling_c/src/lib.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/28cf253a2f027ec1a12ac3a256fe5a4fe6af8275

jans-cedarling/bindings/cedarling_c/src/lib.rs
```diff
@@ -0,0 +1,1119 @@
...
+            if !array_ref.items.is_null() {
+                for i in 0..array_ref.count {
+                    let item_ptr = *array_ref.items.add(i);
+                    if !item_ptr.is_null() {
+                        // Convert the raw pointer back to a CString and drop it
+                        let _ = std::ffi::CString::from_raw(item_ptr);
+                    }
+                }
+                // Free the array of pointers
+                let _ = Vec::from_raw_parts(array_ref.items, array_ref.count, array_ref.count);
```
tareknaser: later in `jans-cedarling/bindings/cedarling_c/src/lib.rs`, we do
```rust
let _ = Vec::from_raw_parts(array_ref.items, array_ref.count, array_ref.count);
```

`Vec::with_capacity(N)` guarantees at least N slots but the allocator may round up and give you more. Reconstructing with `Vec::from_raw_parts(ptr, N, N)` passes the wrong capacity.

Per [`Vec::from_raw_parts` safety docs:](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts) "capacity needs to be the capacity that the pointer was allocated with."

I think that in practice allocators handle this gracefully but it's technically UB.

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
```diff
@@ -0,0 +1,809 @@
...
+            CedarlingErrorCode::Internal,
+            "Internal lock failure while adding instance",
+        );
+    }
+
+    CedarlingInstanceResult::success(instance_id)
+}
+
+/// Drop a Cedarling Instance
+pub fn drop_instance(instance_id: u64) -> CedarlingErrorCode {
```
tareknaser, line 218: `shutdown_instance` properly calls `instance.shut_down()` then removes from the HashMap but `drop_instance` only removes. The `shut_down()` method flushes buffered logs to the Lock Server and closes the connection gracefully. A C caller using `cedarling_drop` instead of `cedarling_shutdown` silently loses buffered log data.

I was comparing this behaviour to other languages' bindings (at `jans-cedarling/bindings/`) and it seems that it's not handled automatically in any of them. The core `blocking::Cedarling` has no `Drop` impl that calls `shut_down()` so all bindings rely on the caller to shut down explicitly.

I am not sure if it's the desired behaviour but it seems to me that having both `drop` and `shutdown` invites misuse.

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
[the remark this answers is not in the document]
tareknaser: > the same pattern used in the Python and Go bindings

You're right that this is consistent across language bindings. I checked and Python, Go, and WASM all return the same thing for a missing key vs a stored JSON null.

That said, I still think it's worth noting as a design concern because it's inconsistent within the C API itself. `get_log_by_id` returns `KeyNotFound` for a missing entry but `context_get` returns success with `"null"`. 

also python and WASM work around this by also exposing `get_data_entry_ctx` which returns metadata so callers can tell if a key actually exists. The C bindings don't have that function yet so C callers have no way to resolve the ambiguity at all.

I'm not sure who to discuss this with but it might be worth updating this behavior across the other bindings too, not just C.

jans-cedarling/bindings/cedarling_c/src/types.rs
```diff
@@ -0,0 +1,202 @@
...
+            Ok(s) => s.into_raw(),
+            Err(_) => ptr::null_mut(),
+        };
+
+        CedarlingResult {
+            error_code: code,
+            data: ptr::null_mut(),
+            error_message: error_msg,
+        }
+    }
```
tareknaser, line 96: If an error message contains an interior NUL byte, `CString::new()` fails and `error_message` becomes NULL but `error_code` is still non-zero. This basically means that a C caller checking `if (result.error_message != NULL)` would miss these errors.

Same pattern is in `CedarlingInstanceResult::error()`

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
```diff
@@ -0,0 +1,809 @@
...
+                let error_msg = format!("Failed to serialize response: {}", e);
+                set_last_error(&error_msg);
+                CedarlingResult::error(CedarlingErrorCode::Internal, &error_msg)
+            },
+        },
+        Err(e) => {
+            let error_msg = format!("Authorization failed: {}", e);
+            set_last_error(&error_msg);
+            CedarlingResult::error(CedarlingErrorCode::AuthorizationError, &error_msg)
+        },
```
tareknaser, line 261: Every error path writes the same message to both `set_last_error()` and the result struct's `error_message`. This means C callers have three places to check for errors, and two of them always say the same thing.

```c
CedarlingResult r;
int ret = cedarling_authorize_unsigned(id, "{bad json}", &r);

char* last_err = cedarling_get_last_error();

printf("return code:     %d\n", ret);
printf("error_code:      %d\n", r.error_code);
printf("error_message:   %s\n", r.error_message);
printf("get_last_error:  %s\n", last_err);

// output:
//   return code:     3
//   error_code:      3
//   error_message:   Failed to parse request JSON: key must be a string at line 1 column 2
//   get_last_error:  Failed to parse request JSON: key must be a string at line 1 column 2

cedarling_free_string(last_err);
cedarling_free_result(&r);
```

`error_message` and `get_last_error` are identical. The caller has to free both.

It's also inconsistent across function types. Functions returning `CedarlingResult` have all three channels. Functions returning `CedarlingStringArray` only have the return code and `get_last_error` (no `error_message` field) so callers need different error-checking patterns depending on which function they called.

I actually checked the other bindings and none of them have this problem. They all have one error channel.

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
```diff
@@ -0,0 +1,809 @@
...
+/// Get context data by key
+pub fn context_get(instance_id: u64, key: &str) -> CedarlingResult {
+    clear_last_error();
+
+    let runtime = runtime_ref();
+
+    let instance = match runtime.get_instance(instance_id) {
+        Ok(Some(instance)) => instance,
+        Ok(None) => {
+            let error_msg = "Instance not found";
```
tareknaser: Python, UniFFI, Go, and WASM all expose `get_data_entry_ctx(key)` which returns the full `DataEntry` with creation time, expiry, and access count. The C binding only has `context_get` (raw value) and `context_list` (all entries). No way to inspect metadata for a single key.

jans-cedarling/bindings/cedarling_c/src/lib.rs
```diff
@@ -0,0 +1,1119 @@
...
+    };
+}
+
+/// # Safety
+/// Caller must pass valid pointers for any non-null pointer arguments.
+/// Any owned pointers returned by this API must be released with the matching `cedarling_free_*` function.
+/// Initialize the Cedarling library
+/// This Function should be called before any other functions
+#[unsafe(no_mangle)]
+pub extern "C" fn cedarling_init() -> c_int {
```
tareknaser: Why include `# Safety` sections on functions that aren't marked `unsafe`?

jans-cedarling/bindings/cedarling_c/src/lib.rs
```diff
@@ -0,0 +1,1119 @@
...
+    ffi_guard_int!({ initialize_runtime() as c_int })
+}
+/// # Safety
+/// Caller must pass valid pointers for any non-null pointer arguments.
+/// Any owned pointers returned by this API must be released with the matching `cedarling_free_*` function.
+/// Create a new Cedarling instance
+///
```
tareknaser, line 74: For `# Safetey` doc commments on `unsafe` functions, it seems that the same generic boilerplate is used. Let's update so that we document their actual preconditions (which pointers must be valid, what to free, etc).

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
haileyesus2433: > `shutdown_instance` properly calls `instance.shut_down()` then removes from the HashMap but `drop_instance` only removes. The `shut_down()` method flushes buffered logs to the Lock Server and closes the connection gracefully. A C caller using `cedarling_drop` instead of `cedarling_shutdown` silently loses buffered log data.
> 
> I was comparing this behaviour to other languages' bindings (at `jans-cedarling/bindings/`) and it seems that it's not handled automatically in any of them. The core `blocking::Cedarling` has no `Drop` impl that calls `shut_down()` so all bindings rely on the caller to shut down explicitly.
> 
> I am not sure if it's the desired behaviour but it seems to me that having both `drop` and `shutdown` invites misuse.

Cedarling separates handle removal from graceful shutdown because shut_down() does real work  it runs async teardown (flush buffered logs to the Lock Server, close connections) via block_on on the blocking client, which does not fit cleanly into Rust’s synchronous Drop trait doing blocking or lock-taking I/O in destructors risks deadlocks, panics, and unpredictable latency. The blocking Cedarling type is also cloneable, so “drop one handle” does not unambiguously mean “tear down the whole instance,” which makes implicit shutdown in Drop easy to get wrong. The log worker’s Drop therefore only warns when buffers remain, steering integrators toward an explicit shut_down() rather than implying delivery guarantees from memory management. what do u guys think @olehbozhok @dagregi

jans-cedarling/bindings/cedarling_c/src/lib.rs
haileyesus2433: am not sure about this because for arrays produced only by CedarlingStringArray::try_new, from_raw_parts(items, count, count) matches how that memory was created, so it is not violating the “wrong capacity” rule in practice.
try_new does this
```rust
        let count = strings.len();
        let mut c_strings: Vec<*mut c_char> = Vec::with_capacity(count);

        for s in strings {
            match CString::new(s) {
                Ok(c_str) => c_strings.push(c_str.into_raw()),
                // ...
            };
        }
        let items_ptr = c_strings.as_mut_ptr();
        std::mem::forget(c_strings);
        Ok(CedarlingStringArray {
            items: items_ptr,
            count,
        })
```

we start with length 0, capacity count, then push exactly count times. That never triggers a grow, so when the Vec is forgotten, len == capacity == count. Allocator rounding does not change the Vec’s logical capacity(); Rust still records the capacity it uses for that buffer (here, count). So the pointer was allocated as the backing store for a Vec<*mut c_char> with that capacity, and reconstructing with the same count for both length and capacity is consistent with from_raw_parts’ contract for this allocation path.
You’re right that the capacity should match what the pointer was originally allocated with but the idea that this is always wrong because with_capacity(N) might secretly allocate more memory isn’t correct.
i am not sure but i think in Rust, even if the allocator grabs a bit more memory behind the scenes, the Vec still treats its capacity as exactly N. It doesn’t expose that extra memory to you.

jans-cedarling/bindings/cedarling_c/src/lib.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/c98458f5b2e6c61ca2ed805ab3406cb26d461c91

jans-cedarling/bindings/cedarling_c/src/lib.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/c98458f5b2e6c61ca2ed805ab3406cb26d461c91

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/372ecd4c803a80bccd89c4d20ef5b71052d6df67

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
dagregi: I don't know much of C to comment and I maybe wrong but why not remove drop_instance here, the other bindings only expose shutdown from what I've seen and just like @haileyesus2433 said the Drop implementation on the log worker is only for logging a warning, so for consistency with the other bindings just using shut_down and removing drop_instance would be my guess unless I missed the point of the discussion

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
haileyesus2433: The C API uses three ways to report errors because C doesn’t have exceptions or `Result` like Rust: it returns an integer status, fills an output struct with `error_code` and `error_message`, and sometimes uses `cedarling_get_last_error()` for cases where there’s no struct (like arrays or simple values) or for basic FFI issues like null pointers. The integer return just mirrors the `error_code`, which is standard in C. To avoid redundancy, i have cleaned up the implementation  so that if an error is already included in a result struct, it’s only stored there and not duplicated in `get_last_error`. The thread-local error is also cleared when needed so old messages don’t stick around. The FFI layer now fills the output struct directly without copying the same error into `get_last_error`.  The docs also clarify that you should mainly rely on the return code and struct for results, and use `get_last_error` mostly for array or scalar-returning APIs.
check https://github.com/JanssenProject/jans/pull/13542/commits/d60ebb304a1c9eeea4d765b95ba2a0e5228379d1

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
olehbozhok: Yeah, it is a complicated question, and it is not the main priority for now

jans-cedarling/bindings/cedarling_c/src/types.rs
tareknaser: This is fixed. Thanks @haileyesus2433

jans-cedarling/bindings/cedarling_c/src/lib.rs
tareknaser: I understand the argument that vec records capacity as N but `RawVec::with_capacity` sets its internal capacity from the actual returned allocation size. not from the requested N. I also understand that in practice this should work fine on all common allocators but I think the fix is very minimal and makes the question go away entirely so it may be worth it to just use boxed slices which guarantee `capacity == len`:
 ```rust
 // Allocation:
 let boxed = c_strings.into_boxed_slice();
 let items_ptr = Box::into_raw(boxed) as *mut *mut c_char;

 // Deallocation:
 let _ = Box::from_raw(std::slice::from_raw_parts_mut(array_ref.items, array_ref.count));
 ```
 what do you think?

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
tareknaser: > just using shut_down and removing drop_instance

When it makes sense to revisit, that's probably the cleanest path forward
Marking this as resolved for now

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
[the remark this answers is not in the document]
tareknaser: @haileyesus2433 
Following up on this, I'd suggest adding a note in the header doc for `cedarling_context_get` that says something to the effect of: "Returns `"null"` both when the key is absent and when the stored value is JSON null. Use `cedarling_context_get_entry` to distinguish the two cases."

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
```diff
@@ -0,0 +1,843 @@
...
+        None
+    };
+
+    match instance.push_data_ctx(key, value, ttl) {
+        Ok(()) => CedarlingResult::success("{}".to_string()),
+        Err(e) => {
+            let error_msg = format!("Failed to push data: {}", e);
+            CedarlingResult::error(CedarlingErrorCode::Internal, &error_msg)
+        },
+    }
```
tareknaser, line 350: All context data functions map every `DataError` variant to `CedarlingErrorCode::Internal`. The C enum already has `InvalidArgument` and `KeyNotFound` codes but they're never used for DataApi errors. 

As an example, `cedarling_context_push(id, "", "{}", 0, &r)` returns `error_code=6` (INTERNAL) with message "Failed to push data: invalid key: key cannot be empty". Should return `error_code=1` (INVALID_ARGUMENT).

docs/cedarling/tutorials/c.md
```diff
@@ -0,0 +1,540 @@
...
+
+C bindings for the Jans Cedarling authorization engine, providing policy-based access control through a native C API.
+
+## Installation
+
+### Prerequisites
+
+- C compiler (GCC, Clang, or MSVC)
+- Rust toolchain
+- cbindgen (`cargo install cbindgen`)
```
tareknaser: The README lists cbindgen CLI as a prerequisite and has a manual header-generation step. The CI also installs the CLI binary. 
As far as I understand, neither is needed. `build.rs` uses cbindgen as a Rust crate dependency and generates the header automatically during `cargo build`.

haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/553f413932b6f000bb0d6d8b780d99181bb844e8

jans-cedarling/bindings/cedarling_c/src/c_interface.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/e3b4ce29aaffa0673f76224ed978a14efe296b5d

jans-cedarling/bindings/cedarling_c/src/lib.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13542/commits/932dc66aa07337905207a5706972a38db8d4e8b3

haileyesus2433:
@coderabbitai could u please approve if u dont have any suggestions

haileyesus2433:
> Thank you for the great work. I have no further comments

thank you verymuch for the reviews

haileyesus2433:
@coderabbitai approve

PR #13856, opened 2026-04-17
feat(jans-cedarling): Cedarling PostgreSQL Extension

### LINKED ISSUE #12071: feat(jans-cedarling): Cedarling PostgreSQL Extension
Currently, Cedarling cannot be used directly within PostgreSQL. Developers who want to leverage Cedarling’s fine-grained access control must implement external integration logic, which adds complexity and overhead. A native PostgreSQL extension would allow seamless usage of Cedarling policies directly in SQL queries and Row-Level Security (RLS).

Create a PostgreSQL extension using [pgrx](https://github.com/pgcentralfoundation/pgrx) that embeds Cedarling into the database.

- Expose SQL functions that evaluate Cedar policies against row data and JWT claims.

- Integrate with PostgreSQL Row-Level Security (RLS) to filter results transparently.

- Provide an ergonomic API for ORMs and application developers.

N/A

**Additional context**
N/A

haileyesus2433:
### Prepare

-------------------

#### Target issue
  
  
  
  
closes #12071

#### Implementation Details
  

-------------------

## Summary by CodeRabbit

* **New Features**
  * PostgreSQL extension: row-level authorization, predicate pushdown, JWT/multi‑issuer & unsigned auth helpers, masking (rules/preview/apply), policy register/use/rollback, schema validation, entity mapping, observability (traces/status/explain), GUC runtime controls, installer, Docker image, and CLI for generating Cedar schemas.

* **Documentation**
  * Full PostgreSQL integration guide with install options, examples, function reference, security and operational notes.

* **CI**
  * New Postgres extension test job and optional packaging/provenance workflow; workspace test/bench runs now exclude the extension by default.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JanssenProject/jans/pull/13856?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/4686cb91-fdce-4df1-bda5-6bf528c538a9?refs=fc0afde663aec87b342a884854c166132c4e4256&source=prChecksComment) |

jans-cedarling/cedarling_pg/src/catalog.rs
```diff
@@ -0,0 +1,98 @@
...
+COMMENT ON TABLE cedarling.policy_versions IS 'Named policy version registry for cedarling_use_policy / cedarling_register_policy_version.';
+",
+    name = "cedarling_pg_catalog",
+);
+
+extension_sql!(
+    r"
+REVOKE EXECUTE ON FUNCTION cedarling_use_policy(text) FROM PUBLIC;
+REVOKE EXECUTE ON FUNCTION cedarling_register_policy_version(text, text) FROM PUBLIC;
+REVOKE EXECUTE ON FUNCTION cedarling_rollback_policy() FROM PUBLIC;
```
tareknaser, line 110: `cedarling_diff_policies(text, text)` is missing from this REVOKE list. Any role can call it and read arbitrary files from disk.

Anyone who can run a query against a database with the extension installed can do something like `SELECT cedarling_diff_policies('/etc/passwd', '/etc/shadow')`. The function runs with the privileges of the postgres OS user so any file that user can read is reachable.

jans-cedarling/cedarling_pg/src/functions/authorized_row.rs
```diff
@@ -0,0 +1,336 @@
...
+/// Callers must apply column masking themselves by pairing this predicate with
+/// `cedarling_mask_row(t, '<table>')` in the SELECT list.
+#[pg_extern(volatile, parallel_restricted)]
+#[allow(clippy::needless_pass_by_value)] // `#[pg_extern]` maps Rust parameters from PostgreSQL call convention; JsonB is moved in.
+pub fn cedarling_authorized_row(
+    resource: pgrx::datum::JsonB,
+    action: Option<&str>,
+    context: Option<pgrx::datum::JsonB>,
+) -> bool {
+    let action = action.unwrap_or("Read");
```
tareknaser: a SQL `NULL` action silently becomes `"Read"` here. 
The SQL signature isn't STRICT, so `cedarling_authorized_row(t, NULL, NULL)` reaches Rust as `None` and lands on this default. 

The docs say "Omitting `action` in SQL is invalid..the Rust default `'Read'` applies only when callers use a wrapper that supplies it" so this is doc/code drift  
it's also the surprising direction. a caller building the action from a variable that turns out to be NULL  gets read-authorized

jans-cedarling/cedarling_pg/src/authz/where_clause.rs
```diff
@@ -0,0 +1,875 @@
...
+    if t.eq_ignore_ascii_case("false") {
+        return Some("FALSE".to_string());
+    }
+    if t.parse::<i64>().is_ok() || t.parse::<f64>().is_ok() {
+        return Some(t.to_string());
+    }
+    if t.starts_with('"') && t.ends_with('"') && t.len() >= 2 {
+        let inner = &t[1..t.len() - 1];
+        return Some(quote_literal_safe(inner));
+    }
```
tareknaser, line 295: Cedar string literals can contain escapes like `\\`, `\"`, `\n`. This function strips the outer quotes and passes the raw inner bytes to `quote_literal_safe` so the SQL literal ends up with the escapes still in it. the row filter then doesn't match real column values.

e.g.,: a Cedar policy that reads
```cedar
permit(principal, action, resource) when { resource.note == "C:\\path" };
```

means the value `C:\path` (7 chars). But this lowering produces `"note" = 'C:\\path'`, and under `standard_conforming_strings = on` (the documented assumption in the comment above `quote_literal_safe`) that SQL literal is the 8-char string `C:\\path`. A row where `note` is literally `C:\path` won't match the predicate.

jans-cedarling/cedarling_pg/src/policy/versions.rs
```diff
@@ -0,0 +1,375 @@
...
+        Ok(prev) => prev,
+        Err(e) => {
+            let pe = PolicyError::Load(e.to_string());
+            extension_log::log_diagnostic(guc_config::CedarlingLogLevelGuc::Warn, &pe.to_string());
+            return false;
+        },
+    };
+
+    authz_cache::global_cache().clear_all();
+    let _ = set_policy_version_guc(&resolved.version_guc_value);
```
tareknaser: when `cedarling_use_policy` is called by a non-superuser, the engine swap and cache clear succeed but the `set_config` here fails silently because `cedarling.policy_version` is Suset. Engine state and GUC state diverge.

The doc says EXECUTE on `cedarling_use_policy` "is revoked from `PUBLIC` .. grant only to roles that may swap policy versions" which sounds to me like non-superuser roles are a supported caller but `cedarling.policy_version` is registered as `GucContext::Suset` (see `guc_config.rs`) and `set_policy_version_guc` calls `set_config('cedarling.policy_version', $1, false)`. 
For a non-superuser, that raises a permission error which the `let _ = …` discards.

docs/cedarling/integrations/postgres.md
```diff
@@ -0,0 +1,479 @@
...
+| `cedarling_authorize_unsigned(principal_json text, resource_json text, action text, context_json text) → bool` | Unsigned authorization (no tokens). |
+| `cedarling_authorized_row(record anyelement, action text, context jsonb) → bool` | RLS-friendly form: materializes the composite row, looks up its Cedar entity mapping, asks the engine. SQL has no `DEFAULT` on `context`; pass `NULL` for an empty object. |
+| `cedarling_authorized_row(resource jsonb, action text, context jsonb) → bool` | JSONB overload for callers that already have a Cedar `EntityData` document. Pass `NULL` for `{}`. |
+| `cedarling_authorized_row_jwt(record anyelement, action text) → bool` | Same as `cedarling_authorized_row` but uses `cedarling.tokens` to drive `authorize_multi_issuer`. Omitting `action` in SQL is invalid; the Rust default `'Read'` applies only when callers use a wrapper that supplies it. |
+| `cedarling_build_resource(record anyelement) → jsonb` | Returns the canonical Cedar `EntityData` JSON that `cedarling_authorized_row` would have produced — useful for debugging. Aborts the statement on invalid rows; do not use inside RLS policies. |
```
tareknaser: this row claims `cedarling_build_resource(record anyelement) → jsonb` but the committed SQL exposes two different functions and both return `text` not `jsonb`.

- `cedarling_build_resource(resource jsonb, entity_type text, entity_id text) → TEXT`
- `cedarling_build_resource_row(record anyelement) → TEXT`

jans-cedarling/cedarling_pg/src/mask/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+        Ok::<(), pgrx::spi::Error>(())
+    })
+    .is_ok()
+}
+
+/// Preview masking logic as a pure SQL function — useful for interactive testing.
+#[pg_extern(stable, parallel_safe)]
+pub fn cedarling_test_masking(
+    original_value: Option<&str>,
+    _data_type: Option<&str>,
```
tareknaser: the parameter is named `_data_type` here (so pgrx names the SQL argument `_data_type`), but the docs at `postgres.md` call it `data_type`. 


If the argument really is unused, maybe we can consider dropping it from the signature (and the docs) 
Or maybe you wanted to keep it as a pplaceholder for future use?

moabu:
@coderabbitai review

jans-cedarling/cedarling_pg/src/authz/bridge.rs
```diff
@@ -0,0 +1,344 @@
...
+    Principal(resource::ResourceEntityDataError),
+    #[error(transparent)]
+    Resource(#[from] resource::ResourceEntityDataError),
+    #[error(transparent)]
+    ContextParse(#[from] serde_json::Error),
+    #[error("Cedar request context must be a JSON object")]
+    ContextNotObject,
+    #[error(transparent)]
+    Authorize(#[from] cedarling::AuthorizeError),
+}
```
dagregi: Why not combine these two into a single enum?

jans-cedarling/cedarling_pg/src/mask/types.rs
```diff
@@ -0,0 +1,306 @@
...
+        assert!(matches!(m, MaskType::Range { min: 50_000, max: 150_000 }));
+    }
+
+    #[test]
+    fn from_parts_unknown_becomes_identity() {
+        assert!(matches!(MaskType::from_parts("bogus_type", None), MaskType::Identity));
+    }
+
+    #[test]
+    fn parse_range_bounds_negative_min() {
```
dagregi, line 318: Is it possible to use negative ranges for both min and max like (-20,-5), if it is it would be nice to add tests for it as well and update the function as well since it wont handle it correctly

dagregi (changes requested):
I will slowly add comments as I go through it and also one thing I noticed is the pg_codgen can traverse directories tables with names like `../../../table` will create the schema elsewhere even if the output is set to the current directory

jans-cedarling/cedarling_pg_codegen/src/main.rs
```diff
@@ -0,0 +1,168 @@
...
+        output.display()
+    );
+    for (table, r) in rendered {
+        if !r.unmapped_columns.is_empty() {
+            eprintln!(
+                "  - {} → {}: skipped unmapped columns ({})",
+                table,
+                r.entity_name,
+                r.unmapped_columns.join(", ")
+            );
```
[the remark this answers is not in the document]
haileyesus2433: this is a standalone host CLI; `eprintln!` is appropriate for stderr summary output

jans-cedarling/cedarling_pg/src/authz/where_clause.rs
```diff
@@ -0,0 +1,875 @@
...
+        match policies_to_sql_predicate(&policies, &table_mapping()) {
+            SqlPredicate::Where(sql) => {
+                assert!(sql.contains("\"age\" >= 18"));
+                assert!(sql.contains("\"country\" = 'CN'"));
+                assert!(sql.contains("NOT"));
+            },
+            other => panic!("expected WHERE predicate, got {other:?}"),
+        }
+    }
+}
```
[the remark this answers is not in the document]
haileyesus2433: this is fine and most of it has been adressed from commits [e313b3d](https://github.com/JanssenProject/jans/pull/13856/commits/e313b3d90416e730965028f0c2482008dd423181) - [de8ad65](https://github.com/JanssenProject/jans/pull/13856/commits/de8ad6567aa04852821584a4a5106bb87fdb098d)

jans-cedarling/cedarling_pg/src/functions/authorized.rs
```diff
@@ -0,0 +1,392 @@
...
+            if id.is_empty() { None } else { Some(id) }
+        });
+    let shadow = matches!(guc_config::mode(), CedarlingMode::Shadow);
+
+    status::record_request();
+
+    let principal_norm = principal_json
+        .map(str::trim)
+        .filter(|s| !s.is_empty())
+        .unwrap_or("");
```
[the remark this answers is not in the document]
haileyesus2433: Already handled by `parse_optional_principal_json` in the authz bridge.

jans-cedarling/cedarling_pg/src/tokens/bundle.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    /// JWT body into the resulting error's Display. The mapping key is the
+    /// only field a `NonStringPayload` error can include, and a raw JWT
+    /// passed as the whole input is rejected by serde with a "expected value"
+    /// message, not by quoting the input.
+    #[test]
+    fn errors_do_not_echo_raw_jwt_bytes() {
+        let fake_jwt = "eyJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJsZWFreSJ9.sig_redacted";
+        // Pass the raw JWT (not a bundle) — serde rejects it as invalid JSON.
+        let err = parse_token_inputs_from_json(fake_jwt).unwrap_err();
```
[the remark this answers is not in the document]
haileyesus2433: Intentional fixture verifying JWT bytes are not echoed in error messages.

jans-cedarling/cedarling_pg_codegen/src/main.rs
```diff
@@ -0,0 +1,168 @@
...
+        }
+        fs::create_dir_all(&args.output)
+            .with_context(|| format!("creating output dir {}", args.output.display()))?;
+        let mut rendered: Vec<(String, EntityRender)> = Vec::new();
+        for table in &tables {
+            let cols = fetch_columns(&mut client, &args.schema, table)?;
+            let render = render_entity(table, &cols);
+            let file = wrap_namespace(&args.namespace, std::slice::from_ref(&render.cedar_text));
+            let out = args.output.join(format!("{table}.cedarschema"));
+            write_output(&out, &file)?;
```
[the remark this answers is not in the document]
haileyesus2433: Table names come from `pg_class.relname` in a named schema; path traversal via normal
identifiers is not realistic.

olehbozhok:
@coderabbitai resume

jans-cedarling/cedarling_pg/src/authz/bridge.rs
olehbozhok: It is OK to have different errors for different methods, but the name should be unambiguous.

docs/cedarling/integrations/postgres.md
```diff
@@ -0,0 +1,507 @@
...
+| `cedarling_authorized_row_jwt(record anyelement, action text) → bool` | Same as `cedarling_authorized_row` but uses `cedarling.tokens` to drive `authorize_multi_issuer`. Omitting `action` in SQL is invalid; the Rust default `'Read'` applies only when callers use a wrapper that supplies it. |
+| `cedarling_build_resource_row(record anyelement) → text` | Materializes a composite row into the canonical Cedar `EntityData` JSON string that `cedarling_authorized_row` would use — useful for debugging. Aborts the statement on invalid rows; do not use inside RLS policies. |
+| `cedarling_build_resource(resource jsonb, entity_type text, entity_id text) → text` | Builds `EntityData` JSON from an existing JSONB document; optional `entity_type` / `entity_id` override or inject `cedar_entity_mapping`. Same abort-on-error semantics as the row variant. |
+| `cedarling_where(table_name text, action text, tokens text) → text` | Predicate pushdown: lowers matching Cedar policies into a SQL `WHERE` fragment. Falls back to `'TRUE'` (with a `WARN` listing unhandled policy ids) when a policy can't be lowered, and to `'FALSE'` on parse/engine errors. |
```
olehbozhok: `Falls back to `'TRUE'` - it looks unsafe

olehbozhok: Maybe better to have a setting parameter that allows specifying that?

docs/cedarling/integrations/postgres.md
```diff
@@ -0,0 +1,507 @@
...
+  that invariant regresses.
+- **Fail-safe by default.** Every error path in `enforcement` mode returns
+  `false`. `shadow` mode always returns `true` and records a trace.
+- **`fail_mode = open` is a deliberate escape hatch.** When the engine is
+  down or a request fails validation, `open` returns `true` instead of
+  denying. That is one GUC change away from bypassing authorization for
+  every row. Keep the default `closed` in production; treat `open` as a
+  break-glass setting with change control.
+- **Fail-open audit visibility depends on `cedarling.log_level`.** When
+  `fail_mode = open` converts an error into an allow, the extension emits
```
olehbozhok: Maybe better to have audit log not depend on `cedarling.log_level`?

olehbozhok:
maybe it makes sense to add a Docker file with build extention

```
FROM rust:1.95-bookworm AS builder

RUN apt-get update && apt-get install -y --no-install-recommends \
    libclang-dev \
    && rm -rf /var/lib/apt/lists/*

ENV LIBCLANG_PATH=/usr/lib/llvm-14/lib

RUN cargo install --locked cargo-pgrx --version 0.18.0
RUN cargo pgrx init --pg16 download

ENV PATH="/root/.pgrx/16.14/pgrx-install/bin:${PATH}"

WORKDIR /build
COPY Cargo.toml Cargo.lock ./

COPY cedarling/ cedarling/
COPY cedarling_pg/ cedarling_pg/
COPY cedarling_pg_codegen/ cedarling_pg_codegen/
COPY sparkv/ sparkv/
COPY http_utils/ http_utils/
COPY test_utils/ test_utils/
COPY bindings/ bindings/

RUN cargo pgrx install --release --manifest-path cedarling_pg/Cargo.toml --features pg16

FROM postgres:16-bookworm

COPY --from=builder \
    /root/.pgrx/16.14/pgrx-install/lib/postgresql/cedarling_pg.so \
    /usr/lib/postgresql/16/lib/cedarling_pg.so

COPY --from=builder \
    /root/.pgrx/16.14/pgrx-install/share/postgresql/extension/cedarling_pg.control \
    /usr/share/postgresql/16/extension/cedarling_pg.control

COPY --from=builder \
    /root/.pgrx/16.14/pgrx-install/share/postgresql/extension/cedarling_pg--0.1.0.sql \
    /usr/share/postgresql/16/extension/cedarling_pg--0.1.0.sql

```

jans-cedarling/cedarling_pg/src/catalog.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13856/commits/4ce43e39de7aeb22b14e89735704c0e068ee7e38

jans-cedarling/cedarling_pg/src/functions/authorized_row.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13856/commits/de8fa44bee7d836b8a4da86462df73472406bbe0

jans-cedarling/cedarling_pg/src/authz/where_clause.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13856/commits/dc4f9800e1107e0e15a71532ce67ab3502c15624

jans-cedarling/cedarling_pg/src/policy/versions.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13856/commits/c7c27aacd8dacee83dcf3935ad2dcd3be15c3dbb

docs/cedarling/integrations/postgres.md
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13856/commits/ac4391a872c4ed7cd804c6749d93fa0cce03daa2

jans-cedarling/cedarling_pg/src/mask/mod.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13856/commits/57620a52f9720dd7fae9688aff574db5fc7de5f4

jans-cedarling/cedarling_pg/src/authz/bridge.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13856/commits/0fe3741f487a01b377fe8385c616b3f70623aedc, @dagregi also if i combine those callers would have to handle all the errors they wouldn't need but to clarify for ambiguity i have renamed it

docs/cedarling/integrations/postgres.md
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13856/commits/b5569291a0456b219648f8affa0d54576f383d0b

docs/cedarling/integrations/postgres.md
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13856/commits/fcf323d817d9ac4c405a77af5182aa860ca86407

jans-cedarling/cedarling_pg/scripts/install.sh
```diff
@@ -0,0 +1,324 @@
...
+    local pg_config
+    pg_config="$(resolve_pg_config_for_source "${pg_version}")"
+    local pg_config_major
+    pg_config_major="$(detect_pg_major "${pg_config}")"
+    local pg_version_major="${pg_version#pg}"
+    if [[ "${pg_config_major}" != "${pg_version_major}" ]]; then
+        warn "PG_VERSION=${pg_version} but ${pg_config} reports PostgreSQL ${pg_config_major}; continuing because PG_CONFIG/PATH was explicit"
+    fi
+
+    local psql_bin="${PSQL:-$(dirname "${pg_config}")/psql}"
```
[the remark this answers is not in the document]
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13856/commits/2bd4656fc5f435f2853cf00ef8cb3ca4e45dfbfe

jans-cedarling/cedarling_pg/src/test_support.rs
```diff
@@ -0,0 +1,64 @@
...
+use std::io::Write;
+use std::path::{Path, PathBuf};
+
+/// Emit a single-quoted YAML scalar, escaping embedded `'` as `''`.
+fn yaml_single_quoted_scalar(value: &str) -> String {
+    format!("'{}'", value.replace('\'', "''"))
+}
+
+/// Write a minimal bootstrap config wired to a local policy-store file.
+pub(crate) fn write_bootstrap_yaml(dir: &Path, policy_path: &Path, app_name: &str) -> PathBuf {
```
olehbozhok, line 27: function `write_bootstrap_yaml` is never used

haileyesus2433:
> maybe it makes sense to add a Docker file with build extention
> 
> ```
> FROM rust:1.95-bookworm AS builder
> 
> RUN apt-get update && apt-get install -y --no-install-recommends \
>     libclang-dev \
>     && rm -rf /var/lib/apt/lists/*
> 
> ENV LIBCLANG_PATH=/usr/lib/llvm-14/lib
> 
> RUN cargo install --locked cargo-pgrx --version 0.18.0
> RUN cargo pgrx init --pg16 download
> 
> ENV PATH="/root/.pgrx/16.14/pgrx-install/bin:${PATH}"
> 
> WORKDIR /build
> COPY Cargo.toml Cargo.lock ./
> 
> COPY cedarling/ cedarling/
> COPY cedarling_pg/ cedarling_pg/
> COPY cedarling_pg_codegen/ cedarling_pg_codegen/
> COPY sparkv/ sparkv/
> COPY http_utils/ http_utils/
> COPY test_utils/ test_utils/
> COPY bindings/ bindings/
> 
> RUN cargo pgrx install --release --manifest-path cedarling_pg/Cargo.toml --features pg16
> 
> FROM postgres:16-bookworm
> 
> COPY --from=builder \
>     /root/.pgrx/16.14/pgrx-install/lib/postgresql/cedarling_pg.so \
>     /usr/lib/postgresql/16/lib/cedarling_pg.so
> 
> COPY --from=builder \
>     /root/.pgrx/16.14/pgrx-install/share/postgresql/extension/cedarling_pg.control \
>     /usr/share/postgresql/16/extension/cedarling_pg.control
> 
> COPY --from=builder \
>     /root/.pgrx/16.14/pgrx-install/share/postgresql/extension/cedarling_pg--0.1.0.sql \
>     /usr/share/postgresql/16/extension/cedarling_pg--0.1.0.sql
> ```

docker files added in here https://github.com/JanssenProject/jans/pull/13856/commits/a83cd6d9f61b7e21f4f0b95c0d1e08c6cec3f068 - https://github.com/JanssenProject/jans/pull/13856/commits/383c70cff17ec89b9721b6332fbbc53fc556f0ec

jans-cedarling/cedarling_pg/Dockerfile
```diff
@@ -0,0 +1,112 @@
...
+    PKGLIB="$(pg_config --pkglibdir)"; \
+    SHAREDIR="$(pg_config --sharedir)"; \
+    install -m 0755 /tmp/cedarling_pg_out/cedarling_pg.so      "${PKGLIB}/cedarling_pg.so"; \
+    install -m 0644 /tmp/cedarling_pg_out/cedarling_pg.control "${SHAREDIR}/extension/cedarling_pg.control"; \
+    install -m 0644 /tmp/cedarling_pg_out/cedarling_pg--*.sql  "${SHAREDIR}/extension/"; \
+    rm -rf /tmp/cedarling_pg_out
+
+# Auto-create the extension on first start. Operators still need to set
+# cedarling.bootstrap_config inside the container before any authorize call.
+COPY cedarling_pg/scripts/docker-initdb.sh /docker-entrypoint-initdb.d/10-cedarling_pg.sh
```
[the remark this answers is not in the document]
haileyesus2433: this is fine

jans-cedarling/cedarling_pg/Dockerfile
```diff
@@ -0,0 +1,112 @@
...
+
+# The workspace `members` list globs `bindings/*` and names
+# `cedarling_pg_codegen`; cargo refuses to load the workspace if those
+# manifests are missing. Strip them so the slim COPY above is enough.
+RUN sed -i \
+        -e 's|"bindings/\*", *||g' \
+        -e 's|, *"bindings/\*"||g' \
+        -e 's|"cedarling_pg_codegen", *||g' \
+        -e 's|, *"cedarling_pg_codegen"||g' \
+        Cargo.toml
```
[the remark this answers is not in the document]
haileyesus2433: this is also fine

jans-cedarling/cedarling_pg/Dockerfile
```diff
@@ -0,0 +1,112 @@
...
+#  - openssl-dev: TLS dependencies
+#  - postgresql${PG_MAJOR}-dev: server headers + pg_config matching the runtime
+#  - bash: cargo pgrx init runs a bash script
+RUN apk add --no-cache \
+        build-base musl-dev linux-headers \
+        clang-dev llvm-dev \
+        pkgconf perl \
+        openssl-dev \
+        postgresql${PG_MAJOR}-dev \
+        bash git
```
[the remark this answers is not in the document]
haileyesus2433: this is fine, since its for development we will always have the latests versions that way we will be able to see if there are breaking chnages in the latest versions if things dont work

jans-cedarling/cedarling_pg/Dockerfile
```diff
@@ -0,0 +1,112 @@
...
+# Run:
+#
+#   docker run --rm -e POSTGRES_PASSWORD=pw -p 5432:5432 cedarling_pg:dev-pg16
+#
+# Operators still need to set `cedarling.bootstrap_config` to a path inside the
+# container before any authorize call will succeed; see the integration docs.
+
+ARG RUST_VERSION=1.95.0
+ARG ALPINE_VERSION=3.21
+ARG PG_MAJOR=17
```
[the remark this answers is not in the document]
haileyesus2433: this is also fine

jans-cedarling/cedarling_pg/src/test_support.rs
haileyesus2433: its used by the pg_tests i have marked it in here https://github.com/JanssenProject/jans/pull/13856/commits/a4f0f2b8107bf2d5fdca8627d6d3da3ae1331ccc

olehbozhok (changes requested):
Here is a review from Claude. Can you pls analyze each case?

# `cedarling_pg` Review — Docs, Security, Tests

Reviewed the `cedarling_pg` changes on this branch vs `main`: documentation, the Rust/pgrx extension source, generated SQL, install script, and the `#[pg_test]` suite.

**Overall:** solid, security-conscious work — fail-closed defaults, parameterized SPI everywhere, careful pure-Rust quoting of identifiers/literals, and a candid security section in the docs. The findings below are the real gaps worth addressing before merge.

---

## 1. Documentation

The docs are high quality and largely accurate (GUC table, function inventory, PG matrix 13–18, asset/provenance names all verified against code). Defects:

### 🔴 CRITICAL
- **`docs/cedarling/integrations/postgres.md:131-135` — the `cosign verify-blob` command will not run.** It only passes `--certificate-oidc-issuer-regexp` and omits a certificate-identity matcher (`--certificate-identity` / `--certificate-identity-regexp`), which `cosign verify-blob` requires. The command errors out, so the recommended verification step fails. Compare the complete, correct form in `docs/contribute/ci-cd/slsa-verification.md:720-724`.

### 🟠 IMPORTANT
- **`postgres.md:294,296` — wrong parameter type `regclass` vs `oid`.** `cedarling_validate_schema` and `cedarling_register_entity_map` actually take `oid`, not `regclass` (`sql/cedarling_pg--0.1.0.sql:372`, `:253`). A bare table-name literal will not auto-cast to `oid`; callers must write `'mytable'::regclass::oid` (as the code comment at `src/policy/schema.rs:87` states). As documented, calls fail.
- **`postgres.md:53` — non-callable example.** The flow narrative shows `cedarling_authorized_row(students, 'Read')` (2 args), but every overload requires 3 args `(record, action, context)` with no SQL `DEFAULT` (`sql/...sql:117-121`). This contradicts the doc's own reference row at `postgres.md:266`.

### 🟡 MINOR
- `cedarling-pg` artifact family is missing from the SLSA "Artifact Families" table (`slsa-verification.md:781-791`), even though `cedarling-pg.intoto.jsonl` is published (`build-packages.yml:1880`) and `postgres.md:529` tells users it exists.
- `cedarling_mask_row` parameter documented as `row` (`postgres.md:305`); actual name is `row_json` (`sql/...sql:230`). Breaks named-argument calls.
- The `instrumentation` value of `cedarling.mode` (`postgres.md:332`) is listed but never explained; code defines it as "evaluate and log, still return the decision" (`src/guc_config.rs:24-27`).

---

## 2. Security

### 🔴 CRITICAL
None.

### 🟠 HIGH

**H1 — Enforcement-affecting GUCs are `Userset`, so any role can weaken enforcement for its own session.**
`src/guc_config.rs:300-352`. `cedarling.mode`, `cedarling.fail_mode`, `cedarling.strategy`, and `cedarling.where_partial_fallback` are all registered `GucContext::Userset`. Any role with query access can run:
```sql
SET cedarling.mode = 'shadow';     -- finalize_decision() returns true unconditionally (authorized.rs:357)
SET cedarling.fail_mode = 'open';  -- finalize_error() returns true on any error (authorized.rs:366)
```
This neutralizes authorization for that session — directly impacting any RLS policy or `SECURITY DEFINER` view relying on `cedarling_authorized*`, since RLS predicates evaluate in the querying user's session.
**Fix:** register these as `GucContext::Suset` (matching `bootstrap_config`, `policy_version`, `mask_hash_salt` at `:449,459,467`).

**H2 — `cedarling_set_mask_config` and `cedarling_register_entity_map` are executable by PUBLIC.**
`src/mask/mod.rs:34`, `src/resource/schema_map.rs:195`. Only the four policy-lifecycle functions are `REVOKE`d (`sql/cedarling_pg--0.1.0.sql:351-354`). `entity_map` controls which Cedar entity type/id a row is evaluated as (`schema_map.rs:39-67`) — a security-relevant input to the decision — yet is mutable by anyone. An attacker can remove a masking rule or remap an entity to change authorization outcomes.
**Fix:** `REVOKE EXECUTE ... FROM PUBLIC` on both functions and `REVOKE INSERT/UPDATE/DELETE ON cedarling.mask_rules, cedarling.entity_map FROM PUBLIC`. Grant writes only to the extension owner / admin role.

### 🟡 MEDIUM

- **M1 — `where_partial_fallback = permit` lowers an un-lowerable predicate to `TRUE`.** Safe only when paired with a row-level `cedarling_authorized*` re-check, which nothing enforces. Default is `deny` (good); mitigated by fixing H1 (making it `Suset`). `src/authz/where_clause.rs:90-93,494-499,645-663`.
- **M2 — Cache not invalidated on `entity_map`/`mask_rules` changes.** Only `use_policy`/`rollback` call `clear_all` (`versions.rs:99,170`). Tightening an entity mapping has no effect until TTL expiry (default 300s) — a stale-allow window. Cache key also omits `current_user`/`search_path` (defense-in-depth note for pooled deployments). `src/authz/cache.rs:48-102`.
- **M3 — `lower_scalar_literal` emits `inf`/`-inf`/`nan`/`infinity` verbatim into SQL.** `src/authz/where_clause.rs:277` (`t.parse::<f64>().is_ok()`). Source is admin-authored policy text (not attacker-controlled at query time) and the constrained alphabet prevents arbitrary SQL injection, but it silently produces an incorrect/erroring predicate, undermining pushdown correctness. **Fix:** reject non-finite floats (`!value.is_finite()`) or quote+cast.

### 🟢 LOW
- **L1/L2** — `cedarling_use_policy`, `cedarling_register_policy_version`, `cedarling_diff_policies` read arbitrary server-side file paths (`versions.rs:66,27,204`). Correctly `REVOKE`d from PUBLIC, so owner-only — appropriate. Paths are unvalidated (no allowlist/canonicalization); an over-broad grant later would become a file-read/SSRF primitive (bootstrap can reference remote URIs). Consider restricting to a configured directory.
- **L3** — Disabling JWT signature/status validation is only a `WARN` log (`cedarling/src/jwt/mod.rs`). Defaults are strict; the test bootstrap disables it (test-only). Ensure no shipped sample/`docker-initdb` bootstrap disables validation.
- **L4** — Cedar diagnostics and resource JSON may surface in the observability trace ring. Token bundle code correctly avoids echoing raw JWTs (`tokens/bundle.rs:116-129`). Ensure trace-reading functions are not granted to PUBLIC if traces can contain row data.

### ✅ Confirmed safe (reviewed)
- **WHERE-pushdown SQL injection:** identifiers constrained to `[A-Za-z0-9_]` (`parse_resource_field:256-264`) and quoted via `quote_ident_safe`; string literals Cedar-unescaped then re-quoted via `quote_literal_safe`. String-state tracking in `split_top_level`/`extract_clauses` is correct. No raw policy text concatenated into SQL.
- **All SPI calls use parameter binding** (`$1`, `.into()`) across `mask/mod.rs`, `schema_map.rs`, `versions.rs`, `tokens/sql.rs`, `where_clause.rs:544`.
- **NULL action errors (does not default to "Read"):** `authorized_row.rs:44-46`, `:302-304` — `None` → `finalize_error(RequestInvalid)` → fail-closed. The recent change is correct.
- **Cache key completeness:** policy segment + tokens/principal + resource + action + context, each length-prefixed (`cache.rs:69-85`). Only `Ok` decisions cached; errors never cached; mutex poison → fail-closed.
- **Masking fail-closed:** unknown `mask_type` → redact; `Hash` without salt → `[HASH_SALT_REQUIRED]` sentinel; `condition_sql` is rejected, never executed (`mask/mod.rs:206-217`). Salt GUC is `Suset`.
- **`sanitize_table_filename` does not exist** in this codebase — no path-traversal sink found. File IO is reachable only through `Suset`/REVOKE-gated functions.

---

## 3. `#[pg_test]` Tests

pgrx idioms are correct: proper `Spi::run` / `Spi::get_one` / `Spi::get_one_with_args`, meaningful assertions, Drop-guards for cleanup, per-pid temp dirs, engine+cache reset between cases. No `assert!(true)` placeholders, no `#[ignore]`. The pure-Rust where-lowering unit suite is genuinely thorough.

**However, coverage is skewed toward the unsigned path and fail-closed / no-engine artifacts.**

### 🔴 CRITICAL gaps
1. **The signed / JWT authorization path is never exercised against a real engine.** `cedarling_authorized` (multi-issuer, `authorized.rs:191`) is only ever called with no engine, in shadow, or with empty input. `get_matching_policies_multi_issuer` and the token-bundle bridge never run in a `#[pg_test]` with a loaded engine. Signed allow **and** deny are untested end-to-end.
2. **A real-engine grant (`allow=true`) is asserted in exactly one place** — the unsigned RLS test (`pg_test_rls_unsigned.rs`). That is the only genuine allow-and-deny pair. `cedarling_authorized_row` is never tested returning a real engine `true`.

### 🟠 IMPORTANT gaps
- `cedarling_diff_policies` has no `#[pg_test]` — only the pure text helpers are unit-tested (`versions.rs:380-402`). SQL-facing function, `diff_mode` dispatch, and file-path error branch untested.
- GUC check-hook rejection untested — no test does `SET cedarling.context = '[]'` (or invalid tokens JSON) and asserts the `SET` errors (`guc_config.rs:261-276`).
- `cedarling_set_tokens` / `cedarling_current_tokens` / `cedarling_clear_tokens` have zero `#[pg_test]` — the jsonb→GUC→jsonb roundtrip is never tested.
- `cedarling_mask_plan` (`mask/mod.rs:82`) untested.
- Rollback with no previous policy (`cedarling_rollback_policy` returning `false`, `versions.rs:162`) untested; only the happy path is.

### 🟡 MINOR gaps
- NULL-action `None` arm of `cedarling_authorized_row` not directly tested (tests pass `Some(...)` or empty string).
- `sanitize_id_columns` (`schema_map.rs:139`) has no malicious-input test (quotes/dots/`--` in id column names).
- No path-traversal test for the file-path functions (arbitrary-path read is reachable from SQL but untrusted-path handling is untested).
- `cedarling_status` / observability counters never asserted from SQL.

### Verdict
Test-code **quality is good**; **authorization-decision coverage is incomplete**. The suite proves the plumbing and the conservative SQL lowering well, but under-verifies the actual authorization decisions the extension exists to make — most importantly the signed/JWT path, which has no real-engine integration test at all.

---

## Recommended fix order
1. **H1** — make `mode`/`fail_mode`/`strategy`/`where_partial_fallback` `Suset`. (highest impact)
2. **H2** — `REVOKE` mask/entity-map functions and tables from PUBLIC.
3. **Test CRITICAL gap** — add a real-engine signed allow/deny integration `#[pg_test]`.
4. **Docs CRITICAL** — fix the `cosign verify-blob` command (quick).
5. Docs IMPORTANT (`regclass`→`oid`, 2-arg example) and remaining test/security mediums.

jans-cedarling/cedarling_pg/src/policy/versions.rs
```diff
@@ -0,0 +1,403 @@
...
+    authz_cache::global_cache().clear_all();
+    if let Err(e) = set_policy_version_guc(&resolved.version_guc_value) {
+        extension_log::log_diagnostic(
+            guc_config::CedarlingLogLevelGuc::Error,
+            &format!("cedarling_use_policy: failed to set cedarling.policy_version: {e}"),
+        );
+        if let Err(e2) = engine::rollback_policy() {
+            extension_log::log_diagnostic(
+                guc_config::CedarlingLogLevelGuc::Error,
+                &format!(
```
tareknaser, line 120: Follow-up to my earlier comment on `set_policy_version_guc` (which has been mostly addressed). 

I noticed one edge case the new rollback path doesn't catch: `engine::rollback_policy()` returns `Ok(None)` when there is no previous engine to revert to (this is what happens on the very first `cedarling_use_policy` call in a backend) 


The `Ok(None)` arm is silently ignored so I think it would work like:
1. `engine::use_policy` succeeds and installs the new policy.
2. `set_policy_version_guc` fails for some reason
3. `rollback_policy()` returns `Ok(None)` which means nothing to roll back to.
4. Function returns `false`.
5. The engine is still running the new policy but `cedarling.policy_version` and `cedarling.policy_history` are stale.

tareknaser (commented):
Thanks for addressing my previous review comments. I have one follow up

jans-cedarling/cedarling_pg/src/policy/versions.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13856/commits/b6546feab1ce7b17c9d29b193e16125e4c5cb4e7

haileyesus2433:
> Here is a review from Claude. Can you pls analyze each case?
> 
> # `cedarling_pg` Review — Docs, Security, Tests
> Reviewed the `cedarling_pg` changes on this branch vs `main`: documentation, the Rust/pgrx extension source, generated SQL, install script, and the `#[pg_test]` suite.
> 
> **Overall:** solid, security-conscious work — fail-closed defaults, parameterized SPI everywhere, careful pure-Rust quoting of identifiers/literals, and a candid security section in the docs. The findings below are the real gaps worth addressing before merge.
> 
> ## 1. Documentation
> The docs are high quality and largely accurate (GUC table, function inventory, PG matrix 13–18, asset/provenance names all verified against code). Defects:
> 
> ### 🔴 CRITICAL
> * **`docs/cedarling/integrations/postgres.md:131-135` — the `cosign verify-blob` command will not run.** It only passes `--certificate-oidc-issuer-regexp` and omits a certificate-identity matcher (`--certificate-identity` / `--certificate-identity-regexp`), which `cosign verify-blob` requires. The command errors out, so the recommended verification step fails. Compare the complete, correct form in `docs/contribute/ci-cd/slsa-verification.md:720-724`.
> 
> ### 🟠 IMPORTANT
> * **`postgres.md:294,296` — wrong parameter type `regclass` vs `oid`.** `cedarling_validate_schema` and `cedarling_register_entity_map` actually take `oid`, not `regclass` (`sql/cedarling_pg--0.1.0.sql:372`, `:253`). A bare table-name literal will not auto-cast to `oid`; callers must write `'mytable'::regclass::oid` (as the code comment at `src/policy/schema.rs:87` states). As documented, calls fail.
> * **`postgres.md:53` — non-callable example.** The flow narrative shows `cedarling_authorized_row(students, 'Read')` (2 args), but every overload requires 3 args `(record, action, context)` with no SQL `DEFAULT` (`sql/...sql:117-121`). This contradicts the doc's own reference row at `postgres.md:266`.
> 
> ### 🟡 MINOR
> * `cedarling-pg` artifact family is missing from the SLSA "Artifact Families" table (`slsa-verification.md:781-791`), even though `cedarling-pg.intoto.jsonl` is published (`build-packages.yml:1880`) and `postgres.md:529` tells users it exists.
> * `cedarling_mask_row` parameter documented as `row` (`postgres.md:305`); actual name is `row_json` (`sql/...sql:230`). Breaks named-argument calls.
> * The `instrumentation` value of `cedarling.mode` (`postgres.md:332`) is listed but never explained; code defines it as "evaluate and log, still return the decision" (`src/guc_config.rs:24-27`).
> 
> ## 2. Security
> ### 🔴 CRITICAL
> None.
> 
> ### 🟠 HIGH
> **H1 — Enforcement-affecting GUCs are `Userset`, so any role can weaken enforcement for its own session.** `src/guc_config.rs:300-352`. `cedarling.mode`, `cedarling.fail_mode`, `cedarling.strategy`, and `cedarling.where_partial_fallback` are all registered `GucContext::Userset`. Any role with query access can run:
> 
> ```sql
> SET cedarling.mode = 'shadow';     -- finalize_decision() returns true unconditionally (authorized.rs:357)
> SET cedarling.fail_mode = 'open';  -- finalize_error() returns true on any error (authorized.rs:366)
> ```
> 
> This neutralizes authorization for that session — directly impacting any RLS policy or `SECURITY DEFINER` view relying on `cedarling_authorized*`, since RLS predicates evaluate in the querying user's session. **Fix:** register these as `GucContext::Suset` (matching `bootstrap_config`, `policy_version`, `mask_hash_salt` at `:449,459,467`).
> 
> **H2 — `cedarling_set_mask_config` and `cedarling_register_entity_map` are executable by PUBLIC.** `src/mask/mod.rs:34`, `src/resource/schema_map.rs:195`. Only the four policy-lifecycle functions are `REVOKE`d (`sql/cedarling_pg--0.1.0.sql:351-354`). `entity_map` controls which Cedar entity type/id a row is evaluated as (`schema_map.rs:39-67`) — a security-relevant input to the decision — yet is mutable by anyone. An attacker can remove a masking rule or remap an entity to change authorization outcomes. **Fix:** `REVOKE EXECUTE ... FROM PUBLIC` on both functions and `REVOKE INSERT/UPDATE/DELETE ON cedarling.mask_rules, cedarling.entity_map FROM PUBLIC`. Grant writes only to the extension owner / admin role.
> 
> ### 🟡 MEDIUM
> * **M1 — `where_partial_fallback = permit` lowers an un-lowerable predicate to `TRUE`.** Safe only when paired with a row-level `cedarling_authorized*` re-check, which nothing enforces. Default is `deny` (good); mitigated by fixing H1 (making it `Suset`). `src/authz/where_clause.rs:90-93,494-499,645-663`.
> * **M2 — Cache not invalidated on `entity_map`/`mask_rules` changes.** Only `use_policy`/`rollback` call `clear_all` (`versions.rs:99,170`). Tightening an entity mapping has no effect until TTL expiry (default 300s) — a stale-allow window. Cache key also omits `current_user`/`search_path` (defense-in-depth note for pooled deployments). `src/authz/cache.rs:48-102`.
> * **M3 — `lower_scalar_literal` emits `inf`/`-inf`/`nan`/`infinity` verbatim into SQL.** `src/authz/where_clause.rs:277` (`t.parse::<f64>().is_ok()`). Source is admin-authored policy text (not attacker-controlled at query time) and the constrained alphabet prevents arbitrary SQL injection, but it silently produces an incorrect/erroring predicate, undermining pushdown correctness. **Fix:** reject non-finite floats (`!value.is_finite()`) or quote+cast.
> 
> ### 🟢 LOW
> * **L1/L2** — `cedarling_use_policy`, `cedarling_register_policy_version`, `cedarling_diff_policies` read arbitrary server-side file paths (`versions.rs:66,27,204`). Correctly `REVOKE`d from PUBLIC, so owner-only — appropriate. Paths are unvalidated (no allowlist/canonicalization); an over-broad grant later would become a file-read/SSRF primitive (bootstrap can reference remote URIs). Consider restricting to a configured directory.
> * **L3** — Disabling JWT signature/status validation is only a `WARN` log (`cedarling/src/jwt/mod.rs`). Defaults are strict; the test bootstrap disables it (test-only). Ensure no shipped sample/`docker-initdb` bootstrap disables validation.
> * **L4** — Cedar diagnostics and resource JSON may surface in the observability trace ring. Token bundle code correctly avoids echoing raw JWTs (`tokens/bundle.rs:116-129`). Ensure trace-reading functions are not granted to PUBLIC if traces can contain row data.
> 
> ### ✅ Confirmed safe (reviewed)
> * **WHERE-pushdown SQL injection:** identifiers constrained to `[A-Za-z0-9_]` (`parse_resource_field:256-264`) and quoted via `quote_ident_safe`; string literals Cedar-unescaped then re-quoted via `quote_literal_safe`. String-state tracking in `split_top_level`/`extract_clauses` is correct. No raw policy text concatenated into SQL.
> * **All SPI calls use parameter binding** (`$1`, `.into()`) across `mask/mod.rs`, `schema_map.rs`, `versions.rs`, `tokens/sql.rs`, `where_clause.rs:544`.
> * **NULL action errors (does not default to "Read"):** `authorized_row.rs:44-46`, `:302-304` — `None` → `finalize_error(RequestInvalid)` → fail-closed. The recent change is correct.
> * **Cache key completeness:** policy segment + tokens/principal + resource + action + context, each length-prefixed (`cache.rs:69-85`). Only `Ok` decisions cached; errors never cached; mutex poison → fail-closed.
> * **Masking fail-closed:** unknown `mask_type` → redact; `Hash` without salt → `[HASH_SALT_REQUIRED]` sentinel; `condition_sql` is rejected, never executed (`mask/mod.rs:206-217`). Salt GUC is `Suset`.
> * **`sanitize_table_filename` does not exist** in this codebase — no path-traversal sink found. File IO is reachable only through `Suset`/REVOKE-gated functions.
> 
> ## 3. `#[pg_test]` Tests
> pgrx idioms are correct: proper `Spi::run` / `Spi::get_one` / `Spi::get_one_with_args`, meaningful assertions, Drop-guards for cleanup, per-pid temp dirs, engine+cache reset between cases. No `assert!(true)` placeholders, no `#[ignore]`. The pure-Rust where-lowering unit suite is genuinely thorough.
> 
> **However, coverage is skewed toward the unsigned path and fail-closed / no-engine artifacts.**
> 
> ### 🔴 CRITICAL gaps
> 1. **The signed / JWT authorization path is never exercised against a real engine.** `cedarling_authorized` (multi-issuer, `authorized.rs:191`) is only ever called with no engine, in shadow, or with empty input. `get_matching_policies_multi_issuer` and the token-bundle bridge never run in a `#[pg_test]` with a loaded engine. Signed allow **and** deny are untested end-to-end.
> 2. **A real-engine grant (`allow=true`) is asserted in exactly one place** — the unsigned RLS test (`pg_test_rls_unsigned.rs`). That is the only genuine allow-and-deny pair. `cedarling_authorized_row` is never tested returning a real engine `true`.
> 
> ### 🟠 IMPORTANT gaps
> * `cedarling_diff_policies` has no `#[pg_test]` — only the pure text helpers are unit-tested (`versions.rs:380-402`). SQL-facing function, `diff_mode` dispatch, and file-path error branch untested.
> * GUC check-hook rejection untested — no test does `SET cedarling.context = '[]'` (or invalid tokens JSON) and asserts the `SET` errors (`guc_config.rs:261-276`).
> * `cedarling_set_tokens` / `cedarling_current_tokens` / `cedarling_clear_tokens` have zero `#[pg_test]` — the jsonb→GUC→jsonb roundtrip is never tested.
> * `cedarling_mask_plan` (`mask/mod.rs:82`) untested.
> * Rollback with no previous policy (`cedarling_rollback_policy` returning `false`, `versions.rs:162`) untested; only the happy path is.
> 
> ### 🟡 MINOR gaps
> * NULL-action `None` arm of `cedarling_authorized_row` not directly tested (tests pass `Some(...)` or empty string).
> * `sanitize_id_columns` (`schema_map.rs:139`) has no malicious-input test (quotes/dots/`--` in id column names).
> * No path-traversal test for the file-path functions (arbitrary-path read is reachable from SQL but untrusted-path handling is untested).
> * `cedarling_status` / observability counters never asserted from SQL.
> 
> ### Verdict
> Test-code **quality is good**; **authorization-decision coverage is incomplete**. The suite proves the plumbing and the conservative SQL lowering well, but under-verifies the actual authorization decisions the extension exists to make — most importantly the signed/JWT path, which has no real-engine integration test at all.
> 
> ## Recommended fix order
> 1. **H1** — make `mode`/`fail_mode`/`strategy`/`where_partial_fallback` `Suset`. (highest impact)
> 2. **H2** — `REVOKE` mask/entity-map functions and tables from PUBLIC.
> 3. **Test CRITICAL gap** — add a real-engine signed allow/deny integration `#[pg_test]`.
> 4. **Docs CRITICAL** — fix the `cosign verify-blob` command (quick).
> 5. Docs IMPORTANT (`regclass`→`oid`, 2-arg example) and remaining test/security mediums.

addressed applicable comments from https://github.com/JanssenProject/jans/pull/13856/commits/e2e86cc25f1e350c10ce4642c2b92bca66abca13 - https://github.com/JanssenProject/jans/pull/13856/commits/0ab17e5254057eba51b29b3e5293e7bac930beb6

olehbozhok:
## Re-check of remaining review items after fix batch (e2e86cc – 0ab17e5)

Re-verified all 13 items from the review. Summary:

### ✅ Fully addressed (9/13)

- **H1** — `mode`/`fail_mode`/`strategy`/`where_partial_fallback` all `Suset` ✅
- **H2** — `cedarling_set_mask_config` + `cedarling_register_entity_map` REVOKEd from PUBLIC; mask_rules/entity_map tables REVOKEd INSERT/UPDATE/DELETE ✅
- **Test critical** — `test_signed_authorized_allow_then_deny` exercises JWT path (allow + deny) against a real engine ✅
- **All 6 doc defects** corrected (cosign command, regclass→oid, 2-arg example, SLSA table, param name `row_json`, instrumentation docs) ✅

### ⬜ Remaining — 3 not addressed, 1 partial

### 1. JWT tokens via GUC — end-to-end untested (was: "token functions pg_test")

This is more important than "no pg_test for 3 thin wrappers."

The production JWT RLS path is:
SELECT cedarling_set_tokens('...'::jsonb);
CREATE POLICY rls ON t FOR SELECT
    USING (cedarling_authorized_row_jwt(t, 'Read'));
SELECT * FROM t;  -- RLS reads tokens from cedarling.tokens GUC

`cedarling_authorized_row_jwt()` (`authorized_row.rs:301-310`) passes `None` for `token_bundle`, so `resolve_token_bundle()` (`authorized.rs:345-351`) **must** read from the GUC. There is no alternative parameter.

The existing signed test (`test_signed_authorized_allow_then_deny`) passes tokens as an explicit argument to `cedarling_authorized($1, $2, $3)`, which bypasses the GUC entirely. The GUC→engine bridge is only tested piecewise:
- GUC check hook: unit-tested (validate.rs)
- Token parsing: unit-tested (bundle.rs)
- Engine authorize: unit-tested (bridge.rs)

But the end-to-end `cedarling_set_tokens → cedarling_authorized_row_jwt → resolve_token_bundle → cedarling_authorized` chain has **no pg_test**.

**Severity: Medium.** The components are individually correct, but the integration path is untested. Would be good to add before the first release.

### 2. `cedarling_diff_policies` — no pg_test

Owner-only utility (REVOKEd from PUBLIC). Rust diff logic is unit-tested (`versions.rs:380-402`). Missing: SQL integration test (file I/O error branch, diff_mode dispatch).

**Severity: Low.** The function is not in the auth path.

### 3. `cedarling_mask_plan` — no pg_test

Simple read-only SELECT from `cedarling.mask_rules`. Owner-only. The underlying table operations are exercised by other mask tests.

**Severity: Low.**

### 4. GUC check-hook rejection — no direct `SET ... = 'invalid'` pg_test

Validation logic (`context_json_is_valid_object`, `tokens_json_is_valid`) is unit-tested. Function-level rejection is integration-tested (`test_input_validation_fail_closed` asserts deny for `context = '[]'`). What's missing: a pg_test that runs `SET cedarling.context = '[]'` and asserts the SET errors.

**Severity: Low.** The check hook is a thin C-unwind bridge; the logic behind it is already verified.

haileyesus2433:
> ## Re-check of remaining review items after fix batch ([e2e86cc](https://github.com/JanssenProject/jans/commit/e2e86cc25f1e350c10ce4642c2b92bca66abca13) – [0ab17e5](https://github.com/JanssenProject/jans/commit/0ab17e5254057eba51b29b3e5293e7bac930beb6))
> Re-verified all 13 items from the review. Summary:
> 
> ### ✅ Fully addressed (9/13)
> * **H1** — `mode`/`fail_mode`/`strategy`/`where_partial_fallback` all `Suset` ✅
> * **H2** — `cedarling_set_mask_config` + `cedarling_register_entity_map` REVOKEd from PUBLIC; mask_rules/entity_map tables REVOKEd INSERT/UPDATE/DELETE ✅
> * **Test critical** — `test_signed_authorized_allow_then_deny` exercises JWT path (allow + deny) against a real engine ✅
> * **All 6 doc defects** corrected (cosign command, regclass→oid, 2-arg example, SLSA table, param name `row_json`, instrumentation docs) ✅
> 
> ### ⬜ Remaining — 3 not addressed, 1 partial
> ### 1. JWT tokens via GUC — end-to-end untested (was: "token functions pg_test")
> This is more important than "no pg_test for 3 thin wrappers."
> 
> The production JWT RLS path is: SELECT cedarling_set_tokens('...'::jsonb); CREATE POLICY rls ON t FOR SELECT USING (cedarling_authorized_row_jwt(t, 'Read')); SELECT * FROM t; -- RLS reads tokens from cedarling.tokens GUC
> 
> `cedarling_authorized_row_jwt()` (`authorized_row.rs:301-310`) passes `None` for `token_bundle`, so `resolve_token_bundle()` (`authorized.rs:345-351`) **must** read from the GUC. There is no alternative parameter.
> 
> The existing signed test (`test_signed_authorized_allow_then_deny`) passes tokens as an explicit argument to `cedarling_authorized($1, $2, $3)`, which bypasses the GUC entirely. The GUC→engine bridge is only tested piecewise:
> 
> * GUC check hook: unit-tested (validate.rs)
> * Token parsing: unit-tested (bundle.rs)
> * Engine authorize: unit-tested (bridge.rs)
> 
> But the end-to-end `cedarling_set_tokens → cedarling_authorized_row_jwt → resolve_token_bundle → cedarling_authorized` chain has **no pg_test**.
> 
> **Severity: Medium.** The components are individually correct, but the integration path is untested. Would be good to add before the first release.
> 
> ### 2. `cedarling_diff_policies` — no pg_test
> Owner-only utility (REVOKEd from PUBLIC). Rust diff logic is unit-tested (`versions.rs:380-402`). Missing: SQL integration test (file I/O error branch, diff_mode dispatch).
> 
> **Severity: Low.** The function is not in the auth path.
> 
> ### 3. `cedarling_mask_plan` — no pg_test
> Simple read-only SELECT from `cedarling.mask_rules`. Owner-only. The underlying table operations are exercised by other mask tests.
> 
> **Severity: Low.**
> 
> ### 4. GUC check-hook rejection — no direct `SET ... = 'invalid'` pg_test
> Validation logic (`context_json_is_valid_object`, `tokens_json_is_valid`) is unit-tested. Function-level rejection is integration-tested (`test_input_validation_fail_closed` asserts deny for `context = '[]'`). What's missing: a pg_test that runs `SET cedarling.context = '[]'` and asserts the SET errors.
> 
> **Severity: Low.** The check hook is a thin C-unwind bridge; the logic behind it is already verified.

resolved from https://github.com/JanssenProject/jans/pull/13856/commits/6b3a339119d42fe56ee1900bcc08b627d6009073 - https://github.com/JanssenProject/jans/pull/13856/commits/e180532a0a62da26037d4ec0a1bc1ae12206223e

dagregi (dismissed):

moabu (dismissed):

olehbozhok (dismissed):
LOW / defense-in-depth: trace & observability functions are PUBLIC-executable over a per-backend ring buffer

  cedarling_last_trace(), cedarling_recent_traces(int), cedarling_explain(text, text), and cedarling_status() are not
  REVOKEd from PUBLIC (sql/cedarling_pg--0.1.0.sql:196,208,241,311), unlike the policy-lifecycle and entity/mask
  writers.

  The trace ring is per-backend (observability/trace.rs), so it persists across sessions that share the same backend
  process — i.e. in any connection-pooled deployment (PgBouncer in session/transaction pooling, app-side pools). A
  low-privilege role can therefore call cedarling_recent_traces() and read trace entries produced by other sessions that
  previously ran on that backend.

  What's exposed per entry: resource_type, resource_id (the row's primary-key value), action, principal_id, and
  diag_errors. This is authorization metadata, not row contents — so impact is limited — but resource_id can be a
  sensitive identifier, and this crosses a session boundary the caller shouldn't see across.

  Note this is metadata-only and gated behind connection pooling — hence LOW, not a merge blocker. Raising it for an
  explicit decision.

  Suggested fix — pick one:

  1. Restrict the readers (preferred if traces aren't needed by app roles):
  REVOKE EXECUTE ON FUNCTION cedarling_last_trace()        FROM PUBLIC;
  REVOKE EXECUTE ON FUNCTION cedarling_recent_traces(int)  FROM PUBLIC;
  REVOKE EXECUTE ON FUNCTION cedarling_explain(text, text) FROM PUBLIC;
  REVOKE EXECUTE ON FUNCTION cedarling_status()            FROM PUBLIC;
  -- GRANT EXECUTE ... TO <observability_role>;  -- as needed
  2. Or document the pooling caveat in docs/cedarling/integrations/postgres.md (security section) — state that trace
  readers may surface cross-session metadata under shared backends and should be granted only to trusted observability
  roles.

  cedarling_explain is borderline: it re-runs authorization on caller-supplied input and returns the caller's own
  diagnostics, so it's less of a cross-session leak — but it still reads the shared ring indirectly via push_trace, so
  include it in whichever option you choose.

  Everything else in the security pass came back clean (H1/H2 confirmed fixed, SQL fully parameterized, pushdown
  fail-closed). This is the only residual.

jans-cedarling/cedarling_pg/src/authz/where_clause.rs
```diff
@@ -0,0 +1,1060 @@
...
+///
+/// We re-parse via `cedar_policy::Policy::parse` first so a malformed source
+/// (which `PolicyMetadata` callers should not produce, but we cannot prove)
+/// is treated as unhandled rather than silently mis-lowered.
+fn lower_policy_to_sql(meta: &PolicyMetadata) -> Option<String> {
+    if cedar_policy::Policy::parse(None, &meta.source).is_err() {
+        return None;
+    }
+
+    let when_clauses = extract_clauses(&meta.source, "when")?;
```
tareknaser, line 482: I think there's a correctness issue here. `lower_policy_to_sql` only looks at `when`/`unless` clauses and ignores constraints in the policy head.

For example, `permit(principal, action, resource == Doc::"42")` lowers to `TRUE`, even though the policy only applies to a single resource. Similarly, `permit(principal, action, resource == Doc::"42") when { resource.public == true }` lowers to `"public" = TRUE`, dropping the resource constraint entirely.

From what I can tell, `matching_policies_for_table` filters policies by entity type rather than specific scope constraints so scoped policies can reach this code path and have their principal/resource restrictions discarded during SQL lowering.

That means `cedarling_where` can produce predicates that match more rows than the underlying Cedar policy allows.

jans-cedarling/cedarling_pg/src/policy/versions.rs
```diff
@@ -0,0 +1,415 @@
...
+                &format!("cedarling_use_policy: registry lookup failed: {e}"),
+            );
+            ResolvedPolicy {
+                bootstrap_path: name_or_path.to_string(),
+                version_guc_value: name_or_path.to_string(),
+            }
+        },
+    };
+
+    let previous = match engine::use_policy(&resolved.bootstrap_path) {
```
tareknaser, line 90: The GUC failure handling looks good now. One thing that still seems worth documenting is that the policy swap is per-backend and not transactional.

`engine::use_policy` updates the Rust `static ENGINE` immediately so a `ROLLBACK` will undo the catalog changes and `set_config` but the backend will keep using the new policy. Also, because the engine is process-local, `cedarling_use_policy` only affects the current backend. other connections keep their own engine instance.

The current comment in `catalog.rs` ("global policy swap") and the wording in `postgres.md` read as if this is cluster-wide. It would be helpful to clarify that the swap is per-backend and non-transactional and that cluster-wide changes should be done by updating `cedarling.bootstrap_config` and reconnecting rather than relying on `cedarling_use_policy` to propagate.

tareknaser (commented):
Thank you! I only have one main comment left then I think we're ready to merge

jans-cedarling/cedarling_pg/src/policy/versions.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/13856/commits/20a8dd84a6bd22392d0f1ca4281d34ebd577f652

haileyesus2433:
> LOW / defense-in-depth: trace & observability functions are PUBLIC-executable over a per-backend ring buffer
> 
> cedarling_last_trace(), cedarling_recent_traces(int), cedarling_explain(text, text), and cedarling_status() are not REVOKEd from PUBLIC (sql/cedarling_pg--0.1.0.sql:196,208,241,311), unlike the policy-lifecycle and entity/mask writers.
> 
> The trace ring is per-backend (observability/trace.rs), so it persists across sessions that share the same backend process — i.e. in any connection-pooled deployment (PgBouncer in session/transaction pooling, app-side pools). A low-privilege role can therefore call cedarling_recent_traces() and read trace entries produced by other sessions that previously ran on that backend.
> 
> What's exposed per entry: resource_type, resource_id (the row's primary-key value), action, principal_id, and diag_errors. This is authorization metadata, not row contents — so impact is limited — but resource_id can be a sensitive identifier, and this crosses a session boundary the caller shouldn't see across.
> 
> Note this is metadata-only and gated behind connection pooling — hence LOW, not a merge blocker. Raising it for an explicit decision.
> 
> Suggested fix — pick one:
> 
> 1. Restrict the readers (preferred if traces aren't needed by app roles):
>    REVOKE EXECUTE ON FUNCTION cedarling_last_trace()        FROM PUBLIC;
>    REVOKE EXECUTE ON FUNCTION cedarling_recent_traces(int)  FROM PUBLIC;
>    REVOKE EXECUTE ON FUNCTION cedarling_explain(text, text) FROM PUBLIC;
>    REVOKE EXECUTE ON FUNCTION cedarling_status()            FROM PUBLIC;
>    -- GRANT EXECUTE ... TO <observability_role>;  -- as needed
> 2. Or document the pooling caveat in docs/cedarling/integrations/postgres.md (security section) — state that trace
>    readers may surface cross-session metadata under shared backends and should be granted only to trusted observability
>    roles.
> 
> cedarling_explain is borderline: it re-runs authorization on caller-supplied input and returns the caller's own diagnostics, so it's less of a cross-session leak — but it still reads the shared ring indirectly via push_trace, so include it in whichever option you choose.
> 
> Everything else in the security pass came back clean (H1/H2 confirmed fixed, SQL fully parameterized, pushdown fail-closed). This is the only residual.

addressed the remaining comments in here https://github.com/JanssenProject/jans/pull/13856/commits/2f3e7da6b2696f2669a832edebb359136bb8e681

haileyesus2433:
@coderabbitai approve

dagregi (approved):

jans-cedarling/cedarling_pg/src/catalog.rs
```diff
@@ -19,6 +19,12 @@ use pgrx::extension_sql;
 #[allow(unused_imports)]
 use crate::mask::cedarling_set_mask_config;
 #[allow(unused_imports)]
+use crate::observability::status::cedarling_status;
+#[allow(unused_imports)]
```
olehbozhok, line 23: Why should we have `#[allow(unused_imports)]`?

olehbozhok (approved):

tareknaser (approved):
Thank you

jans-cedarling/cedarling_pg/src/catalog.rs
haileyesus2433: i documented it in the code but its because `extension_sql!`'s `requires = [...]` clause is resolved by `pgrx`'s SQL entity graph by bare entity name, not as a Rust path  so the use statements aren't decorative, they bring the bare identifiers into scope so the macro input parses. I tried the cleaner-looking refactor first: drop the uses and  inline fully-qualified paths in 
`requires = [crate::observability::status::cedarling_status, ...]`. 
cargo check accepted it cleanly, but cargo pgrx test failed at SQL generation with
 `Could not find requires target of cedarling_pg_policy_function_privileges:crate::observability::status::cedarling_status`, 
 which cascaded every pg_test into the "could not obtain test mutex" failure because the extension never installed.

moabu (approved):

PR #14149, opened 2026-05-25
Update ExternalResourceOwnerPasswordCredentialsService.java

### LINKED ISSUE #14148: jans-auth-server: Enhance ROPC script to throw custom error and error_description
We are running custom ROPC scripts and at the moment, there is no way of telling the end user, why the script failed and tell them a bit more about that. 

For example, we want to create custom error and error_description, conforming to OIDC specs and use following sample in our ROPC script: 

``` python
exception = context.createWebApplicationException(503,'{"error": "access_denied", "error_description": "User is blacklisted"}')
context.setWebApplicationException(exception)
```
This should send the below response to the end user: 
{"error": "access_denied", "error_description": "User is blacklisted"}
with HTTP status code as 503. 

For this, we would have to enhance the ExternalResourceOwnerPasswordCredentialsService.java file to have some better Exception handling. At the moment, it does not do anything about the exception, but just saves it in the Script configurations. 

https://github.com/JanssenProject/jans/blob/main/jans-auth-server/server/src/main/java/io/jans/as/server/service/external/ExternalResourceOwnerPasswordCredentialsService.java#L71

``` java
        } catch (Exception ex) {
            log.error(ex.getMessage(), ex);
            saveScriptError(customScriptConfiguration.getCustomScript(), ex);
            return false;
        }
```

We would like to change it to this: 

``` java
       } catch (WebApplicationException e) {
            if (log.isTraceEnabled()) {
                log.trace("WebApplicationException from script", e);
            }
            throw e;
        } catch (Exception ex) {
            log.error(ex.getMessage(), ex);
            saveScriptError(context.getScript().getCustomScript(), ex);
            throw new WebApplicationException(errorResponseFactory
                    .newErrorResponse(Response.Status.INTERNAL_SERVER_ERROR)
                    .entity(errorResponseFactory.getErrorAsJson(TokenErrorResponseType.ACCESS_DENIED, "", "Unable to run 'executeExternalAuthenticate' method in ROPC script."))
                    .build());
        }

```

In this way, if we are throwing custom WebApplicationException with proper `error` and `error_description` then, it will be thrown as it is. Else, if there is any other type of exception, then it shall be converted to WebApplicationException and the error user shall be shown: 

`{"error":"access_denied","error_description":"The resource owner or authorization server denied the request.","reason":"Unable to run 'executeExternalAuthenticate' method in ROPC script."}`

ayushjain0702:
Added custom script WebApplicationException handling.

### Prepare

-------------------

### Description
Added Custom WebApplicationException handling for ROPC scripts. The current implementation does not allow to customize the error and error_description to be sent to the end user. 

#### Target issue
https://github.com/JanssenProject/jans/issues/14148
  
closes #14148 

#### Implementation Details
Added Custom WebApplicationException handling for ROPC scripts. The current implementation does not allow to customize the error and error_description to be sent to the end user. 

-------------------
- [] Relevant unit and integration tests have been added/updated
- [] Relevant documentation has been updated if any (i.e. user guides, installation and configuration guides, technical design docs etc)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of REST-style errors from external authentication scripts so clients now receive proper error responses when scripts raise such errors.
  * Such REST-style errors are now rethrown (with optional trace logging) rather than being suppressed; other script failures continue to be logged, recorded, and handled gracefully.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/7c85ec19-7937-443c-a7a6-2a205989662b?refs=99f45ceeb9c5d6c6ca04b72cfe961bbbad038bf9&source=prChecksComment) |

jans-auth-server/server/src/main/java/io/jans/as/server/service/external/ExternalResourceOwnerPasswordCredentialsService.java
```diff
@@ -68,10 +78,18 @@ private boolean executeExternalAuthenticate(CustomScriptConfiguration customScri
...
+            if (log.isTraceEnabled()) {
+                log.trace("WebApplicationException from script", e);
+            }
+            throw e;
         } catch (Exception ex) {
             log.error(ex.getMessage(), ex);
-            saveScriptError(customScriptConfiguration.getCustomScript(), ex);
-            return false;
+            saveScriptError(context.getScript().getCustomScript(), ex);
+            throw new WebApplicationException(errorResponseFactory
```
yuriyz: We already have WebApplicationException re-thrown above. Why do we break normal execution for `Exception` ? What case are we trying to cover in line 89 ?

yuriyz (changes requested):

jans-auth-server/server/src/main/java/io/jans/as/server/service/external/ExternalResourceOwnerPasswordCredentialsService.java
ayushjain0702: Hi @yuriyz , 
The first catch is to catch the exception sent by the Custom ROPC script. LIke below in the Jython code:

``` python
exception = context.createWebApplicationException(503,'{"error": "access_denied", "error_description": "User is blacklisted"}')
context.setWebApplicationException(exception)
```

The second one is to create the exception generally, in case the Script throws any other kind of exception. Like NPE, or Http error while trying to connect to some external endpoint etc. 

We are trying to make the script similar to `ExternalAuthorizationChallengeService.java` file. Here, in the code, we are using context.getAuthzRequest() function. But, in case of ROPC, since, there is no Authorization request, we get NPE in that case. 

Hope this helps to understand what we are trying to achieve here.

yuriyz: @ayushjain0702 We can't apply same behavior as in `ExternalAuthorizationChallengeService.java` because for authorization challenge we have dedicated script for this endpoint only. ROPC script is used by Authorization Endpoint and by Token Endpoint. Throwing exception can lead to unexpected behavior. For example it's perfectly valid to get ROPC fail and let authorization go on and pass execution to lets say `PostAuthn` or other script.

You can easily cover case which you've described above by wrapping code inside custom script in `try/catch` block and throw `WebApplicationException` explicitly. This will work for NPE, http request failure or other error. In this way response will be propagated outside.

ayushjainmeta: Hi @yuriyz , 

We have tried this. The exception thrown by the ROPC custom script is not handled properly. The exception is only saved in the Script configuration. We have tested this multiple times and only after that we created this pull request. 

Can you please suggest any other method of doing it then?

Thanks.

ayushjain0702: Hi @yuriyz ,

We have tried this. The exception thrown by the ROPC custom script is not handled properly. The exception is only saved in the Script configuration. We have tested this multiple times and only after that we created this pull request.

Can you please suggest any other method of doing it then?

Thanks.

yuriyz: @ayushjain0702 I understand, we have to catch and re-throw `WebApplicationException`. Like this:
```
} catch (WebApplicationException e) {
            if (log.isTraceEnabled()) {
                log.trace("WebApplicationException from script", e);
            }
            throw e;
        } catch (Exception ex) {
            log.error(ex.getMessage(), ex);
            saveScriptError(customScriptConfiguration.getCustomScript(), ex);
            return false;
```
In this way whatever is thrown inside script can be wrapped inside script with with try/catch and thrown as `WebApplicationException`. Note, for `Exception` we `return false`. Will it work for you? If yes, please correct PR so we can let it go.

yuriyz: Also correct coderabbitai comment, it blocks merging.

yuriyz (approved):

yuriyz:
@ayushjain0702 thanks for making the changes. Accepted and merged.

Issue #12953, opened 2026-01-06
feat(jans-cedarling): improve policy store URI detection using content type


dagregi:
**Is your feature request related to a problem? Please describe.**

The current policy store URI detection in `jans-cedarling/cedarling/src/bootstrap_config/decode.rs` relies on checking if the URI ends with .cjar to determine whether it's an archive or if it's from a lock server. This approach is fragile because URLs may not always have extensions, archives could be served through different endpoints, and the logic can't distinguish between policy store JSON files and archives that don't follow the .cjar naming convention.

**Describe the solution you'd like**

Instead of checking file extensions, the URI detection should:
1. Make a HEAD/GET request to determine the content type
2. Check if the URL points to a policy store file or an archive
3. Use http headers or content inspection to make the determination

**Describe alternatives you've considered**

- Maybe separating the URI in the config for lock servers and cjar archives
- Content sniffing by downloading a small portion of the file

**Additional context**

This optimization would improve the reliability of policy store loading from remote sources and reduce false positives/negatives in archive detection.

PR #14174, opened 2026-05-29
feat(jans-cedarling): updating policies from URL by ttl

### LINKED ISSUE #14127: feat(jans-cedarling): updating policies from URL by ttl
## Summary

Cedarling currently loads policy stores from URL-based sources only once at startup ([cedarling/src/init/policy_store.rs](jans-cedarling/cedarling/src/init/policy_store.rs)). Add periodic background refresh that respects HTTP cache headers and avoids re-downloading unchanged data.

## Scope

Applies to URL-based policy store sources:
- `PolicyStoreSource::CjarUrl` — remote `.cjar` ZIP archive
- `PolicyStoreSource::LockServer` — policy store URI fetched from Lock Master service

Other sources (`Json`, `Yaml`, `FileJson`, `FileYaml`, `CjarFile`, `Directory`, `ArchiveBytes`) are out of scope — local/embedded, no remote refresh applicable. When `CEDARLING_POLICY_STORE_REFRESH_INTERVAL > 0` is set with a non-URL source, the property is silently ignored.

No manual refresh API. Background async worker only.

## Goals

- Periodic refresh of remote policy store in a background task
- Respect HTTP cache headers when present: `Cache-Control` (`max-age`, `no-cache`, `no-store`), `Expires`, `ETag`, `Last-Modified`
- Use `HEAD` requests to detect changes without downloading the full body
- Fallback to conditional `GET` if server does not support `HEAD` (405 / 501), logged as warning once per endpoint
- Fallback to plain `GET` with body hash comparison if server returns no cache validators (no `ETag`, no `Last-Modified`), logged as warning once per endpoint; refresh cadence then driven solely by `CEDARLING_POLICY_STORE_REFRESH_INTERVAL`
- Conditional `GET` via `If-None-Match` / `If-Modified-Since` → handle `304 Not Modified`
- Tolerant to all server behaviors — never panic, never serve empty policy set

## New Bootstrap Properties

- `CEDARLING_POLICY_STORE_REFRESH_INTERVAL` — base refresh interval, seconds. `0` disables refresh (default = `0`, preserves current behavior)
- `CEDARLING_POLICY_STORE_REFRESH_INTERVAL_MAX` — upper clamp for server-provided `max-age` / `Expires`
- (optional) `CEDARLING_POLICY_STORE_REFRESH_METHOD` — `auto` (default) | `head` | `get`

## Behavior

1. Initial fetch stores validators when present: `ETag`, `Last-Modified`, effective `max-age` / `Expires`. Missing validators → none cached, conditional requests skipped for that header.
2. Next refresh time = `min(server max-age or Expires if present, CEDARLING_POLICY_STORE_REFRESH_INTERVAL)` clamped by `_MAX`, plus randomized jitter to avoid thundering herd.
3. On each tick:
   - If validators cached → send `HEAD` (or conditional `GET`) with them
   - If no validators ever received from this endpoint → send plain `GET`, compare body hash to last-loaded hash; skip swap if identical
   - `304` → keep current policy store, reschedule
   - `200` with new content → parse, atomic swap on success
   - `200` with identical body hash → keep current, reschedule (handles servers ignoring conditional headers)
   - `405` / `501` on `HEAD` → switch endpoint to conditional `GET`, log warning once
   - Malformed cache header values → log debug, treat header as absent, continue
   - `5xx` / network error / timeout → exponential backoff, keep stale policy store, log error, increment failure counter
4. `Cache-Control: no-cache` / `no-store` → always conditional `GET`, ignore cached `max-age`.

## Robustness Requirement

Implementation must tolerate every realistic server behavior without panic or data loss:
- Missing `ETag`, `Last-Modified`, `Cache-Control`, `Expires` headers
- Malformed header values (unparseable dates, negative `max-age`, invalid quoted-string `ETag`)
- Servers returning `200` with identical body instead of `304`
- Servers ignoring conditional request headers
- Truncated / partial responses
- Connection resets mid-download
- Slow-loris responses (covered by request timeout)

All failure paths keep the previously loaded policy store active.

## Background Worker

- Spawn via existing `cedarling/src/http/spawn_task.rs` abstraction (tokio on native, `wasm_bindgen_futures::spawn_local` on WASM)
- Sleep via existing `cedarling/src/async_sleep.rs` abstraction (tokio on native, `gloo_timers` on WASM)
- Worker shuts down cleanly when Cedarling instance is dropped

## Concurrency / Atomicity

- Active policy store stored as `ArcSwap<PolicyStore>` (`arc_swap` crate already used in `jwt::key_service`, `jwt::validation::validator_cache`)
- Readers `load()` an `Arc<PolicyStore>` lock-free per authorization request
- Refresh worker calls `store()` after full successful load + parse — single atomic pointer swap, no mutex contention
- Readers in-flight when swap happens continue using their already-loaded `Arc`; next request observes new store

## Load Failure Handling

If response body cannot be processed — ZIP extraction fails (CjarUrl), `metadata.json` missing or invalid, Cedar schema parse fails, `PolicySet::from_str` fails, Lock Master response shape invalid (LockServer) — keep the previously loaded policy store, log the error, treat refresh as failed (counts toward backoff). Do not panic, do not serve an empty policy set.

## Observability

Metrics (added to `MetricsCollector` operational stats, `HashMap<String, i64>`):

- `policy_store_refresh_last_success_timestamp_seconds` — wall-clock time of last successful refresh (including 304 Not Modified)
- `policy_store_refresh_last_attempt_timestamp_seconds` — wall-clock time of last refresh attempt regardless of outcome
- `policy_store_refresh_next_attempt_timestamp_seconds` — scheduled time of next refresh attempt
- `policy_store_refresh_consecutive_failures` — count of consecutive failed attempts, reset to `0` on success

Last outcome — exactly one of the keys below is present with value `1`; the others are absent from the map. On each new attempt the previous outcome key is removed and the new one is set:

- `policy_store_refresh_last_outcome_success`
- `policy_store_refresh_last_outcome_not_modified`
- `policy_store_refresh_last_outcome_http_error`
- `policy_store_refresh_last_outcome_network_error`
- `policy_store_refresh_last_outcome_parse_error`

Logs per refresh attempt carry the full detail (source URL, HTTP method, status code, validator values sent/received, outcome, duration) — counters intentionally minimal.

## Out of Scope

- Manual refresh API
- Push-based updates (Long pooling, SSE)
- Refresh for local sources (`FileJson`, `FileYaml`, `CjarFile`, `Directory`, `Json`, `Yaml`, `ArchiveBytes`)
- Auth headers for private URLs (`HttpClient` has no auth support today)
- Semantic policy validation beyond Cedar parse-time checks

haileyesus2433:
### Prepare

-------------------

#### Target issue
  
  
  
  
closes #issue-number-here

#### Implementation Details
  

-------------------

Closes #14127, 

## Summary by CodeRabbit

* **New Features**
  * Opt-in background refresh for remote policy stores with configurable interval, atomic in-memory swaps, three-stage fetch strategy, byte-hash short-circuiting, conditional/HEAD-aware fetching, capped backoff/jitter, and snapshot-consistent authorization lookups.
  * Cache-validator-aware HTTP fetch APIs and sparse, persisted refresh telemetry (outcomes, strategies, streaks, cumulative counters).

* **Documentation**
  * Docs and README examples updated to document refresh config, refresh semantics, cache-header handling, and emitted metrics.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/6d952791-0b96-4d9c-8dff-3e3275255292?refs=ddb775c5f3b3762fe83f21e8d33235620ddbd3c7&source=prChecksComment) |

mo-auto:
Error: Hi @haileyesus2433, You did not reference an open issue in your [PR](#14174). I attempted to create an [issue](#14175) for you.
Please update that [issues'](#14175) title and body and make sure I correctly referenced it in the above PRs body.

haileyesus2433:
@coderabbitai please review

jans-cedarling/cedarling/src/authz/metrics.rs
```diff
@@ -566,6 +756,15 @@ impl MetricsCollector {
     }
 }
 
+/// Inserts `value` into `map` under `key` only when `value` is non-zero.
+/// Used to keep "no observation yet" indistinguishable from missing in the
+/// emitted metric map.
+fn insert_if_nonzero(map: &mut HashMap<String, i64>, key: &str, value: i64) {
```
olehbozhok: Instead of `i64`, we can take atomic for usability, but outcome will be the same

jans-cedarling/cedarling/src/bootstrap_config/raw_config/config.rs
```diff
@@ -443,6 +443,15 @@ pub struct BootstrapConfigRaw {
...
     pub status_list_refresh_interval_max: u64,
+
+    /// Base refresh interval, in seconds, for periodic background refresh of
+    /// remote policy stores (`CjarUrl` / `LockServer`). `0` disables refresh and
+    /// preserves the load-once-at-startup behavior. Non-zero values below `5`
+    /// are clamped to `5`. A server `Cache-Control: max-age` / `Expires` hint
+    /// can *shorten* the next interval but never extends it.
+    #[serde(rename = "CEDARLING_POLICY_STORE_REFRESH_INTERVAL", default)]
+    #[serde(deserialize_with = "deserialize_policy_store_refresh_interval")]
+    pub policy_store_refresh_interval_secs: u64,
```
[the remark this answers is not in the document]
olehbozhok: It is OK; it doesn't have complex logic

jans-cedarling/cedarling/src/http/mod.rs
```diff
@@ -194,6 +196,106 @@ impl HttpClient {
...
+    /// [`HeadOutcome::NotSupported`] so the caller can downgrade to plain GET.
+    ///
+    /// Intentionally does **not** retry on failure: HEAD here is a probe, and
+    /// the surrounding refresh worker either downgrades to a plain GET on the
+    /// same tick or simply runs again on the next interval. Burning the retry
+    /// budget on a probe would only delay the downgrade.
+    pub(crate) async fn head_validators(&self, uri: &str) -> Result<HeadOutcome, HttpClientError> {
+        let response = self.raw_client.head(uri).send().await.map_err(|e| {
+            let msg = e.to_string();
+            HttpRequestError::new(HttpRequestReasonError::MaxRetriesExceeded, None)
```
dagregi, line 313: Is this error correct with single attempt and no retry but using `MaxRetriesExceeded`

jans-cedarling/cedarling/src/bootstrap_config/raw_config/json_util.rs
```diff
@@ -88,6 +89,23 @@ where
...
+) -> Result<u64, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let value: u64 = deserialize_or_parse_string_as_json(deserializer)?;
+    Ok(if value == 0 {
+        0
+    } else {
+        value.max(PolicyStoreConfig::MIN_REFRESH_INTERVAL_SECS)
+    })
```
olehbozhok: I think better to clamp value inside service initialization method. To have only one point of failure. And if it clamped we can add log warn message

jans-cedarling/cedarling/src/http/cache_headers.rs
```diff
@@ -0,0 +1,305 @@
...
+//! All malformed values are treated as absent. The worker never panics on a
+//! surprising header value — it falls back to its configured interval.
+
+use chrono::{DateTime, Utc};
+use reqwest::header::HeaderMap;
+use std::time::Duration;
+
+/// Cache state extracted from a single HTTP response.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub(crate) struct CacheValidators {
```
olehbozhok: Maybe better name for this structure is `CacheHeadersState`?

jans-cedarling/cedarling/src/init/policy_store.rs
```diff
@@ -139,6 +139,68 @@ async fn load_policy_store_from_lock_master(
     extract_first_policy_store(&agama_policy_store)
 }
 
+/// Parses already-fetched Lock-Master JSON bytes into a [`PolicyStoreWithID`].
+/// Used by the policy-store refresh worker, which performs the HTTP fetch itself
+/// to be able to send conditional-request headers.
+pub(crate) fn parse_lock_master_bytes(
+    bytes: &[u8],
+) -> Result<PolicyStoreWithID, PolicyStoreLoadError> {
+    let agama_policy_store: LegacyAgamaPolicyStore = serde_json::from_slice(bytes)?;
```
olehbozhok, line 240: We should check what type of policy store the Lock server will  support, I think it will be cjar file also.

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
```diff
@@ -0,0 +1,1198 @@
...
+
+impl Drop for PolicyStoreRefreshHandle {
+    fn drop(&mut self) {
+        drop(self.shutdown.take());
+    }
+}
+
+/// Spawn the background refresh worker. Returns a [`PolicyStoreRefreshHandle`]
+/// whose `Drop` signals the worker to exit.
+#[allow(clippy::too_many_arguments)]
```
dagregi: No need for this suppression

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
```diff
@@ -0,0 +1,1198 @@
...
+            source,
+            interval_secs,
+            http_client,
+            rebuilder,
+            authz_swap,
+            metrics,
+            log,
+            shutdown_rx,
+        )
+        .await;
```
dagregi: Maybe extract the common parameters into struct then you can remove the lint suppression for this function as well

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
```diff
@@ -0,0 +1,1198 @@
...
+}
+
+/// Per-source strategy tracking and transition counters.
+struct StrategyState {
+    current: RefreshStrategy,
+    /// Consecutive ticks where the current strategy didn't help.
+    degraded_count: u32,
+    /// When we last attempted to upgrade strategy via a Conditional probe.
+    /// Only relevant when `current != Conditional`.
+    last_probe_at: Option<DateTime<Utc>>,
```
olehbozhok, line 116: So maybe better to have this value inside respective enum?

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
```diff
@@ -0,0 +1,1198 @@
...
+        }
+    }
+}
+
+impl StrategyState {
+    /// Decide which strategy to run for this tick. Returns `(strategy,
+    /// is_probe)`. `is_probe = true` indicates we're attempting to upgrade
+    /// from a degraded state — the tick logic uses this to decide whether
+    /// to interpret outcomes as upgrade signals.
+    fn choose_for_tick(&mut self) -> (RefreshStrategy, bool) {
```
olehbozhok: It would be more readable to return structure with named parameters instead of `(RefreshStrategy, bool)`

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
```diff
@@ -0,0 +1,1198 @@
...
+        log,
+        url,
+        bytes,
+        new_hash,
+        new_validators,
+    )
+    .await
+}
+
+#[allow(clippy::too_many_arguments)]
```
olehbozhok: You can create structure with RO parameters

olehbozhok (changes requested):
## Review: background policy-store refresh

Design is solid — `Arc<ArcSwap<Authz>>` with per-request `.load()` snapshot gives a clean consistency model (in-flight authz finishes on the pre-swap store), shutdown via `oneshot`+`Arc` refcount is correct, and the strategy ladder / backoff are well covered by unit tests. Two things worth addressing before merge:

### 1. Redundant rebuild+swap on the first tick

`RefreshState::default()` starts with empty `validators` and `last_body_hash = None`, and the store loaded at startup never seeds it. So the first tick sends a conditional GET with no `If-None-Match`/`If-Modified-Since`, gets a full `200`, computes `hash != None`, and parses+rebuilds+swaps a store that is byte-identical to the one already loaded.

Not a correctness bug, but every process pays one wasted parse + JWKS-refetch + entity-builder rebuild + swap shortly after startup. Consider seeding `validators` / `last_body_hash` from the initial load so the first tick can short-circuit on `304` or the body hash.

```rust
// init/policy_store_refresh.rs — RefreshState starts blind:
let mut state = RefreshState::default(); // validators empty, last_body_hash None
```

### 2. `RefreshOutcome::ParseError` conflates parse and rebuild failures

In `parse_swap_and_record`, a failure in `rebuilder.rebuild(...)` (JWT service init, entity builder, trusted-issuer validation) is recorded as `ParseError` and logged as `"parse failure"`, even though the body parsed fine and the failure is in the rebuild stage.

```rust
let new_authz = match rebuilder.rebuild(parsed).await {
    Ok(a) => a,
    Err(e) => {
        state.consecutive_failures = state.consecutive_failures.saturating_add(1);
        log.log_any(/* ... "parse failure for {url}: {e}" ... */);
        return RefreshOutcome::ParseError; // <- actually a rebuild error
    }
};
```

This makes the `policy_store_refresh.outcome_parse_error` metric and the logs misleading on a dashboard (a bad JWKS endpoint shows up as a "parse error"). Suggest a distinct `RefreshOutcome::RebuildError` variant (new integer enum value, documented alongside the others), or at least correct the log wording.

### Nits (non-blocking)

- `CacheValidators::no_store` is parsed and documented as informational but never read.
- `TryFrom<PolicyStoreConfigRaw>` uses `Self { source, ..Default::default() }`, which builds and discards the default YAML string.

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
```diff
@@ -0,0 +1,1216 @@
...
+            state.strategy.force_degrade();
+            tick_plain_get(ctx, state).await
+        },
+        HeadOutcome::Headers(new_validators) => {
+            if !new_validators.has_validator() {
+                // No ETag, no Last-Modified — HEAD added nothing. Count
+                // toward degrade and fall through to GET so we still get
+                // hash-based detection.
+                state.strategy.record_degraded();
+                return tick_plain_get(ctx, state).await;
```
tareknaser, line 645: When a HEAD request succeeds but returns no validators, we call `record_degraded()` and then fall back to `tick_plain_get`. If that GET later succeeds and returns a fresh body, `parse_swap_and_record` resets `consecutive_failures` but the degrade step has already been recorded and isn't undone.

This means that the HEAD→GET path still loses progress even though the refresh ultimately succeeded. Is that intentional?? If the HEAD request didn't provide any useful information but the GET completed the refresh successfully, counting it as a degrade step seems a bit harsh.

It may make more sense to only record a degrade when the entire tick fails to produce a useful result

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
```diff
@@ -0,0 +1,1216 @@
...
+        Ok(b) => b.to_vec(),
+        Err(e) => {
+            state.consecutive_failures = state.consecutive_failures.saturating_add(1);
+            ctx.log.log_any(
+                LogEntry::new(BaseLogEntry::new_system_opt_request_id(LogLevel::WARN, None))
+                    .set_message(format!(
+                        "policy store refresh: decode error for {url}: {e} (status {status})"
+                    )),
+            );
+            return RefreshOutcome::NetworkError;
```
tareknaser: Body decode failures are currently mapped to `RefreshOutcome::NetworkError` but the request itself completed successfully. This seems more like a response processing issue than a network connectivity failure (the upstream responded but the body couldn't be decoded)

The metrics taxonomy already distinguishes between `NetworkError`, `HttpError`, and `ParseError` so would it make more sense to classify this closer to `HttpError` (or maybe add a dedicated variant)? That would make it easier to distinguish between "couldn't reach the upstream" and "received a response but couldn't read the body" when investigating refresh failures.

jans-cedarling/cedarling/src/init/policy_store.rs
```diff
@@ -222,10 +376,14 @@ async fn load_policy_store_from_cjar_url(
...
+    Ok(LoadedPolicyStore {
+        store: PolicyStoreWithID {
+            id: store_id,
+            store: legacy_store,
+            metadata: Some(store_metadata),
+        },
+        body_hash: Some(body_hash),
+        validators,
     })
 }
```
tareknaser: The native and WASM variants of `load_policy_store_from_cjar_url` are near-identical. The only real difference is the `spawn_blocking` wrapper for `convert_to_legacy`. I thikn they could be unified by extracting the shared fetch + parse into one helper and gating only the `convert_to_legacy` step behind cfg

Same structural pattern already shows up in `parse_cjar_bytes`

jans-cedarling/cedarling/src/http/cache_headers.rs
```diff
@@ -0,0 +1,295 @@
...
+            // Negative or zero ⇒ already stale, treat as fresh_for=0 so the
+            // worker will revalidate immediately on its next tick.
+            Some(if delta.num_seconds() > 0 {
+                Duration::from_secs(delta.num_seconds().cast_unsigned())
+            } else {
+                Duration::ZERO
+            })
+        } else {
+            None
+        };
```
tareknaser, line 113: `fresh_for` is currently calculated as `expires_at - now` (wall clock) but RFC 9111 §5.3 defines freshness lifetime as `Expires_value - Date_value` which uses the response's `Date` header rather than the client's current time.

Using local `now` here means any clock skew between the client and origin affects the computed freshness. For example, if a server's clock is 30 seconds ahead and it sends `Expires: <now+60s>`, a client whose clock is 30 seconds behind will see the response as almost stale immediately.

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14174/commits/fd511b9821a8dacb6e6af55a1a3523f4a887c05a

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14174/commits/fd511b9821a8dacb6e6af55a1a3523f4a887c05a

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14174/commits/fd511b9821a8dacb6e6af55a1a3523f4a887c05a

jans-cedarling/cedarling/src/bootstrap_config/raw_config/json_util.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14174/commits/0f10fe74a097ec57f4fdc18677436ba98787dc71

jans-cedarling/cedarling/src/http/cache_headers.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14174/commits/56b90f7f62cf860cf11691e79c8857f49f6f8c18

jans-cedarling/cedarling/src/init/policy_store.rs
haileyesus2433: i have added CJAR detecting in https://github.com/JanssenProject/jans/pull/14174/commits/d622267ee8751ccee7b9dcef5c9f6ce3e7280f12

jans-cedarling/cedarling/src/authz/metrics.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14174/commits/ad28ae0602a0f8d498f13016280f5d7a9cd7df20

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14174/changes/18b2ccfd1e907d749974133abac697734bc3c619

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14174/changes/18b2ccfd1e907d749974133abac697734bc3c619

haileyesus2433:
> ## Review: background policy-store refresh
> Design is solid — `Arc<ArcSwap<Authz>>` with per-request `.load()` snapshot gives a clean consistency model (in-flight authz finishes on the pre-swap store), shutdown via `oneshot`+`Arc` refcount is correct, and the strategy ladder / backoff are well covered by unit tests. Two things worth addressing before merge:
> 
> ### 1. Redundant rebuild+swap on the first tick
> `RefreshState::default()` starts with empty `validators` and `last_body_hash = None`, and the store loaded at startup never seeds it. So the first tick sends a conditional GET with no `If-None-Match`/`If-Modified-Since`, gets a full `200`, computes `hash != None`, and parses+rebuilds+swaps a store that is byte-identical to the one already loaded.
> 
> Not a correctness bug, but every process pays one wasted parse + JWKS-refetch + entity-builder rebuild + swap shortly after startup. Consider seeding `validators` / `last_body_hash` from the initial load so the first tick can short-circuit on `304` or the body hash.
> 
> ```rust
> // init/policy_store_refresh.rs — RefreshState starts blind:
> let mut state = RefreshState::default(); // validators empty, last_body_hash None
> ```
> 
> ### 2. `RefreshOutcome::ParseError` conflates parse and rebuild failures
> In `parse_swap_and_record`, a failure in `rebuilder.rebuild(...)` (JWT service init, entity builder, trusted-issuer validation) is recorded as `ParseError` and logged as `"parse failure"`, even though the body parsed fine and the failure is in the rebuild stage.
> 
> ```rust
> let new_authz = match rebuilder.rebuild(parsed).await {
>     Ok(a) => a,
>     Err(e) => {
>         state.consecutive_failures = state.consecutive_failures.saturating_add(1);
>         log.log_any(/* ... "parse failure for {url}: {e}" ... */);
>         return RefreshOutcome::ParseError; // <- actually a rebuild error
>     }
> };
> ```
> 
> This makes the `policy_store_refresh.outcome_parse_error` metric and the logs misleading on a dashboard (a bad JWKS endpoint shows up as a "parse error"). Suggest a distinct `RefreshOutcome::RebuildError` variant (new integer enum value, documented alongside the others), or at least correct the log wording.
> 
> ### Nits (non-blocking)
> * `CacheValidators::no_store` is parsed and documented as informational but never read.
> * `TryFrom<PolicyStoreConfigRaw>` uses `Self { source, ..Default::default() }`, which builds and discards the default YAML string.

resolved from https://github.com/JanssenProject/jans/pull/14174/commits/3b6251ccd66f95abf8e90825d103df9 -https://github.com/JanssenProject/jans/pull/14174/commits/305e9356601651229dd339f15353cfb666755244

jans-cedarling/cedarling/src/http/mod.rs
haileyesus2433: Addressed in commits https://github.com/JanssenProject/jans/commit/c382f4e69d1aff6234d27e31128a6e3d3a36edc3 to https://github.com/JanssenProject/jans/commit/305e9356601651229dd339f15353cfb666755244

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14174/commits/0ff5591b2cdae27094cc6ff5034080b06f7ac325

jans-cedarling/cedarling/src/http/cache_headers.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14174/commits/932c20c92fa7c32d83bb0975926a4464bf51e02c

jans-cedarling/cedarling/src/init/policy_store.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14174/commits/a9f337abe0f65d156f8bc8c8188ef3d8d0c0e374

jans-cedarling/cedarling/src/bootstrap_config/raw_config/json_util.rs
olehbozhok: I don't see any logging in this commit

haileyesus2433: sorry my mistake in here https://github.com/JanssenProject/jans/pull/14174/commits/6eccf3e15816b5f766f4e2b85d25f2a138b600aa

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
haileyesus2433: Good catch on the asymmetry. However, this behavior is intentional.

`consecutive_failures` and `degraded_count` are tracking different concerns and serve different purposes, so I don't think they should be merged.

- `consecutive_failures` tracks **tick reliability** and drives the exponential backoff logic. `next_delay()` uses this value to increase the wait time between failed refresh attempts. It is reset on `Success` and `NotModified` because those outcomes indicate the refresh cycle produced a usable result.
- `degraded_count` tracks **strategy effectiveness** and drives the `Conditional → HeadThenGet → PlainGet` fallback ladder. It is reset via `record_helped()` only when the current strategy provides value that the next-cheaper strategy could not.

For `HeadThenGet`, that means things like:

- A `304 Not Modified`, where validators allow us to avoid downloading the body.
- A HEAD response that returns validators, allowing us to determine whether a subsequent GET is necessary.

The "HEAD returned no validators" case is intentionally treated as a signal that `HeadThenGet` is not providing value over `PlainGet`.

Even if the subsequent GET succeeds, the HEAD request itself was effectively a wasted round trip: `PlainGet` would have produced the same result with one fewer RTT. The purpose of the strategy ladder is to stop paying that cost when a server consistently fails to provide validators.

If degradation were tied only to the overall tick outcome, a server that never emits validators would keep the worker stuck in `HeadThenGet` indefinitely, paying an extra RTT on every refresh cycle despite receiving no benefit from the HEAD request.

dagregi (dismissed):

jans-cedarling/cedarling/src/bootstrap_config/decode.rs
```diff
@@ -164,6 +117,68 @@ impl BootstrapConfig {
...
 }
 
+/// Build [`PolicyStoreConfig`] from the three mutually-exclusive raw source
+/// fields (`CEDARLING_POLICY_STORE_LOCAL`, `_URI`, `_LOCAL_FN`).
+/// Returns an error if none or more than one are set, and also normalizes the
+/// URI case (`.cjar` suffix → archive source, otherwise Lock Server).
+fn build_policy_store_config(
+    raw: &BootstrapConfigRaw,
+) -> Result<PolicyStoreConfig, BootstrapConfigLoadingError> {
+    match (
```
olehbozhok, line 128: isn't more readable to use if-else statement?

olehbozhok: I see, current approach is better

jans-cedarling/cedarling/src/init/policy_store.rs
```diff
@@ -74,23 +74,51 @@ fn extract_first_policy_store(
...
+/// - failing that, still short-circuit on the body-hash compare when the
+///   upstream returns a byte-identical body despite ignored conditional
+///   headers.
+///
+/// Both are `None` / empty for non-URL sources (local files, inline JSON/YAML,
+/// in-memory archive bytes) — the refresh worker doesn't spawn there anyway.
+pub(crate) struct LoadedPolicyStore {
+    pub store: PolicyStoreWithID,
+    pub body_hash: Option<u64>,
+    pub validators: crate::http::cache_headers::CacheHeadersState,
```
olehbozhok: maybe we can import it in header of file?

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
```diff
@@ -0,0 +1,1226 @@
...
+        }
+    }
+
+    async fn parse(&self, bytes: &[u8]) -> Result<PolicyStoreWithID, PolicyStoreLoadError> {
+        // Magic-byte sniff — the ZIP local-file-header signature `PK\x03\x04`
+        // disambiguates `.cjar` archives from JSON regardless of source type.
+        // Future-proofs the Lock Server path: if Lock Server starts serving
+        // `.cjar` archives at a URL whose suffix doesn't end in `.cjar`, we
+        // route to the archive parser instead of failing with a JSON error.
+        if bytes.starts_with(b"PK\x03\x04") {
```
olehbozhok, line 433: I saw in code that we also check using `.ends_with(".cjar")`, should we also try to use that approach? But as additional we also cen use current (magic bytes) approach.

dagregi: This was mentioned in #12953 and was addressed in https://github.com/JanssenProject/jans/pull/14201  I was waiting for this pr to be merged

jans-cedarling/cedarling/src/init/service_config.rs
```diff
@@ -16,6 +16,17 @@ use bootstrap_config::BootstrapConfig;
 pub(crate) struct ServiceConfig {
     pub policy_store: PolicyStoreWithID,
     pub http_client: HttpClient,
+    /// Hash of the policy-store body bytes captured during initial load — `Some`
+    /// only for URL-based sources (Lock Server, `.cjar` URL). Seeded into the
+    /// refresh worker so the first tick can short-circuit when the upstream
+    /// returns a byte-identical body. `None` for local sources (the refresh
+    /// worker doesn't spawn there).
+    pub initial_body_hash: Option<u64>,
```
olehbozhok: Instead of passing body hash, maybe it worth to move 
`load_policy_store(&bootstrap.policy_store_config, &http_client).await?;` to service initialization method?

olehbozhok (dismissed):

jans-cedarling/cedarling/src/init/policy_store.rs
```diff
@@ -74,23 +74,51 @@ fn extract_first_policy_store(
...
+) -> Result<LoadedPolicyStore, PolicyStoreLoadError> {
+    let loaded = match &config.source {
         PolicyStoreSource::Json(policy_json) => {
             let agama_policy_store = serde_json::from_str::<LegacyAgamaPolicyStore>(policy_json)
                 .map_err(PolicyStoreLoadError::ParseJson)?;
-            extract_first_policy_store(&agama_policy_store)?
+            LoadedPolicyStore {
+                store: extract_first_policy_store(&agama_policy_store)?,
+                body_hash: None,
+                validators: crate::http::cache_headers::CacheHeadersState::default(),
```
olehbozhok: So you used import for `CacheHeadersState`  on 94 line but decided to leave all other lines?

jans-cedarling/cedarling/src/init/service_config.rs
haileyesus2433: resolved in here https://github.com/JanssenProject/jans/pull/14174/commits/19363aad54961e20df0cd4f1185c4521661435ce

jans-cedarling/cedarling/src/init/policy_store.rs
haileyesus2433: addressed the remaining in here https://github.com/JanssenProject/jans/pull/14174/commits/ef8ac07e22f132adc22517dd4f631cba461ed701

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
olehbozhok: I thought about having value in the `RefreshStrategy` enum. If it makes sense.

olehbozhok (dismissed):
Pls look closely at https://github.com/JanssenProject/jans/pull/14174#pullrequestreview-4449498830  (comments outside the diff).
And create issues for all relevant points. It would be great to write links to the issues in the comments of this PR

dagregi (dismissed):

haileyesus2433:
> Pls look closely at [#14174 (review)](https://github.com/JanssenProject/jans/pull/14174#pullrequestreview-4449498830) (comments outside the diff). And create issues for all relevant points. It would be great to write links to the issues in the comments of this PR

most of the comments could be addresses in this PR so i have fixed them in https://github.com/JanssenProject/jans/pull/14174/commits/c6c21ed0a3f55bd7bb90a53eaff9cfdd088f0094 i only left the cjar detection related comments which are being worked on in https://github.com/JanssenProject/jans/pull/14201 by @dagregi

jans-cedarling/cedarling/src/init/policy_store_refresh.rs
```diff
@@ -0,0 +1,1442 @@
...
+            ..StrategyState::default()
+        };
+        s.force_degrade();
+        assert_eq!(
+            s.current,
+            RefreshStrategy::PlainGet,
+            "force_degrade at the floor must leave current unchanged",
+        );
+        assert_eq!(s.conditional_to_head_transitions, 0);
+        assert_eq!(s.head_to_plain_transitions, 0);
```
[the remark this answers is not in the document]
haileyesus2433: this is fine

jans-cedarling/http_utils/src/lib.rs
```diff
@@ -388,4 +408,59 @@ mod tests {
...
+        let max_retries =
+            HttpRequestError::new(HttpRequestReasonError::MaxRetriesExceeded, None);
+        assert!(!max_retries.is_decode_error());
+        assert!(max_retries.is_max_retries_exceeded());
+
+        let status =
+            HttpRequestError::new(HttpRequestReasonError::HttpStatusError, None);
+        assert!(!status.is_decode_error());
+        assert!(!status.is_max_retries_exceeded());
+        assert!(status.is_http_status_error());
```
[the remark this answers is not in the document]
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14174/commits/71ab100876a61e5c7305ceac818889a4156db9e2

jans-cedarling/cedarling/src/http/mod.rs
```diff
@@ -412,4 +557,285 @@ mod test {
...
+
+        let result = client
+            .get_bytes_conditional(&url, &validators)
+            .await
+            .expect("request");
+
+        assert!(matches!(
+            result,
+            super::ConditionalFetch::NotModified { .. }
+        ));
```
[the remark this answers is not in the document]
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14174/commits/71ab100876a61e5c7305ceac818889a4156db9e2

jans-cedarling/cedarling/src/http/mod.rs
```diff
@@ -207,13 +216,148 @@ impl HttpClient {
...
+            .await?;
+
+        let status = response.status();
+        let headers = response.headers().clone();
+        let new_validators = CacheHeadersState::from_headers(&headers, chrono::Utc::now());
+
+        if status == reqwest::StatusCode::NOT_MODIFIED {
+            return Ok(ConditionalFetch::NotModified {
+                validators: new_validators,
+            });
```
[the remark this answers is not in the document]
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14174/commits/71ab100876a61e5c7305ceac818889a4156db9e2

haileyesus2433:
@coderabbitai if no more comments approve

dagregi (dismissed):

haileyesus2433:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> 🧩 Analysis chain
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> # Check the open review comments on PR `#14174`
> gh pr view 14174 --repo JanssenProject/jans --json reviewDecision,reviews,state
> echo "---"
> # List review threads/comments
> gh api repos/JanssenProject/jans/pulls/14174/reviews --jq '.[] | {id: .id, user: .user.login, state: .state, submitted_at: .submitted_at}'
> echo "---"
> # Check for unresolved review comments from coderabbitai
> gh api repos/JanssenProject/jans/pulls/14174/comments --jq '.[] | select(.user.login == "coderabbitai[bot]") | {id: .id, body: .body[:100], in_reply_to_id: .in_reply_to_id}'
> ```
> 
> Length of output: 133902
> 
> `@haileyesus2433` There are still open review comments from the last two review rounds that need to be addressed before I can approve. Here's a summary of the outstanding items:
> 
> **From the latest reviews (2026-06-09), not yet resolved:**
> 
> 🟠 **Major — outside diff (need fixes before approval):**
> 
> 1. **`no-store` freshness not zeroed** (`cache_headers.rs` ~line 120–142): `Cache-Control: no-store, max-age=600` still produces `fresh_for=600` instead of forcing revalidation — `no-store` must be treated the same as `no-cache` in `parse_cache_control`.
> 2. **HEAD validators committed before GET/rebuild succeeds** (`policy_store_refresh.rs` ~line 616–625): `state.validators` is updated with `new_validators` before the downstream GET/rebuild path completes. If the GET fails, future HEADs will incorrectly short-circuit as `NotModified` while the older policy set is still active.
> 3. **Reprobe cooldown not started on downgrade** (`policy_store_refresh.rs` ~line 151–176): Neither `record_degraded()` nor `force_degrade()` stamps `last_probe_at`, so the first tick after every downgrade immediately retries `Conditional` instead of waiting `STRATEGY_REPROBE_INTERVAL_SECS`.
> 
> 🔵 **Trivial (nits — missing descriptive assertion messages):**
> 
> * Several `assert!`/`assert_eq!`/`assert_ne!` calls in tests across `http/mod.rs`, `http_utils/src/lib.rs`, and `policy_store_refresh.rs` still lack the required descriptive messages per the project's coding guidelines.
> 
> Once those major items are addressed, I'm happy to approve. The trivial nits can be done in the same pass or filed as follow-up work if preferred.

@coderabbitai all of those are addresed from commit [c6c21ed](https://github.com/JanssenProject/jans/pull/14174/commits/c6c21ed0a3f55bd7bb90a53eaff9cfdd088f0094) - [71ab100](https://github.com/JanssenProject/jans/pull/14174/commits/71ab100876a61e5c7305ceac818889a4156db9e2)

dagregi (dismissed):

olehbozhok:
## Review: 4 findings 

Checked each against current `main`. All items that affected correctness (unbounded body read, error misclassification, stale policy_count gauge, sticky fresh_for, validator overwrite, force_degrade duplication, weak ETag comparison) are fixed. What remains is correctness-of-observability and maintenance debt.

---

### 1. `choose_for_tick` skips `HeadThenGet` as an intermediate upgrade step

**Severity: Low** (narrow scenario, no data loss)

**What:** `choose_for_tick()` (policy_store_refresh.rs:153) has two arms: when `current == Conditional`, stay put; otherwise probe `Conditional` after `STRATEGY_REPROBE_INTERVAL_SECS` (1800s). There is no "probe `HeadThenGet` first" arm. A worker at `PlainGet` must jump directly to a full conditional GET (body download included) to test whether the upstream has recovered; it never tries the lighter `HeadThenGet` probe.

**Failure scenario:** Upstream supports ETag/Last-Modified but ignores `If-None-Match` (returns `200` with the same body instead of `304`). The conditional GET downloads the full body, the hash matches, and the short-circuit kicks in with `is_probe=true`, so neither `record_degraded()` nor `record_probe_success()` fires. `current` stays `PlainGet`. The worker retries after 1800s — same result forever. A `HeadThenGet` probe (cheap HEAD, no body) would have shown the validators match, triggering `record_helped()` and upgrading to `HeadThenGet`.

**Suggestion:** In `choose_for_tick()`, when `current == PlainGet`, probe `HeadThenGet` (not `Conditional`). Only probe `Conditional` from `HeadThenGet`. This gives a two-step upgrade: `PlainGet → HeadThenGet → Conditional`, matching the degradation ladder.

---

### 2. `AuthzRebuilder::rebuild` unconditionally re-fetches JWKS on every refresh

**Severity: Medium** (unnecessary HTTP traffic + background task churn on every policy change)

**What:** `rebuild()` (policy_store_refresh.rs:257) calls `JwtService::new(...)` on every invocation, which performs OIDC discovery + JWKS fetch for every trusted issuer, then spawns per-issuer background pollers (`keep_jwks_updated`) and a token-cache cleanup task. This happens on every policy-store refresh even when the `trusted_issuers` set is byte-identical to the currently active one.

**Impact:** A policy-only edit (Cedar rules change, but `trusted_issuers` unchanged) triggers N OIDC + JWKS HTTP fetches plus N+1 new background tasks on every refresh tick. Old `Arc<JwtService>` instances survive until the old `Arc<Authz>` is reclaimed, so multiple JWKS-polling tasks coexist for the same issuer during the overlap window. On a short refresh interval (<60s) against a frequently-changing policy body, this multiplies background task count linearly.

**Suggestion:** Pass `current_authz: Option<&Authz>` to `rebuild()`. When the new `trusted_issuers` deeply-equals the current one, clone the existing `Arc<JwtService>` instead of constructing a new one.

---

### 3. `rebuild` duplicates `ServiceFactory::authz_service` pipeline, drops the default-entities warn loop

**Severity: Low** (silent log omission; drift risk on future changes)

**What:** `rebuild()` (policy_store_refresh.rs:261–296) re-implements the same construction steps as `ServiceFactory::authz_service` + `entity_builder` + `jwt_service` (service_factory.rs:62–171) but omits the `default_entities.warns()` log loop (service_factory.rs:125–136). Any future change to the startup construction (new validation step, new config field) must be manually mirrored here, or the refresh path silently builds a differently-configured `Authz` than the one used at startup.

**Suggestion:** Extract a shared `fn build_authz(policy_store, jwt_config, ...) -> Result<Authz, ...>` called by both `ServiceFactory` and `AuthzRebuilder`.

---

### 4. 14 `policy_store_refresh_*` atomics hand-listed in 3+ places with no compiler guard

**Severity: Low** (maintenance hazard — omission compiles silently)

**What:** The 14 atomic counters are declared as individual fields in `MetricsCollector` (metrics.rs:284–309), zero-initialised in both `MetricsCollector::new()` and `::disabled()`, and re-listed in `inject_policy_store_refresh_state()` for snapshot emission. Adding or removing a counter requires identical edits in all three locations (plus the `record_policy_store_refresh` match); an omission compiles silently — zeros either never get set or the snapshot silently drops a field.

**Suggestion:** Extract a `#[derive(Default)] PolicyStoreRefreshMetrics` sub-struct with one `fn inject(...)` and one `fn record_outcome(...)`. Replace the 14 inline fields with a single `policy_store_refresh: PolicyStoreRefreshMetrics` in `MetricsCollector`.

haileyesus2433:
> ## Review: 4 findings
> Checked each against current `main`. All items that affected correctness (unbounded body read, error misclassification, stale policy_count gauge, sticky fresh_for, validator overwrite, force_degrade duplication, weak ETag comparison) are fixed. What remains is correctness-of-observability and maintenance debt.
> 
> ### 1. `choose_for_tick` skips `HeadThenGet` as an intermediate upgrade step
> **Severity: Low** (narrow scenario, no data loss)
> 
> **What:** `choose_for_tick()` (policy_store_refresh.rs:153) has two arms: when `current == Conditional`, stay put; otherwise probe `Conditional` after `STRATEGY_REPROBE_INTERVAL_SECS` (1800s). There is no "probe `HeadThenGet` first" arm. A worker at `PlainGet` must jump directly to a full conditional GET (body download included) to test whether the upstream has recovered; it never tries the lighter `HeadThenGet` probe.
> 
> **Failure scenario:** Upstream supports ETag/Last-Modified but ignores `If-None-Match` (returns `200` with the same body instead of `304`). The conditional GET downloads the full body, the hash matches, and the short-circuit kicks in with `is_probe=true`, so neither `record_degraded()` nor `record_probe_success()` fires. `current` stays `PlainGet`. The worker retries after 1800s — same result forever. A `HeadThenGet` probe (cheap HEAD, no body) would have shown the validators match, triggering `record_helped()` and upgrading to `HeadThenGet`.
> 
> **Suggestion:** In `choose_for_tick()`, when `current == PlainGet`, probe `HeadThenGet` (not `Conditional`). Only probe `Conditional` from `HeadThenGet`. This gives a two-step upgrade: `PlainGet → HeadThenGet → Conditional`, matching the degradation ladder.
> 
> ### 2. `AuthzRebuilder::rebuild` unconditionally re-fetches JWKS on every refresh
> **Severity: Medium** (unnecessary HTTP traffic + background task churn on every policy change)
> 
> **What:** `rebuild()` (policy_store_refresh.rs:257) calls `JwtService::new(...)` on every invocation, which performs OIDC discovery + JWKS fetch for every trusted issuer, then spawns per-issuer background pollers (`keep_jwks_updated`) and a token-cache cleanup task. This happens on every policy-store refresh even when the `trusted_issuers` set is byte-identical to the currently active one.
> 
> **Impact:** A policy-only edit (Cedar rules change, but `trusted_issuers` unchanged) triggers N OIDC + JWKS HTTP fetches plus N+1 new background tasks on every refresh tick. Old `Arc<JwtService>` instances survive until the old `Arc<Authz>` is reclaimed, so multiple JWKS-polling tasks coexist for the same issuer during the overlap window. On a short refresh interval (<60s) against a frequently-changing policy body, this multiplies background task count linearly.
> 
> **Suggestion:** Pass `current_authz: Option<&Authz>` to `rebuild()`. When the new `trusted_issuers` deeply-equals the current one, clone the existing `Arc<JwtService>` instead of constructing a new one.
> 
> ### 3. `rebuild` duplicates `ServiceFactory::authz_service` pipeline, drops the default-entities warn loop
> **Severity: Low** (silent log omission; drift risk on future changes)
> 
> **What:** `rebuild()` (policy_store_refresh.rs:261–296) re-implements the same construction steps as `ServiceFactory::authz_service` + `entity_builder` + `jwt_service` (service_factory.rs:62–171) but omits the `default_entities.warns()` log loop (service_factory.rs:125–136). Any future change to the startup construction (new validation step, new config field) must be manually mirrored here, or the refresh path silently builds a differently-configured `Authz` than the one used at startup.
> 
> **Suggestion:** Extract a shared `fn build_authz(policy_store, jwt_config, ...) -> Result<Authz, ...>` called by both `ServiceFactory` and `AuthzRebuilder`.
> 
> ### 4. 14 `policy_store_refresh_*` atomics hand-listed in 3+ places with no compiler guard
> **Severity: Low** (maintenance hazard — omission compiles silently)
> 
> **What:** The 14 atomic counters are declared as individual fields in `MetricsCollector` (metrics.rs:284–309), zero-initialised in both `MetricsCollector::new()` and `::disabled()`, and re-listed in `inject_policy_store_refresh_state()` for snapshot emission. Adding or removing a counter requires identical edits in all three locations (plus the `record_policy_store_refresh` match); an omission compiles silently — zeros either never get set or the snapshot silently drops a field.
> 
> **Suggestion:** Extract a `#[derive(Default)] PolicyStoreRefreshMetrics` sub-struct with one `fn inject(...)` and one `fn record_outcome(...)`. Replace the 14 inline fields with a single `policy_store_refresh: PolicyStoreRefreshMetrics` in `MetricsCollector`.

i have created a separate issue for all of them in under improvements #14245 #14244 #14243 #14242

haileyesus2433:
@coderabbitai approve

dagregi (approved):

PR #14201, opened 2026-06-03
feat(jans-cedarling): improve policy store URI detection using content-type

### LINKED ISSUE #12953: feat(jans-cedarling): improve policy store URI detection using content type

dagregi:
### Prepare

-------------------

#### Target issue
  
  
  
  
closes #12953

#### Implementation Details
  

-------------------

## Summary by CodeRabbit

* **New Features**
  * Policy stores can now be specified via URI; the system auto-detects whether the fetched content is an archive or legacy/API format using response headers and content.
  * URI-based policy stores are treated as remote sources and can participate in scheduled refresh when configured.
* **Documentation**
  * Updated `CEDARLING_POLICY_STORE_URI` guidance to use content-based detection (archive vs legacy/API), not file extensions.
* **Tests**
  * Added integration and loader tests to validate URI loading, archive detection, and authorization success.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/50b810c7-bb64-4c47-8d88-2a4df69a3b88?refs=3c5b4176385d48da0e5d1920ccae105575d0991f&source=prChecksComment) |

jans-cedarling/cedarling/src/init/policy_store.rs
```diff
@@ -198,11 +204,57 @@ pub(crate) async fn load_policy_store(
...
 
+/// Loads the policy store from a URI with content-type-based format detection.
+async fn load_policy_store_from_uri(
+    uri: &str,
+    http_client: &HttpClient,
+    strict_schema_validation: bool,
+) -> Result<LoadedPolicyStore, PolicyStoreLoadError> {
+    if uri.to_lowercase().ends_with(".cjar") {
+        return load_policy_store_from_cjar_url(uri, http_client, strict_schema_validation).await;
+    }
```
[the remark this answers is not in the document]
dagregi: This was left for a faster path for cjar endpoints that end with '.cjar'

jans-cedarling/cedarling/src/bootstrap_config/policy_store_config.rs
```diff
@@ -118,6 +118,14 @@ pub enum PolicyStoreSource {
     /// in the directory structure format (`metadata.json`, `schema.cedarschema`, `policies/`, etc.).
     Directory(PathBuf),
 
+    /// An unresolved URI whose source type (archive vs lock server) is detected
+    /// at load time via HTTP content-type inspection and/or magic byte checking.
+    ///
+    /// During loading, [`load_policy_store`](crate::init::policy_store::load_policy_store)
+    /// resolves this by making an HTTP request to determine whether the URI points
+    /// to a Cedar Archive (`.cjar`) or a Lock Master endpoint.
+    Uri(String),
```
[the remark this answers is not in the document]
haileyesus2433: This introduces a critical regression in the refresh path for the new `PolicyStoreSource::Uri` variant. While the PR correctly routes `CEDARLING_POLICY_STORE_URI` into `PolicyStoreSource::Uri(_)` (`decode.rs`) and updates `refresh_enabled()` (`policy_store_config.rs`) to treat it as refresh-capable, the corresponding refresh source mapping was never updated. `RefreshSource::from_policy_store_source` (`init/policy_store_refresh.rs`) still only handles `LockServer` and `CjarUrl`, causing `PolicyStoreSource::Uri(_)` to fall through to `None`. Because the caller in `lib.rs` immediately short-circuits on that `None`, any configuration using `CEDARLING_POLICY_STORE_URI` together with `CEDARLING_POLICY_STORE_REFRESH_INTERVAL_SECS` will silently stop spawning the background refresh worker after this change. There is no warning or log message to indicate the failure, making this particularly difficult for operators to detect. As a result, the configuration appears refresh-enabled (`refresh_enabled()` returns `true`) while refresh never actually runs. This effectively breaks background policy updates for what is likely the most common URL-based configuration path introduced by the PR. The existing test coverage (`refresh_source_construction_from_url` in `policy_store_refresh.rs`) was not extended to include the new `Uri` variant, which allowed the regression to go unnoticed.

jans-cedarling/cedarling/src/init/policy_store.rs
```diff
@@ -198,11 +204,53 @@ pub(crate) async fn load_policy_store(
...
+        .headers()
+        .get(reqwest::header::CONTENT_TYPE)
+        .and_then(|v| v.to_str().ok())
+        .unwrap_or("")
+        .to_owned();
+
+    let bytes = http_client.read_response_capped(response).await?;
+    let body_hash = crate::init::policy_store_refresh::body_hash(&bytes);
+
+    if (content_type.is_empty()
```
haileyesus2433: The new content-type detection logic is still too restrictive and will reject valid archive responses from real-world servers. Currently, archive parsing only occurs when the body starts with ZIP magic bytes and the Content-Type is either empty, application/octet-stream, or application/zip. That means legitimate ZIP responses using common MIME types such as application/x-zip-compressed (the legacy IIS MIME type for .zip) or vendor-specific archive types like application/x-cedar-archive will incorrectly fall through to the JSON parsing path and fail with a ParseJson error. The behavior is also unhelpful when an upstream server returns an HTML error page (for example, a login redirect or 401 page), since the code attempts JSON parsing and surfaces a cryptic serde_json::Error instead of explaining that content-type detection failed. This makes the implementation more fragile than the previous .cjar suffix-based approach for the exact misconfigured-MIME scenarios the issue is trying to address. A more robust approach would be to trust known JSON and ZIP MIME types, use content sniffing only when the MIME is ambiguous, and return a clear detection error for unexpected content types rather than a misleading JSON parse failure.

jans-cedarling/cedarling/src/init/policy_store.rs
```diff
@@ -40,6 +43,9 @@ pub enum PolicyStoreLoadError {
     /// Failed to load policy store from directory.
     #[error("Failed to load policy store from directory: {0}")]
     Directory(String),
+    /// Failed to detect policy store from an unresolved URI.
```
haileyesus2433: I think this variant is added but **never constructed** anywhere in the diff. The detection failure path (`load_policy_store_from_uri`) instead falls through to `parse_lock_master_bytes`, which emits `ParseJson`. Either remove the variant, or Actually emit it from the detection logic.

jans-cedarling/cedarling/src/init/policy_store.rs
haileyesus2433: also `contains` matches substrings, so a header value of `application/octet-stream-encrypted` or `vendor/wrap-application/zip-foo` would pass. The robust way is to parse the MIME essence (everything before the first `;`, lowercased, trimmed). i think `reqwest` already pulls `mime` transitively  you can do `mime::Mime::from_str(&ct).ok().map(|m| m.essence_str().to_owned())`, or at minimum `ct.split(';').next().unwrap().trim().eq_ignore_ascii_case("application/zip")`. Fix this together with the detection rewrite

haileyesus2433 (changes requested):
In `decode.rs`:

```rust
/// Build [`PolicyStoreConfig`] from the three mutually-exclusive raw source
/// fields (`CEDARLING_POLICY_STORE_LOCAL`, `_URI`, `_LOCAL_FN`).
/// Returns an error if none or more than one are set, and also normalizes the
/// URI case (`.cjar` suffix → archive source, otherwise Lock Server).
//   ^^^ no longer true — normalization moved to load time
```

```rust
// Case: get the policy store from a URI (auto-detect .cjar archives)
//                                        ^^^ no longer true at this layer
(None, Some(policy_store_uri), None) => Ok(PolicyStoreConfig {
    source: PolicyStoreSource::Uri(policy_store_uri),
    ...
```

In `policy_store_config.rs:19-23`:

```rust
/// Base refresh interval in seconds for URL-based policy store sources
/// (`CjarUrl`, `LockServer`). `0` disables background refresh ...
//                  ^^^ Uri is missing from this list
```

Please update both stale doc-comments

jans-cedarling/cedarling/src/init/policy_store.rs
dagregi: resolved in https://github.com/JanssenProject/jans/pull/14201/commits/c2bc20a69a831830b3db5affd1ad0466d971d038 since we already have the byte the content-type detection is unnecessary

jans-cedarling/cedarling/src/init/policy_store.rs
dagregi: removed in https://github.com/JanssenProject/jans/pull/14201/commits/c2bc20a69a831830b3db5affd1ad0466d971d038

olehbozhok:
We don't need to update the jans-cedarling/Cargo.lock file. 
It will be updated in https://github.com/JanssenProject/jans/pull/14259

olehbozhok (changes requested):
## Code Review: URI content-based policy store detection

### 1. [nit] `CjarUrl` / `LockServer` variants are now unreachable from config

`decode.rs` no longer creates `PolicyStoreSource::CjarUrl` or `PolicyStoreSource::LockServer` — all URIs go through `PolicyStoreSource::Uri`. These variants are still alive in `From<&str>` (serde deserialization) and test code, so nothing is broken. Still, worth either removing them or filing a follow-up cleanup issue.

### 2. [medium] Missing integration test for `Uri` refresh path

There's a unit test for `RefreshSource::from_policy_store_source` with `Uri`, but no integration test verifying that:
- the refresh worker is actually spawned for `PolicyStoreSource::Uri`
- policy store updates after server-side content changes

Analogous tests exist for `LockServer`/`CjarUrl` — worth adding one for `Uri`.

### 3. [low] Content-Type header unused for fast-path detection

`load_policy_store_from_uri` always reads the full body first, then checks ZIP magic bytes. A fast path could inspect `content-type: application/json` and parse immediately without buffering the entire body. Minor optimization — not a blocker.

### 4. [wrong] Previous reviewer's "regression in refresh path" claim is incorrect

The comment claimed `RefreshSource::from_policy_store_source` doesn't handle `Uri`, causing silent fallthrough to `None`. This is **incorrect** — the PR already includes `PolicyStoreSource::Uri(u) => Some(Self::Uri { url: u.clone() })` (policy_store_refresh.rs:416), plus updated `url()` and `fetch_and_parse`. The refresh path works correctly.

dagregi:
> ## Code Review: URI content-based policy store detection
> ### 1. [nit] `CjarUrl` / `LockServer` variants are now unreachable from config
> 
> `decode.rs` no longer creates `PolicyStoreSource::CjarUrl` or `PolicyStoreSource::LockServer` — all URIs go through `PolicyStoreSource::Uri`. These variants are still alive in `From<&str>` (serde deserialization) and test code, so nothing is broken. Still, worth either removing them or filing a follow-up cleanup issue.
> ### 2. [medium] Missing integration test for `Uri` refresh path
> 
> There's a unit test for `RefreshSource::from_policy_store_source` with `Uri`, but no integration test verifying that:
> 
>     * the refresh worker is actually spawned for `PolicyStoreSource::Uri`
> 
>     * policy store updates after server-side content changes
> 
> 
> Analogous tests exist for `LockServer`/`CjarUrl` — worth adding one for `Uri`.
> ### 3. [low] Content-Type header unused for fast-path detection
> 
> `load_policy_store_from_uri` always reads the full body first, then checks ZIP magic bytes. A fast path could inspect `content-type: application/json` and parse immediately without buffering the entire body. Minor optimization — not a blocker.
> ### 4. [wrong] Previous reviewer's "regression in refresh path" claim is incorrect
> 
> The comment claimed `RefreshSource::from_policy_store_source` doesn't handle `Uri`, causing silent fallthrough to `None`. This is **incorrect** — the PR already includes `PolicyStoreSource::Uri(u) => Some(Self::Uri { url: u.clone() })` (policy_store_refresh.rs:416), plus updated `url()` and `fetch_and_parse`. The refresh path works correctly.

1. Users can still use from config by using `PolicyStoreSource` when initializing cedarling with `Cedarling::new()`, for example if you change the policy source [here](https://github.com/JanssenProject/jans/pull/14201/changes#diff-439e7744bb96f80f22a6d31413c11e2125354ae5e3e843e1bf8b23a98e321a1fR1295) to `PolicyStoreSource::CjarUrl(uri)` it would still accept it and work fine
2. Integration test already exists [here](https://github.com/JanssenProject/jans/pull/14201/changes#diff-439e7744bb96f80f22a6d31413c11e2125354ae5e3e843e1bf8b23a98e321a1fR1277)
3.  Since content-type is server-controlled content sniffing is better plus @haileyesus2433 mentioned it if the server returns with a custom content-type it would still be a failure

olehbozhok (changes requested):
## Review: Documentation and docstring cleanup needed

The core logic changes look good after the refactor (dropping Content-Type heuristic, using only ZIP magic bytes). However, several docs and docstrings still reference the old approach. These should be cleaned up before merging.

### 1. `docs/cedarling/reference/cedarling-policy-store.md:26`

```
Cedarling supports two policy store formats and automatically detects the correct format based on file extension or URL:
```

"file extension or URL" is misleading — URI detection no longer uses file extension or URL patterns. Should say something like:

```
Cedarling supports two policy store formats and automatically detects the correct format based on the file extension (for local files) or the response body content (for URIs):
```

### 2. `docs/cedarling/reference/cedarling-policy-store.md:28-35` — Detection table

| Current | Should be |
|---------|-----------|
| `CEDARLING_POLICY_STORE_URI`, response body is an archive | `CEDARLING_POLICY_STORE_URI`, response body starts with ZIP magic bytes |
| `CEDARLING_POLICY_STORE_URI`, any other response body | `CEDARLING_POLICY_STORE_URI`, response body is JSON (Lock Master format) |

### 3. `docs/cedarling/reference/cedarling-properties.md:33-35`

```
- URL points to an archive → loads as Cedar Archive (`.cjar`)
- Other URLs → loads as legacy JSON from Lock Server
```

This implies URL-based detection, which is no longer the case. Should reference magic byte detection:

```
- Response body starts with ZIP magic bytes → loads as Cedar Archive (`.cjar`)
- Response body is JSON → loads as legacy JSON from Lock Server
```

### 4. `docs/cedarling/reference/cedarling-properties.md:49` — Refresh interval description

```
pointing at a Lock Server endpoint or a `.cjar`
```

The `.cjar` reference here is a leftover from the old extension-based detection. Since `Uri` sources now cover both cases, this should be updated to reflect that all `CEDARLING_POLICY_STORE_URI` sources are eligible for refresh:

```
pointing at a Lock Server endpoint or a Cedar Archive URL
```

### 5. `jans-cedarling/cedarling/src/bootstrap_config/policy_store_config.rs:123-124` — `Uri` docstring

```rust
/// An unresolved URI whose source type (archive vs lock server) is detected
/// at load time via HTTP content-type inspection and/or magic byte checking.
```

"HTTP content-type inspection" was dropped in commit `c2bc20a69a8`. Should be just:

```rust
/// An unresolved URI whose source type (archive vs lock server) is detected
/// at load time via ZIP magic byte checking on the response body.
```

### 6. `jans-cedarling/cedarling/src/bootstrap_config/decode.rs:121-123` — Truncated docstring

```rust
/// Build [`PolicyStoreConfig`] from the three mutually-exclusive raw source
/// fields (`CEDARLING_POLICY_STORE_LOCAL`, `_URI`, `_LOCAL_FN`).
/// Returns an error if none or more than one are set, and also normalizes the
fn build_policy_store_config(
```

The phrase "and also normalizes the" is a dangling fragment from when `.cjar` suffix normalization was documented. Should be removed:

```rust
/// Build [`PolicyStoreConfig`] from the three mutually-exclusive raw source
/// fields (`CEDARLING_POLICY_STORE_LOCAL`, `_URI`, `_LOCAL_FN`).
/// Returns an error if none or more than one are set.
fn build_policy_store_config(
```

None of these are blocking issues, but they would prevent confusion for future readers.

olehbozhok (dismissed):

olehbozhok (approved):

olehbozhok:
@coderabbitai do you have suggestions, or can you approve?

olehbozhok:
@coderabbitai approve

haileyesus2433 (approved):

PR #14450, opened 2026-06-27
fix(jans-cedarling)!: fail closed when a status list refresh fails

### LINKED ISSUE #14409: Revoked tokens can stay accepted if status-list refresh keeps failing
If I understand the current flow correctly, Cedarling fetches a status list once when setting up an issuer and that initial fetch is fail-closed: if it fails, tokens that depend on that list are rejected. After that, a background task refreshes the list every `ttl`.

I am not sure what happens if those later refreshes start failing. In `jwt/status_list/cache.rs`, a failed fetch just gets logged and the loop continues so we keep serving the last successfully cached status list:

```rust
let status_list_jwt =
    match StatusListJwtStr::get_from_url(&status_list_url, &http_client).await {
        Ok(jwt) => jwt,
        Err(e) => {
            logger.log_any(JwtLogEntry::new(
                format!(
                    "failed to fetch an updated the status list from '{0}': {1}",
                    status_list_url.as_str(),
                    e
                ),
                Some(LogLevel::ERROR),
            ));
            continue;
        },
    };
```

The validate/deserialize steps right after this do the same thing. They log and `continue`.

So if a token is revoked after the last good refresh, and refreshes keep failing after that, Cedarling will keep accepting it based on stale revocation data. What I want to say is that a revoked token can stay valid indefinitely as long as the refresh path never recovers.

I'd like to talk this through before writing any code because the fix is a judgment call. Serving the stale list on a failed refresh might be on purpose to stay available when the issuer has a temporary outage. If we do want to bound it, a few options:

- reject tokens once the cached list is older than some multiple of `ttl`
- add a config property for the max age we'll trust, off by default
- drop the list after a few refreshes in a row fail

tareknaser:
### Description

Right now the background task that refreshes a trusted issuer's status list keeps serving the last cached copy when a refresh fails. So if a token is revoked after the last good refresh and the status endpoint stays unreachable, Cedarling keeps accepting that token (#14409). This PR makes the refresh fail closed: when it can't get a fresh list, it drops the cached one so tokens that depend on it get rejected until the endpoint comes back.

The drop is scoped to the one issuer whose endpoint is down, so other issuers keep working. I kept it contained like this rather than disabling Cedarling globally, since one unreachable status endpoint shouldn't take down all authorization. Happy to widen it if you'd prefer.

#### Implementation Details

All in `cedarling/src/jwt/status_list/cache.rs` (`keep_status_list_updated`):

- On a failed refresh, drop the cached list instead of keeping the stale copy. The validator already rejects a token whose list is missing, so dependent tokens fail closed on their own. Logs a WARN once when the list is dropped.
- Clear the per-issuer token cache too, so a token already cached as valid gets re-checked instead of riding its own TTL.
- Re-add the list on the next good refresh so it recovers by itself.
- Add a test for the drop, plus a small `fail_status_list_endpoint` helper on the test mock server.

It fails closed on the first failed refresh (the fetch already retries at the HTTP layer, so an error means the endpoint is really down). Easy to soften to "after N failures in a row" if you'd rather.

-------------------

Closes #14409

## Summary by CodeRabbit

* **Bug Fixes**
  * Status list refreshes now “fail closed” by evicting stale cached entries when a refresh fails or returns an invalid/unparseable status list.
  * Token validation is tightened after a failed refresh, rejecting tokens that rely on the cached status list until the next successful refresh.
  * Successful refreshes now reliably repopulate the cache, including recreating entries after eviction.
* **Tests**
  * Added a test to verify cached status-list eviction when the status endpoint becomes unavailable.
* **Documentation**
  * Updated reference docs and configuration comments to describe the new fail-closed refresh behavior and interval limit setting.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/2a6f66ec-21e2-4471-a932-dd420c322cdc?refs=eb5ebca7350687bd1483e20fd9faef4f4c8226c6&source=prChecksComment) |

dagregi (dismissed):

olehbozhok (approved):
I added documentation message

PR #14461, opened 2026-06-30
perf(jans-cedarling/bindings/python): cut PyO3 conversion overhead on hot path  

### LINKED ISSUE #14028: perf(jans-cedarling): drive Cedarling token authorization below 1 ms across core and all bindings
### Description

- Development target issue
  - Cross-platform performance epic.

- Implementation Details
  - Remove the four dominant hot-path costs identified in the audit:
    1. Triple JWT parse in the signed pass.
    2. JSON round-trip for the Cedar `Context` build (both passes).
    3. Per-request `clear_expired` write-lock sweep on the token cache.
    4. `AuthorizeResult` clone inside the latency timer.
  - Defensively eliminate concurrency ceilings (sharded token cache, `ArcSwap` validator/JWKS, lock-free pushed-data store, lock-free memory logger).
  - Cut binding overhead at the FFI boundary per language.
  - Lock the burn-down with per-binding CI perf gates.

-------------------

### Prepare

- [ ] Read contribution guidelines
- [ ] Read license information
- [ ] Confirm the goal contract: `< 1 ms` mean, warm steady state, cache-hit, `≤ 4` tokens.
- [ ] Confirm cold-start and network-going validation are out of scope.

-------------------

### Identified code changes

Track via the sub-issues below. Each must close before the epic closes.

-------------------

- [ ] Mean latency `< 1 ms` for both passes in all binding CI lanes on the standard fixtures.
- [ ] p95 inside the agreed variance window in all binding CI lanes.
- [ ] Allow path with logging off allocates zero `String`/`Vec`/`HashMap` in the per-request layer (tracking allocator on a debug build).
- [ ] 32-thread authorize storm shows no global writer contention.
- [ ] No correctness regressions in JWT validation, policy decisions, or binding API behavior.

-------------------

### Document the changes

- [ ] Update binding READMEs (Java/UniFFI, Go, Python, WASM, C) with the new performance profile and the `< 1 ms` contract.
- [ ] Update [`jans-cedarling/cedarling/README.md`](jans-cedarling/cedarling/README.md) with the new logging profile recommendations.

haileyesus2433:
### Prepare

-------------------

#### Target issue
  
  
  
  
closes #14028

#### Implementation Details
  

-------------------

## Summary by CodeRabbit

* **Performance Improvements**
  * Reduced repeated Python-to-internal conversions during authorization by reusing cached representations of request `tokens` and `context` when building internal request inputs.
* **Bug Fixes**
  * Improved Python request field handling with more explicit, controlled properties; `context` is now exposed as read-only via a `MappingProxyType` view, while updates use dedicated setters to refresh cached state predictably.
* **Documentation**
  * Refreshed Cedarling Python docs: clarified JSON-compatible `DataEntry.value` types, updated unsigned request parameter naming (`principal`), documented removed methods, adjusted error variants, and added `PolicyEffect` and `PolicyMetadata`.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/93069ce8-18a5-4ed7-989c-7df87435a8a9?refs=3477543e9278573ebac36cf35e779c125d34fb5f&source=prChecksComment) |

jans-cedarling/bindings/cedarling_python/src/authorize/request_unsigned.rs
```diff
@@ -32,15 +33,21 @@ use serde_pyobject::from_pyobject;
...
+    #[pyo3(get, set)]
     pub action: String,
     /// cedar_policy resource data
+    #[pyo3(get, set)]
     pub resource: EntityData,
     /// context to be used in cedar_policy
+    #[pyo3(get, set)]
     pub context: Py<PyDict>,
+    /// Sticky cache of `context` deserialized to `serde_json::Value`; populated on first `to_cedarling()` call so subsequent calls skip the PyDict→Value round-trip.
+    cached_context: OnceLock<serde_json::Value>,
```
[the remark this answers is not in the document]
haileyesus2433: this is also resolved in https://github.com/JanssenProject/jans/pull/14461/commits/24fabd54e289051c1b68a8c8ff2a32699f8e622f

olehbozhok (dismissed):

olehbozhok:
Maybe we should write somewhere to reuse the same structure

dagregi (dismissed):

olehbozhok (dismissed):

jans-cedarling/bindings/cedarling_python/src/authorize/request_unsigned.rs
```diff
@@ -58,20 +68,34 @@ impl RequestUnsigned {
...
     pub fn to_cedarling(&self) -> Result<cedarling::RequestUnsigned, PyErr> {
         let principal = self.principal.clone().map(|p| p.into());
 
-        let context = Python::attach(|py| -> Result<serde_json::Value, PyErr> {
-            let context = self.context.clone_ref(py).into_bound(py);
-            from_pyobject(context).map_err(|err| {
-                PyRuntimeError::new_err(format!("Failed to convert context to json: {}", err))
-            })
-        })?;
+        let context = if let Some(cached) = self.cached_context.get() {
```
tareknaser: The setter clears the cache on reassignment (`req.context = {...}`) but not on in-place updates. Since `context` is exposed via `#[pyo3(get)]`, `req.context["k"] = v` mutates the shared dictionary directly without invalidating the cached value.

AFAIK, the getter returns the same underlying dict, and `authorize_unsigned`/`authorize_multi_issuer` both reuse the same request object by calling `request.borrow().to_cedarling()` so the `OnceLock` persists across calls. That means `to_cedarling()` continues returning the original converted context even after in-place edits. 

For example, if the initial context converts successfully and is later mutated to something non-JSON, `to_cedarling()` still returns the cached value instead of failing.

Would it make sense for the getter to return a copy (or otherwise prevent in-place mutation), rather than relying on the field comment to document this behavior? The same applies in `request_multi_issuer.rs`.

jans-cedarling/bindings/cedarling_python/src/authorize/request_multi_issuer.rs
```diff
@@ -40,16 +41,29 @@ use serde_pyobject::from_pyobject;
 ///     context={"location": "miami"}
 /// )
 /// ```
-#[pyclass(get_all, set_all)]
+#[pyclass]
 pub struct AuthorizeMultiIssuerRequest {
     /// List of TokenInput objects
+    /// (custom setter below clears `cached_tokens` on assignment; in-place
+    /// edits like `req.tokens.append(t)` are NOT detected and leave the cache
```
tareknaser: `req.tokens.append(t)` can't leave the cache stale because it doesn't actually modify the underlying field.

`tokens` is `#[pyo3(get)]` over a `Vec<TokenInput>`, so every access returns a new Python list. Calling `req.tokens.append(t)` only appends to that temporary list. it never writes back to `req.tokens`.

tareknaser: So unlike `context` (which is a shared dictionary where in-place mutations can indeed leave the cache stale), this warning doesn't seem to describe a situation that can actually happen.

Unless I'm missing something, I think we should reword this comment so it doesn't read the same as the one for `context`

haileyesus2433: Same treatment applied from https://github.com/JanssenProject/jans/pull/14461/commits/e6604fb3f4aaf48eefa8dd22c807214d45fd9b3e -  https://github.com/JanssenProject/jans/pull/14461/commits/121511848d6fa952c0423f303b9fc90366576dfa

jans-cedarling/bindings/cedarling_python/src/authorize/request_unsigned.rs
haileyesus2433: you're right that the current setter only handles the reassignment path and in-place edits on the shared dict leave the cache stale.

On the "return a copy from the getter" suggestion i explored that but landed somewhere slightly different. A plain dict.copy() would break the natural Python idiom silently  req.context["k"] = v would mutate a throwaway copy and the write would vanish with no signal, which feels like a strictly worse failure mode than the staleness it's trying to prevent.

Instead, I've changed the getter to return a types.MappingProxyType view over the underlying dict. That makes in-place mutation fail loudly at the mutation site with TypeError 'mappingproxy' object does not support item assignment, so the user can't accidentally desync the cache. Reassignment (req.context = {...}) continues to work as before and refreshes the cache via the existing setter.

from https://github.com/JanssenProject/jans/pull/14461/commits/e6604fb3f4aaf48eefa8dd22c807214d45fd9b3e -  https://github.com/JanssenProject/jans/pull/14461/commits/121511848d6fa952c0423f303b9fc90366576dfa

jans-cedarling/bindings/cedarling_python/src/authorize/request_multi_issuer.rs
haileyesus2433: So, CodeRabbit flagged the same cached_context design on a second pass and pointed out mutation vectors that MappingProxyType didn't cover (external aliases to the source dict, nested-container mutations reachable through the read-only view). Both are real staleness paths, and silent-wrong-answer on an authorization decision isn't worth defending a micro-optimization.

So I've rolled back the context caching on both RequestUnsigned and AuthorizeMultiIssuerRequest. context is back to plain #[pyo3(get, set)], and to_cedarling() deserializes on every call  always reflects the live dict, no cache to go stale.

Kept cached_tokens on AuthorizeMultiIssuerRequest. It's safe against the same class of mutation vectors: Vec<TokenInput> extraction clones the wrappers (TokenInput: Clone + from_py_object), and the #[pyo3(get)] getter materializes a fresh PyList each access  so external mutations to source TokenInput objects or in-place edits like req.tokens.append(t) can't reach the underlying Vec. The only way to mutate is reassignment, which the setter catches and invalidates the cache on.

dagregi (approved):

PR #14564, opened 2026-07-16
feat(jans-cedarling): add batch authorize benchmarks and cross-binding contract coverage

### LINKED ISSUE #14494: test(jans-cedarling): add batch authorize benchmarks and cross-binding contract coverage

Batch authorization is a first-class API and its semantic contract must hold
identically across every binding. We need reproducible benchmarks so setup
amortization is visible in numbers and doesn't silently regress, plus contract
tests that verify the batch APIs behave identically across bindings.

**Benchmarks.** Extend the existing benchmark suite with N=10 and N=25 batch
scenarios for both flows, comparing:

- `authorize_unsigned_batch(N items)` vs. N × `authorize_unsigned`.
- `authorize_multi_issuer_batch(N items)` vs. N × `authorize_multi_issuer`.

Report wall-clock speedup, per-item p50/p95, and setup-amortization overhead.
Wire into the cross-binding benchmark contract established for existing calls.

**Contract tests (shared across bindings).**

- **Result equivalence.** Given identical inputs and no state changes between
  calls, `authorize_*_batch(items)` returns results equal to the sequence
  `items.map(authorize_*)` — same decisions, same diagnostics reasons.
- **Ordering.** Shuffle items; verify results reorder with them.
- **Empty-items rejection.** Returns the same `BatchValidationError` variant in
  every binding.
- **Per-item Cedar evaluation error** lands in that item's result, not as a
  batch-level failure.
- **`batch_id` correlation.** The response's `batch_id` is a valid UUIDv7 and
  matches the `batch_id` field on the emitted per-item decision-log entries.

N/A

**Additional context**
N/A

haileyesus2433:
### Prepare

-------------------

#### Target issue
  
  
  
  
closes #14494 

#### Implementation Details
  

-------------------

### DIFF
--- a/jans-cedarling/cedarling/src/tests/batch_contract.rs
+++ b/jans-cedarling/cedarling/src/tests/batch_contract.rs
@@ -114,18 +114,11 @@ fn diagnostic_reasons(result_response: &cedar_policy::Response) -> Vec<String> {
 #[tokio::test]
 async fn batch_unsigned_matches_sequence_of_single() {
     let cedarling = unsigned_cedarling().await;
-    let principal_ok = create_test_principal(
-        "Jans::TestPrincipal1",
-        "p1",
-        json!({"is_ok": true}),
-    )
-    .expect("principal should build");
-    let principal_bad = create_test_principal(
-        "Jans::TestPrincipal1",
-        "p1",
-        json!({"is_ok": false}),
-    )
-    .expect("principal should build");
+    let principal_ok = create_test_principal("Jans::TestPrincipal1", "p1", json!({"is_ok": true}))
+        .expect("principal should build");
+    let principal_bad =
+        create_test_principal("Jans::TestPrincipal1", "p1", json!({"is_ok": false}))
+            .expect("principal should build");
 
     // Mixed: allow-item, deny-item (bad principal), allow-item.
     let items = vec![
@@ -163,10 +156,10 @@ async fn batch_unsigned_matches_sequence_of_single() {
         .await
         .expect("batch should succeed");
 
-    assert_eq!(
-        seq_ok.len(),
-        batch_ok.results.len(),
-        "result count must match between batch and sequence"
+    assert_eq!(seq_ok.len(), batch_ok.results.len());
+    assert!(
+        batch_ok.results.iter().all(|r| r.decision),
+        "is_ok=true principal must Allow every item — check fixture drift"
     );
     for (i, (s, b)) in seq_ok.iter().zip(batch_ok.results.iter()).enumerate() {
         assert_eq!(s.decision, b.decision, "decision mismatch at item {i}");
@@ -199,6 +192,10 @@ async fn batch_unsigned_matches_sequence_of_single() {
         .await
         .expect("batch should succeed");
 
+    assert!(
+        batch_bad.results.iter().all(|r| !r.decision),
+        "is_ok=false principal must Deny every item — check fixture drift"
+    );
     for (i, (s, b)) in seq_bad.iter().zip(batch_bad.results.iter()).enumerate() {
         assert_eq!(s.decision, b.decision, "deny decision mismatch at item {i}");
         assert_eq!(
@@ -239,17 +236,18 @@ async fn batch_multi_issuer_matches_sequence_of_single() {
     }
 
     let batch = cedarling
-        .authorize_multi_issuer_batch(BatchAuthorizeMultiIssuerRequest::new(
-            tokens,
-            items.clone(),
-        ))
+        .authorize_multi_issuer_batch(BatchAuthorizeMultiIssuerRequest::new(tokens, items.clone()))
         .await
         .expect("multi-issuer batch should succeed");
 
+    assert_eq!(sequence.len(), batch.results.len());
+    // Positive/negative anchors — allow items at 0/2, deny items at 1/3.
+    // Guards against a fixture drift where both sides silently all-Deny.
+    let decisions: Vec<bool> = batch.results.iter().map(|r| r.decision).collect();
     assert_eq!(
-        sequence.len(),
-        batch.results.len(),
-        "result count must match between batch and sequence"
+        decisions,
+        vec![true, false, true, false],
+        "batch decisions must match the allow/deny/allow/deny item pattern"
     );
     for (i, (s, b)) in sequence.iter().zip(batch.results.iter()).enumerate() {
         assert_eq!(s.decision, b.decision, "decision mismatch at item {i}");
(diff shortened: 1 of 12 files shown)


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/63be02e1-7ede-4a10-916f-a8ff39eb6522?refs=2c35cde1383eea8e44f0e659f621e72c22e65ddb&source=prChecksComment) |

haileyesus2433:
@coderabbitai review

olehbozhok (changes requested):
**Suggestion: add pre-bench decision validation for Rust criterion benchmarks**

Five of six binding benches (C, Go, Python, WASM, Java) validate that every
item returns `Allow` *before* collecting timing samples — catching
misconfigured fixtures or unexpected Deny before measurement begins.

The Rust criterion bench (`authz_authorize_batch_benchmark.rs`) is the
outlier: it only asserts result count inside the measured loop and never
checks decisions. If the policy engine changes behaviour or fixtures drift,
the bench silently measures a Deny path instead of the intended Allow path.

Concrete fix for `bench_unsigned_batch`:

```rust
// Pre-bench trial: confirm every item would Allow
let trial = runtime
    .block_on(cedarling.authorize_unsigned_batch(request.clone()))
    .expect("trial batch call");
assert!(
    trial.results.iter().all(|r| r.decision),
    "all items must Allow — check fixtures"
);
```

Same pattern applies to the other three bench functions
(`bench_unsigned_sequence`, `bench_multi_issuer_batch`,
`bench_multi_issuer_sequence`). The trial uses `runtime.block_on`, which is
safe because the outer functions already own a `Runtime`.

jans-cedarling/cedarling/src/tests/batch_contract.rs
```diff
@@ -0,0 +1,369 @@
...
+        ))
+        .await
+        .expect("batch should succeed");
+
+    assert_eq!(
+        seq_ok.len(),
+        batch_ok.results.len(),
+        "result count must match between batch and sequence"
+    );
+    for (i, (s, b)) in seq_ok.iter().zip(batch_ok.results.iter()).enumerate() {
```
tareknaser, line 169: This test (and `batch_multi_issuer_matches_sequence_of_single` below) only checks that the batch result matches the sequence-of-single result but not what those decisions should actually be. If a fixture or policy drift made every item Deny, the batch and the sequence would both deny and the test would still pass (all-Deny equals all-Deny)

I think the multi-issuer one is the sharper case here. It builds `multi_issuer_allow_item()` and `multi_issuer_deny_item(...)` but it never checks that the "allow" items actually allow.

The reverse-order test right below pins `baseline_decisions == vec![true, false, ...]` and the per-flow tests in `authorize_multi_issuer_batch.rs` pin decisions too so this feels like a gap that's specific to these two tests.

We should probably add a positive/negative anchor here. For example, assert that the allow items come back `true` and the bad-principal / wrong-id items come back `false`.

jans-cedarling/cedarling/src/tests/batch_contract.rs
```diff
@@ -0,0 +1,369 @@
...
+    )
+    .expect("principal should build");
+    let principal_bad = create_test_principal(
+        "Jans::TestPrincipal1",
+        "p1",
+        json!({"is_ok": false}),
+    )
+    .expect("principal should build");
+
+    // Mixed: allow-item, deny-item (bad principal), allow-item.
```
tareknaser: nit: the comment says three items with a deny in the middle but `items` is two same-shaped allow items. The deny path actually comes later from `principal_bad`

jans-cedarling/cedarling/src/tests/batch_contract.rs
tareknaser: This is the same as "don't silently pass on a Deny path" idea Oleh raised on the bench

tareknaser:
> **Suggestion: add pre-bench decision validation for Rust criterion benchmarks**

cc: @olehbozhok

The `req.clone()` (O(N) for the batch, per item for the sequence) and the `assert_eq!` both happen inside the timed closure so their cost ends up being part of the measurement.

If I understand correctly, the goal here is to measure the setup amortization. Was including the clone in the measured section intentional or would it make more sense to move it to the setup step so both sides are measuring the same thing?

jans-cedarling/cedarling/src/tests/batch_contract.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14564/commits/efe618c999164323de4c19184207aa523ca4f65d

jans-cedarling/cedarling/src/tests/batch_contract.rs
haileyesus2433: resolved https://github.com/JanssenProject/jans/pull/14564/commits/dcb36f1cca12c2e4cc4886410a83b10af2dfc296

haileyesus2433:
> > **Suggestion: add pre-bench decision validation for Rust criterion benchmarks**
> 
> cc: @olehbozhok
> 
> The `req.clone()` (O(N) for the batch, per item for the sequence) and the `assert_eq!` both happen inside the timed closure so their cost ends up being part of the measurement.
> 
> If I understand correctly, the goal here is to measure the setup amortization. Was including the clone in the measured section intentional or would it make more sense to move it to the setup step so both sides are measuring the same thing?

Fair point, and no it wasn't intentional. The req.clone() is there because authorize_unsigned_batch takes the request by value (moves it) and criterion's iter closure runs N times, each needing its own owned input. black_box(req.clone()) was the shortest thing that satisfied the type-checker; measuring it was an accident, not a design choice. its resolved in here https://github.com/JanssenProject/jans/pull/14564/commits/e92e81a0726ff3a0ab4412ef9eddb1ab5c232ff2

haileyesus2433:
> **Suggestion: add pre-bench decision validation for Rust criterion benchmarks**
> 
> Five of six binding benches (C, Go, Python, WASM, Java) validate that every item returns `Allow` _before_ collecting timing samples — catching misconfigured fixtures or unexpected Deny before measurement begins.
> 
> The Rust criterion bench (`authz_authorize_batch_benchmark.rs`) is the outlier: it only asserts result count inside the measured loop and never checks decisions. If the policy engine changes behaviour or fixtures drift, the bench silently measures a Deny path instead of the intended Allow path.
> 
> Concrete fix for `bench_unsigned_batch`:
> 
> ```rust
> // Pre-bench trial: confirm every item would Allow
> let trial = runtime
>     .block_on(cedarling.authorize_unsigned_batch(request.clone()))
>     .expect("trial batch call");
> assert!(
>     trial.results.iter().all(|r| r.decision),
>     "all items must Allow — check fixtures"
> );
> ```
> 
> Same pattern applies to the other three bench functions (`bench_unsigned_sequence`, `bench_multi_issuer_batch`, `bench_multi_issuer_sequence`). The trial uses `runtime.block_on`, which is safe because the outer functions already own a `Runtime`.

resolved in https://github.com/JanssenProject/jans/pull/14564/commits/ad840e9ddedb6870c82b880cec51a7780a429e8f

dagregi (approved):

olehbozhok (approved):

PR #14624, opened 2026-07-24
chore(jans-cedarling): reduce PolicyStore visibility from pub to pub(crate)

### LINKED ISSUE #14614: chore(jans-cedarling): reduce `PolicyStore` visibility from `pub` to `pub(crate)`
**Issue:**

`PolicyStore` (`cedarling/src/common/policy_store.rs:44`) is currently declared `pub` and re-exported via `pub use crate::common::policy_store::PolicyStore` in `cedarling/src/lib.rs:90`. It is only used within the crate itself — in `init/policy_store.rs`, `common/policy_store/manager.rs`, and `common/policy_store/legacy_store/mod.rs`. No external consumer relies on this type directly.

**Proposal:**

1. Change `pub struct PolicyStore` → `pub(crate) struct PolicyStore`
2. Remove the `pub use` re-export from `lib.rs`
3. Ensure internal use paths still compile

This shrinks the crate's public API surface, which is a goal per the code style guidelines in `AGENTS.md` ("`pub` items must always be exported in the root `lib.rs` to be truly public" and "prefer to expose minimal API surface").

**Note:** All fields of `PolicyStore` are currently `pub` as well. Consider whether they should also be reduced to `pub(crate)` in the same change.

olehbozhok:
### Prepare

-------------------

#### Target issue
  
  
  
  
closes #14614

#### Implementation Details
  

**Goal:** reduce `PolicyStore` visibility from `pub` to `pub(crate)`.

`PolicyStore` was exported publicly (re-exported via `bindings` in `lib.rs`), but nothing outside the crate needs it. Making it `pub(crate)` first required removing the public API surface that referenced it and the dead code that kept unused fields/methods "live".

**Steps**

1. **Drop the public re-export** — removed `pub use crate::common::policy_store::PolicyStore;` from the `bindings` module in `lib.rs`.

2. **Remove dead `PolicyStore` surface** so nothing external depends on it:
   - Deleted unused fields `name`, `description`, `cedar_version`.
   - Deleted the unused `apply_default_entities_limits()` method (which in turn removed the `semver::Version` and `DefaultEntitiesLimits` dependencies, plus the `default_entities_limits` enforcement path and related legacy-store/manager helpers and tests).

3. **Downgrade visibility** — `PolicyStore` → `pub(crate)`. Applied the same narrowing to closely coupled types now crate-internal: `TrustedIssuer`, `PoliciesContainer`, and their methods.

4. **Simplify schema lookup** exposed by the cleanup — `get_entity_schema` returned `(type_name, type_schema)` but every caller discarded the schema half; replaced with `get_entity_type_name` returning only the name. Call sites in `build_ctx.rs` updated. This also let unused `Attribute` variant payloads (`String`, `Long`, `Boolean`, `Set`, `Extension`) become fieldless.

5. **Cleanup** — final commit resolves clippy warnings surfaced by the removals.

Net: no behavioral change; `PolicyStore` and its coupled types are now crate-private, and the compiler enforces the removed surface stays unused.

-------------------

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved entity reference resolution for entity and shared-entity attributes by consistently deriving the target entity type during schema lookup.
  * Missing or invalid entity identifiers now follow the required/optional contract consistently.
* **Changes**
  * Updated Cedar JSON schema handling (entity type-name resolution; action “applies_to” now supports context-only; simplified attribute shapes and required handling for primitive types).
  * Legacy policy-store metadata fields (name/description/Cedar version) are now ignored during loading.
* **Tests**
  * Refreshed JSON schema and policy-store fixtures/expected error messages to match the updated parsing behavior.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/7f1d6987-e271-47d8-82c8-572d2c193e4b?refs=035d979f6fdee0c85c90e613d407e590b6ebec0c&source=prChecksComment) |

jans-cedarling/cedarling/src/common/policy_store/legacy_store/mod.rs
```diff
@@ -468,13 +464,6 @@ impl<'de> Deserialize<'de> for LegacyPolicyStore {
...
             .ok_or_else(|| de::Error::custom("policy store entry must be a JSON object"))?;
 
-        let name = obj.get("name").ok_or_else(|| {
-            de::Error::custom("missing required field 'name' in policy store entry")
-        })?;
-        let name = name
-            .as_str()
-            .ok_or_else(|| de::Error::custom("'name' must be a string"))?;
-
         let schema = obj
```
tareknaser, line 477: I think dropping the `name` field also removes the `missing required field 'name'` validation that used to happen here. As far as I can tell, nothing actually reads `PolicyStore.name`, so removing the field itself makes sense. That said, this check was also validating that a legacy policy store included a `name` field when it was deserialized.

Unless I'm missing something, this path is still used since `init/policy_store.rs` deserializes `LegacyAgamaPolicyStore` for the JSON, YAML, and remote source variants.

Would it make sense to keep the validation even though the field is gone? The old code was checking two things: that `name` is present and that it's a string, so it seems useful to preserve both:

```rust
let name = obj.get("name").ok_or_else(|| {
    de::Error::custom("missing required field 'name' in policy store entry")
})?;
if !name.is_string() {
    return Err(de::Error::custom("'name' must be a string"));
}
```

The new format should be unaffected since `MetadataValidator::validate_policy_store_info` still rejects an empty name. I think this is only relevant for the legacy deserializer.

jans-cedarling/cedarling/src/common/policy_store/legacy_store/test.rs
```diff
@@ -157,86 +155,8 @@ fn test_broken_policy_parsing_error_in_policy_store() {
...
-    let err = result.expect_err("Expected error for missing policy_stores field");
-    assert!(
-        err.to_string()
-            .contains("missing required field 'policy_stores' in policy store"),
-        "Error should mention missing policy_stores, got: {err}"
-    );
-}
-
 #[test]
 fn test_invalid_policy_store_entry() {
```
tareknaser, line 159: I think `test_missing_required_fields` was removed from just above here but I don't think it was testing `name`. It was asserting `missing required field 'policy_stores' in policy store`, which is a different validation and one that still seems to exist in `legacy_store/mod.rs`.

So I wonder if this test just got caught in the blast radius by accident.

There is an identical assertion over in `common/policy_store/test.rs` that file isn't compiled (see my next comment). So I think after this PR there wouldn't be any test covering that validation anymore

jans-cedarling/cedarling/src/common/policy_store/test.rs
```diff
@@ -258,32 +258,11 @@ fn test_missing_required_fields() {
 
 #[test]
 fn test_invalid_policy_store_entry() {
```
tareknaser: I think this whole file is dead.

Looks like it got orphaned back in #13592 when the legacy split happened

jans-cedarling/cedarling/src/common/policy_store/legacy_store/mod.rs
olehbozhok: I don't really understand why we need to check what we don't use, but OK, I left this check.

jans-cedarling/cedarling/src/common/policy_store/legacy_store/test.rs
olehbozhok: Serde handles this automatically; the dead copy was removed. Not worth a separate test.

jans-cedarling/cedarling/src/common/policy_store/test.rs
olehbozhok: thanks, removed

dagregi (approved):

olehbozhok:
@coderabbitai do you have suggestions or can approve?

haileyesus2433 (approved):

jans-cedarling/cedarling/src/common/policy_store/legacy_store/test.rs
tareknaser: right. thanks for clarifying

PR #14635, opened 2026-07-27
feat(cedarling-js): add JavaScript runtime agnostic sdk for Cedarling.

### LINKED ISSUE #14582: feat(cedarling): add a stable JavaScript SDK over the WASM bindings
## Summary

Create a new JavaScript/TypeScript npm package that provides a stable, idiomatic, strongly typed and developer-friendly API over the existing `@janssenproject/cedarling_wasm` package.

The new package, provisionally named `@janssenproject/cedarling`, will become the primary Cedarling package promoted to JavaScript developers. The existing WASM package will remain the low-level implementation dependency.

## Motivation

The current `@janssenproject/cedarling_wasm` package exposes an API closely coupled to the generated WASM bindings. Applications may consequently depend on implementation details such as:

- Explicit WASM initialization
- Runtime-specific loading behavior
- Manual request serialization
- Generated method and class names
- WASM resource-management methods
- Changes to the JavaScript/WASM communication format

This creates a fragile API boundary for JavaScript consumers.

A stable JavaScript SDK would separate the public developer API from the underlying WASM implementation. Cedarling could then evolve its WASM communication, serialization, loading and performance strategies without unnecessarily introducing breaking changes for JavaScript applications.

The package should also offer the same clear and consistent Cedarling developer experience in browsers and server-side JavaScript environments.

## Goals

- Provide a stable and idiomatic JavaScript/TypeScript API
- Support browsers and server-side JavaScript
- Provide strong TypeScript types for configuration, authorization requests, decisions, diagnostics and errors
- Hide WASM initialization and runtime-loading details
- Hide serialization across the JavaScript/WASM boundary
- Prevent generated WASM objects and resource-management methods from leaking into application code
- Normalize Cedarling results into plain JavaScript objects
- Keep token validation and Cedarling authorization semantics explicit
- Allow the underlying WASM bindings to evolve independently from the public API
- Establish a documented semantic-versioning boundary for JavaScript consumers

## Proposed API direction

The SDK should expose one primary authorization API that works consistently in browsers and server-side JavaScript environments.

```ts
// This is sample Non Normative example

import { createCedarling } from "@janssenproject/cedarling";

const cedarling = await createCedarling({
  applicationName: "Task Manager",
  policyStore: {
    source: "url",
    url: "https://example.com/task-manager.cjar",
  },
});

const decision = await cedarling.authorize({
  principal: {
    kind: "tokens",
    tokens: [
      {
        mapping: "Jans::Access_token",
        token: accessToken,
      },
    ],
  },

  action: {
    type: "TaskManager::Action",
    id: "EditTask",
  },

  resource: {
    type: "TaskManager::Task",
    id: task.id,
    attributes: {
      ownerId: task.ownerId,
      assigneeId: task.assigneeId,
      status: task.status,
    },
  },

  context: {},
});

if (decision.allowed) {
  // Perform the requested operation.
}
```

Possible convenience methods include:

```ts
const allowed = await cedarling.can(request);

// or
await cedarling.assertAllowed(request);

// or
const decision = await cedarling.decide(request);
```

The expected behavior is:

- `authorize()` returns the complete authorization decision and diagnostics
- `can()` returns only a boolean
- `assertAllowed()` throws a typed access-denied error when access is denied

The final public API must be reviewed before being declared stable.

## API principles

- Applications use the same primary import in supported browser and server environments
- Consumers do not initialize the WASM module directly
- Consumers do not manually serialize authorization requests
- Consumers do not call generated snake_case binding methods
- Public methods and properties follow JavaScript naming conventions
- Public inputs and results are strongly typed
- Authorization results are returned as plain JavaScript objects
- An authorization denial is a normal result, not an engine exception
- Invalid input, invalid configuration and engine failures produce typed errors
- The underlying WASM package remains available as a low-level implementation package
- Runtime-specific behavior is handled internally by adapters
- Public API compatibility follows documented semantic-versioning rules

## Runtime support

The initial package should provide a consistent public API for:

- Modern browsers
- Supported server-side JavaScript environments, initially Node.js

Dahkenangnon:
### Prepare

-------------------

#### Target issue

Closes #14582

#### Implementation Details

This PR introduces `@janssenproject/cedarling`, a JavaScript and TypeScript SDK wrapping the generated `@janssenproject/cedarling_wasm` package.

Key changes:

- Adds a typed, runtime-neutral API for authorization, context management, issuer inspection, decision logs, and client lifecycle management.
- Supports Node.js, browsers, Bun, Deno, Electron, Cloudflare Workers/workerd, and edge runtimes through runtime-specific package entry points.
- Adds validation and contract tests that fail early when critical upstream Cedarling WASM exports or response shapes change.
- Adds unit, contract, browser, packed-consumer, and end-to-end tests across the supported runtimes.
- Adds runnable React/Node.js, Hono, Electron, and Next.js examples.
- Excludes `bindings/cedarling_js` from the parent Rust workspace because it is an npm project built on top of the WASM binding.
- Integrates the SDK into the existing Cedarling workflows:
  - `test-cedarling.yml` builds the WASM package locally and runs the complete JavaScript SDK check suite.
  - `build-packages.yml` shares the generated WASM artifact, builds and validates the SDK package, coordinates release versions, and publishes it to npm using the existing wasm pkg pattern.
- Adds consumer documentation and a focused maintainer guide covering the SDK boundary, project structure, supported runtimes, testing strategy, build order, and CI release lifecycle.

No new workflow files or npm publishing credentials were introduced.

-------------------

- [x] Relevant documentation has been updated

Validation included:

- `npm ci`
- `npm run check`
- TypeScript type checking
- Unit and WASM contract tests
- Supported-runtime test suites
- Browser and packed-consumer tests
- Package build and dry-run publication checks
- GitHub Actions workflow syntax validation

Documentation is included in the SDK root README and `docs/README.md`.

_Not applicable: this PR has documentation impact, and the relevant documentation has been updated._


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/d2b15579-b997-4266-a4d4-deff6270bc59?refs=59c23e6bec816f333571eec3a7d95f8e4f9ec7d3&source=prChecksComment) |

olehbozhok:
coderabbit review

duttarnab (commented):
1. There should be a section for build from source under installation:
https://github.com/JanssenProject/jans/blob/jans-cedarling-14582/jans-cedarling/bindings/cedarling_js/README.md#installation

2. Why this extra `dangerouslyDisableSignatureValidation ` is required we already have bootstrap properties to control validation checks? Also `logging.type ` is not required.
3. How addition `principal` and `resource` attributes will be added .
https://github.com/JanssenProject/jans/blob/jans-cedarling-14582/jans-cedarling/bindings/cedarling_js/README.md#clientauthorizeunsignedrequest

4. How to access complete authz-result json containing `diagnostics ` , `reason `, `errors `, 'request_id'.
5. Describe about the test cases, how to run test cases for unsigned and multi-issuer. 
6. Regarding initialization, define how the other bootstrap properties will be defined in args for initialization.

Dahkenangnon:
Thanks for the review. I’ll update the README to address all five points and clarify how the JavaScript SDK maps to Cedarling’s existing bootstrap configuration.

1. **Build from source:** Acknowledged. I’ll add a “Build from source” subsection under Installation, including prerequisites, building `cedarling_wasm` first, installing dependencies, building the SDK, and running the complete validation suite.

2. **Validation and logging configuration:** `jwt.dangerouslyDisableSignatureValidation` is the typed JavaScript mapping for the existing `CEDARLING_JWT_SIG_VALIDATION` bootstrap property; it does not introduce another validation mechanism. Signature validation is enabled by default, so I’ll remove the redundant `false` value from the introductory example.

   Similarly, `logging.type` maps to `CEDARLING_LOG_TYPE`. The current mappings are `off` → `off`, `memory` → `memory`, and `console` → `std_out`; Lock integration is configured separately through the SDK’s `lock` options. Cedarling’s four underlying logging modes are documented in the [[Cedarling logging reference](https://docs.jans.io/nightly/cedarling/reference/cedarling-logs/#cedarling-audit-logs)](https://docs.jans.io/nightly/cedarling/reference/cedarling-logs/#cedarling-audit-logs). Logging is optional, so I’ll move it out of the minimal initialization example and explain that memory logging is required only when using `client.logs`.

3. **Principal and resource attributes:** Both principal and resource entities accept an `attributes` object:

   ```ts
   principal: {
     type: "TaskApp::User",
     id: "alice",
     attributes: {
       role: "editor",
       department: "engineering",
     },
   },
   resource: {
     type: "TaskApp::Task",
     id: "task_101",
     attributes: {
       ownerId: "alice",
       status: "open",
     },
   },
   ```

   I’ll add this example and explain that these attributes must conform to the policy-store schema. In multi-issuer authorization, principal attributes are derived from mapped token claims, while resource attributes are supplied in the request.

4. **Complete authorization result:** The complete normalized result is available through `authResult.value`:

   ```ts
   if (authResult.ok) {
     const {
       decision,
       requestId,
       diagnostics: { reasons, errors },
     } = authResult.value;
   }
   ```

   The SDK uses JavaScript naming conventions: `request_id` becomes `requestId`, `reason` becomes `reasons`, and diagnostic `{ id, error }` entries become `{ policyId, message }`. I’ll document this result shape and show how to serialize it with `JSON.stringify(authResult.value)`. The raw WASM response remains internal to preserve the SDK’s stable API boundary.

5. **Unsigned and multi-issuer tests:** I’ll describe what the unit and real-WASM contract tests cover and add the focused commands:

   ```bash
   npm run test:unit -- authorize-unsigned
   npm run test:contract -- authorize-unsigned

   npm run test:unit -- authorize-multi-issuer
   npm run test:contract -- authorize-multi-issuer

   npm run check
   ```

   I’ll also clarify that the current multi-issuer contract tests use signed synthetic JWTs but disable external signature and status validation for deterministic offline execution. If end-to-end JWKS signature verification is required, that should be covered by an additional dedicated test.


6. **Initialization and remaining bootstrap properties:** Acknowledged. I’ll add a configuration reference showing how the supported bootstrap properties are expressed through `createCedarling()` arguments, including their defaults and mappings. This will cover `applicationName`, `policyStore`, `logging`, `authorization`, `jwt`, `contextStore`, `tokenCache`, `issuerLoading`, `http`, and `lock`.

   For example:

   - `applicationName` → `CEDARLING_APPLICATION_NAME`
   - `logging.*` → `CEDARLING_LOG_*`
   - `jwt.*` → `CEDARLING_JWT_*`
   - `authorization.*` → schema-validation and decision-log properties
   - `contextStore.*` → `CEDARLING_DATA_STORE_*`
   - `tokenCache.*` → `CEDARLING_TOKEN_CACHE_*`
   - `issuerLoading.*` → `CEDARLING_TRUSTED_ISSUER_LOADER_*`
   - `http.*` → `CEDARLING_HTTP_*`
   - `lock.*` → `CEDARLING_LOCK_*`

   The SDK intentionally uses typed JavaScript options instead of accepting unrestricted raw bootstrap objects. I’ll also identify any bootstrap properties that are unsupported or not applicable to WebAssembly runtimes, so consumers know exactly what can be configured during initialization.

Dahkenangnon:
These above will be implement with the following rules:

- if really consumer facing, into the cedarling/README.md
- if maintainer or developer facing, will go into cedarling_js/docs/README.md
Because the reason of the existence of this sdk is to hide the internal dance of the wasm module while offering javascript native interfaces to the js ecosystem dev.

EX: Building from source will be in docs/, while the fix related to the `jwt.dangerouslyDisableSignatureValidation` and complete authz results in the root readme.


Does it sounds correct ?

faisalsiddique4400 (commented):
## Blocking

### Signature validation is untested, and the shipped example disables it

Every contract test that runs real WASM authorization disables it:

- [`authorization.test.ts:107`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/tests/contract/authorization.test.ts#L107)
- [`authorize-multi-issuer.test.ts:58`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/tests/contract/authorize-multi-issuer.test.ts#L58)
- [`issuers.test.ts:15`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/tests/contract/issuers.test.ts#L15), [`:50`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/tests/contract/issuers.test.ts#L50), [`:85`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/tests/contract/issuers.test.ts#L85)

The rationale (deterministic offline runs) is already acknowledged in the thread. The gap that remains: if a change stopped verifying signatures, no test in this suite would fail.

The demo IdP at [`common/idp.js:115`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/demos/cedarling-js-examples/common/idp.js#L115) already generates an RSA keypair and serves JWKS, so a valid-signature case and a tampered-signature case can both run offline.

The same defect appears at runtime: [`common/test-config.json:10-11`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/demos/cedarling-js-examples/common/test-config.json#L10-L11) sets `dangerouslyDisableSignatureValidation` and `dangerouslyDisableStatusValidation` for every example, so the demos' "signed" path accepts forged tokens. Tests don't verify, and the shipped example turns verification off — those should be fixed together.

### The publish job cannot authenticate

`build_cedarling_js` runs `npm publish` at [`build-packages.yml:750`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L750) with no `NODE_AUTH_TOKEN`, no `.npmrc`, and no `registry-url:` on [`setup-node` at :622](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L622). The sibling `build_cedarling_wasm` job sets it at [:445](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L445).

The job therefore depends entirely on npm OIDC trusted publishing, hence `npm@11.9.0` at [:629](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L629). Trusted publishing must be configured against a package that already exists on npm; `@janssenproject/cedarling` does not. The first release will fail until someone publishes manually and configures the trusted publisher for this exact workflow and job name.

The PR description states the SDK is published "using the existing npm publishing secret." No secret is referenced in the job — either the description or the implementation needs correcting.

### `npm ci` runs against a swapped-in `file:` dependency

[`build-packages.yml:638`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L638) runs `npm version "${WASM_VERSION}"` after the downloaded artifact has replaced `cedarling_wasm/pkg`, then [:687](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L687) runs `npm ci` against the committed lockfile. `dependencies` pins `file:../cedarling_wasm/pkg`, and npm validates linked-package versions against the lock, so a version-bumped artifact is likely to produce "package.json and package-lock.json are not in sync."

This passes in `test-cedarling.yml` only because `wasm-pack` there builds the unbumped `Cargo.toml` version. The `workflow_dispatch` dry run noted under Release path would confirm or rule this out.

### Three call sites reject where their siblings do not

These re-throw when the caught error is not a recognized SDK code:

- [`client.ts:460`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/client/client.ts#L460) in `#authorize`
- [`client.ts:583`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/client/client.ts#L583) in `#finishClose`
- [`client.ts:679`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/client/client.ts#L679) in `createCedarling`

The issuer, context, and log helpers handle the same case by wrapping. See [`client.ts:302-315`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/client/client.ts#L302-L315), which uses `isSdkErrorCode(...) ? error : createSdkError(...)`.

The doc comment indicates the throw is intentional for unexpected failures. The issue is the asymmetry: `authorize`, `close`, and `createCedarling` can reject while structurally identical methods never do, and the API shape leads callers to check `result.ok` and stop there. That produces an unhandled rejection in production.

[`README.md:366`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/README.md#L366) states the SDK "does not throw errors on expected operational failures". Accurate as written, but "expected" is load-bearing and easy to miss. Either wrap consistently or document the rejection path.

### `authz-middleware.js` — unguarded dereference and an ownership fallback

[`authz-middleware.js:88`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/demos/cedarling-js-examples/react-nodejs/backend/cedarling/authz-middleware.js#L88) reads `diagnostics.reasons` unguarded, while [`:72`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/demos/cedarling-js-examples/react-nodejs/backend/cedarling/authz-middleware.js#L72) correctly guards the same value with `if (!decision && diagnostics)`. A deny without diagnostics throws inside the error path.

[`:13`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/demos/cedarling-js-examples/react-nodejs/backend/cedarling/authz-middleware.js#L13) sets `taskOwner = task ? task.owner : userId`, so a missing resource defaults ownership to the caller and any owner-based policy allows. With `x-user-id` defaulting to `'bob'` at [`:10`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/demos/cedarling-js-examples/react-nodejs/backend/cedarling/authz-middleware.js#L10), and signature validation off per the section above, this is the file consumers are most likely to copy. It needs an explicit "intentionally insecure, do not copy" banner at minimum.

## Non-blocking

The guard at [`assert-publishable.mjs:28`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/scripts/assert-publishable.mjs#L28) checks `name === "cedarling_wasm"`, but the dependency is `@janssenproject/cedarling_wasm`, so the branch never executes. Returning `false` would also mark it unsafe, which inverts the apparent intent. Staging rewrites the version first, so nothing breaks today.

With `allowedAlgorithms` omitted, all twelve algorithms are enabled: [`prepare.ts:24-36`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/configuration/prepare.ts#L24-L36), applied at [`prepare.ts:503`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/configuration/prepare.ts#L503). `none` is absent, so this is not urgent. Defaulting to symmetric HS\* alongside RS/ES/PS is the standard setup for algorithm confusion. Asymmetric by default, with HS\* opt-in, would be safer.

[`prepare.ts:21`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/configuration/prepare.ts#L21) defines `UINT32_MAX` as "the largest integer representable by the generated unsigned 32-bit fields", but five fields bound to `Number.MAX_SAFE_INTEGER` instead — `CEDARLING_DATA_STORE_MAX_TTL` and `_DEFAULT_TTL` at [:437-451](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/configuration/prepare.ts#L437-L451), `CEDARLING_HTTP_MAX_RESPONSE_SIZE_BYTES` at [:618](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/configuration/prepare.ts#L618), and the three `CEDARLING_LOCK_*_INTERVAL` values. Validation accepts values the WASM boundary will reject or wrap.

`httpUrl()` at [`prepare.ts:177`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/configuration/prepare.ts#L177) rejects credentials in the URL but accepts plaintext `http:`. It gates `policyStore.url` and `lock.configurationUrl`, the latter alongside `CEDARLING_LOCK_SSA_JWT`. Requiring `https:` outside loopback would match the algorithm-default reasoning above.

`withAuthorizationShortcuts` at [`client.ts:80`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/client/client.ts#L80) sets `denied: false` on the error branch, so `if (result.denied) reject()` fails open while `if (!result.allowed)` is safe. Since the shortcuts exist to encourage terse call sites, that asymmetry belongs in the type's doc comment.

`selectAuthorizationRequest` in [`authorization/dispatch.ts`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/authorization/dispatch.ts) reads `type` and `request` and silently ignores unknown envelope fields, while every other boundary in the SDK calls `rejectUnknown`.

SDK and WASM versions are permanently coupled: [`build-packages.yml:680`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L680) runs `npm version "${WASM_VERSION}"` and [`stage-release.mjs:110`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/scripts/stage-release.mjs#L110) pins exactly. A WASM patch therefore forces an SDK republish, and the published `cedarling_wasm` version stops matching the Rust crate. Confirming this is the intended long-term behaviour would be useful.


[`query.ts:94`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/logs/query.ts#L94) is unreachable. Given the `present === 0` check above it and the `id !== undefined` early return, `requestId` and `tag` cannot both be undefined at that point.

[`stage-release.mjs:120`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/scripts/stage-release.mjs#L120) spreads `...sourceManifest`, so all 10 `devDependencies` (electron, playwright, webpack, workerd) and 24 `test:*` scripts land in the published `package.json`. Consumers do not install devDeps, so the effect is cosmetic.

`test-config.json` declares [`expectedFailures: ["authorizeMultiIssuer"]`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/demos/cedarling-js-examples/common/test-config.json#L77-L79) described as a "known bug", echoed by a `console.log` in the demo's `init.js`. A linked tracking issue would be better than prose in a fixture.

The new [`cedarling_js_tests` job](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/test-cedarling.yml#L245) builds the WASM crate and installs Bun, Deno, and three Playwright browsers to run eight runtime suites, with no `timeout-minutes` and no `persist-credentials: false` on checkout. The latter is set on the new `build-packages.yml` job.

## Release path

[`build-packages.yml`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml) triggers on `workflow_run` and `workflow_dispatch` only, never on `pull_request`. The 257 lines of publish logic added here (version coordination, manifest staging, tarball verification, `npm publish`) have not executed. The green checks on this PR come from `test-cedarling.yml`. A `workflow_dispatch` dry run against nightly would validate it — and would settle the `npm ci` and npm-authentication questions above — before anything depends on it.

Note that the new SDK publish at [`build-packages.yml:750`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L750) fails loudly, whereas the existing WASM publish at [line 501](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L501) continues on error. Since `build_cedarling_js` depends on that job and then polls `npm view` for five minutes, a silently failed WASM publish surfaces here as a confusing timeout.

Minor: the new [`collect-cedarling-js-digests` tag step at :1977](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L1977) interpolates `inputs.target_tag` and `workflow_run.head_branch` directly into `run:`. This matches the existing pattern used by the other `collect-*` jobs in the file, so it is not a regression, but `build_cedarling_js` itself uses the safer `env:` form and the new job could match it.

CodeRabbit skipped this PR ("249 files exceed the limit of 150"). With 29,181 lines of committed lockfiles and an Electron demo carrying its own webpack configs, the PR exceeds the bots' limits. Splitting the SDK from the demos would restore that coverage.

## Checked, no findings

Publish security — injection and secret handling: actions are SHA-pinned, `harden-runner` present, `persist-credentials: false`, `github.repository` fork guard, `--provenance` with OIDC, SLSA provenance job, WASM package name and version asserted before publish. No injection or token-leak paths found. (Authentication is covered separately under Blocking.)

No committed secrets. The demo IdP generates its keypair at runtime with `generateKeyPairSync`. The Electron demo sets `contextIsolation: true`, `sandbox: true`, `nodeIntegration: false`, and routes external URLs through `shell.openExternal`.

Error model: [`errors.ts:229`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/errors/errors.ts#L229) produces frozen `Error` objects, strips URL credentials, and gates retained causes through a WeakSet allowlist. One trade-off: every message becomes a fixed generic string, so the underlying Cedar and WASM detail is discarded. This may make production debugging harder than expected; a debug flag preserving detail would address it.

Input validation is the strongest part of the SDK: `inspectOwnProperty` avoids invoking getters, `rejectUnknown` catches misspelled options deterministically, and values are snapshotted before crossing the WASM boundary. Lifecycle handling — in-flight accounting, a single memoized close promise, drain-before-shutdown — is careful. [`scripts/verify-consumer.mjs`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/scripts/verify-consumer.mjs) packs, installs offline into a scratch consumer, asserts the WASM dependency hoists rather than nests, and exercises both ESM and CJS entry points.

Archive bytes are copied during validation at [`prepare.ts:302`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/configuration/prepare.ts#L302), so the factory does not need to copy again. The SDK lockfile is committed so `npm ci` resolves in CI. The Cargo workspace exclusion is harmless, though `bindings/cedarling_js` contains no `Cargo.toml`, so the entry is a no-op.

jans-cedarling/bindings/cedarling_js/README.md
```diff
@@ -0,0 +1,398 @@
...
+---
+
+## Initialization
+
+Initialize a Cedarling client using `createCedarling(config)`. This returns a `Result` containing the active `Cedarling` instance.
+
+```ts
+import { createCedarling } from "@janssenproject/cedarling";
+
+const result = await createCedarling({
```
olehbozhok, line 90: I think it would be better to avoid custom mapping to Bootstrap properties. Any changes in the core library would require corresponding updates in this library, which increases the maintenance burden.

I also like that every binding follows the same approach, with the official documentation serving as the single source of truth. Thanks for all the work you've put into this—it's a significant effort. That said, I think passing the values directly would be a simpler and more maintainable solution.

Dahkenangnon: Thanks. I agree that direct pass-through would create a thinner binding, but [`cedarling_wasm` already provides that low-level API](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_wasm/src/lib.rs#L69-L82).

 The goal of `@janssenproject/cedarling` is a stable, Web-native SDK. Its typed [`CedarlingOptions`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/configuration/types.ts#L301-L339) also supports JavaScript-specific sources such as [`Uint8Array` archives and async loaders](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/configuration/types.ts#L41-L71), which require routing through [`init_from_archive_bytes`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_wasm/src/lib.rs#L84-L103).

 Passing `CEDARLING_*` values directly would expose the generated/core contract to every consumer and duplicate the existing WASM API. The [official bootstrap documentation](https://docs.jans.io/nightly/cedarling/reference/cedarling-properties/) should remain the source of truth for Cedarling semantics, while this SDK maintains a curated, versioned JS interface with compatibility tests.

 I therefore propose keeping the typed mapping. Consumers needing raw bootstrap access can use `cedarling_wasm`, or we can discuss a separate explicitly low-level API.

Dahkenangnon:
> ## Blocking
> ### Signature validation is untested, and the shipped example disables it
> Every contract test that runs real WASM authorization disables it:
> 
> * [`authorization.test.ts:107`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/tests/contract/authorization.test.ts#L107)
> * [`authorize-multi-issuer.test.ts:58`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/tests/contract/authorize-multi-issuer.test.ts#L58)
> * [`issuers.test.ts:15`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/tests/contract/issuers.test.ts#L15), [`:50`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/tests/contract/issuers.test.ts#L50), [`:85`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/tests/contract/issuers.test.ts#L85)
> 
> The rationale (deterministic offline runs) is already acknowledged in the thread. The gap that remains: if a change stopped verifying signatures, no test in this suite would fail.
> 
> The demo IdP at [`common/idp.js:115`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/demos/cedarling-js-examples/common/idp.js#L115) already generates an RSA keypair and serves JWKS, so a valid-signature case and a tampered-signature case can both run offline.
> 
> The same defect appears at runtime: [`common/test-config.json:10-11`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/demos/cedarling-js-examples/common/test-config.json#L10-L11) sets `dangerouslyDisableSignatureValidation` and `dangerouslyDisableStatusValidation` for every example, so the demos' "signed" path accepts forged tokens. Tests don't verify, and the shipped example turns verification off — those should be fixed together.
> 
> ### The publish job cannot authenticate
> `build_cedarling_js` runs `npm publish` at [`build-packages.yml:750`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L750) with no `NODE_AUTH_TOKEN`, no `.npmrc`, and no `registry-url:` on [`setup-node` at :622](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L622). The sibling `build_cedarling_wasm` job sets it at [:445](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L445).
> 
> The job therefore depends entirely on npm OIDC trusted publishing, hence `npm@11.9.0` at [:629](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L629). Trusted publishing must be configured against a package that already exists on npm; `@janssenproject/cedarling` does not. The first release will fail until someone publishes manually and configures the trusted publisher for this exact workflow and job name.
> 
> The PR description states the SDK is published "using the existing npm publishing secret." No secret is referenced in the job — either the description or the implementation needs correcting.
> 
> ### `npm ci` runs against a swapped-in `file:` dependency
> [`build-packages.yml:638`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L638) runs `npm version "${WASM_VERSION}"` after the downloaded artifact has replaced `cedarling_wasm/pkg`, then [:687](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L687) runs `npm ci` against the committed lockfile. `dependencies` pins `file:../cedarling_wasm/pkg`, and npm validates linked-package versions against the lock, so a version-bumped artifact is likely to produce "package.json and package-lock.json are not in sync."
> 
> This passes in `test-cedarling.yml` only because `wasm-pack` there builds the unbumped `Cargo.toml` version. The `workflow_dispatch` dry run noted under Release path would confirm or rule this out.
> 
> ### Three call sites reject where their siblings do not
> These re-throw when the caught error is not a recognized SDK code:
> 
> * [`client.ts:460`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/client/client.ts#L460) in `#authorize`
> * [`client.ts:583`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/client/client.ts#L583) in `#finishClose`
> * [`client.ts:679`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/client/client.ts#L679) in `createCedarling`
> 
> The issuer, context, and log helpers handle the same case by wrapping. See [`client.ts:302-315`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/client/client.ts#L302-L315), which uses `isSdkErrorCode(...) ? error : createSdkError(...)`.
> 
> The doc comment indicates the throw is intentional for unexpected failures. The issue is the asymmetry: `authorize`, `close`, and `createCedarling` can reject while structurally identical methods never do, and the API shape leads callers to check `result.ok` and stop there. That produces an unhandled rejection in production.
> 
> [`README.md:366`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/README.md#L366) states the SDK "does not throw errors on expected operational failures". Accurate as written, but "expected" is load-bearing and easy to miss. Either wrap consistently or document the rejection path.
> 
> ### `authz-middleware.js` — unguarded dereference and an ownership fallback
> [`authz-middleware.js:88`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/demos/cedarling-js-examples/react-nodejs/backend/cedarling/authz-middleware.js#L88) reads `diagnostics.reasons` unguarded, while [`:72`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/demos/cedarling-js-examples/react-nodejs/backend/cedarling/authz-middleware.js#L72) correctly guards the same value with `if (!decision && diagnostics)`. A deny without diagnostics throws inside the error path.
> 
> [`:13`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/demos/cedarling-js-examples/react-nodejs/backend/cedarling/authz-middleware.js#L13) sets `taskOwner = task ? task.owner : userId`, so a missing resource defaults ownership to the caller and any owner-based policy allows. With `x-user-id` defaulting to `'bob'` at [`:10`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/demos/cedarling-js-examples/react-nodejs/backend/cedarling/authz-middleware.js#L10), and signature validation off per the section above, this is the file consumers are most likely to copy. It needs an explicit "intentionally insecure, do not copy" banner at minimum.
> 
> ## Non-blocking
> The guard at [`assert-publishable.mjs:28`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/scripts/assert-publishable.mjs#L28) checks `name === "cedarling_wasm"`, but the dependency is `@janssenproject/cedarling_wasm`, so the branch never executes. Returning `false` would also mark it unsafe, which inverts the apparent intent. Staging rewrites the version first, so nothing breaks today.
> 
> With `allowedAlgorithms` omitted, all twelve algorithms are enabled: [`prepare.ts:24-36`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/configuration/prepare.ts#L24-L36), applied at [`prepare.ts:503`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/configuration/prepare.ts#L503). `none` is absent, so this is not urgent. Defaulting to symmetric HS* alongside RS/ES/PS is the standard setup for algorithm confusion. Asymmetric by default, with HS* opt-in, would be safer.
> 
> [`prepare.ts:21`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/configuration/prepare.ts#L21) defines `UINT32_MAX` as "the largest integer representable by the generated unsigned 32-bit fields", but five fields bound to `Number.MAX_SAFE_INTEGER` instead — `CEDARLING_DATA_STORE_MAX_TTL` and `_DEFAULT_TTL` at [:437-451](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/configuration/prepare.ts#L437-L451), `CEDARLING_HTTP_MAX_RESPONSE_SIZE_BYTES` at [:618](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/configuration/prepare.ts#L618), and the three `CEDARLING_LOCK_*_INTERVAL` values. Validation accepts values the WASM boundary will reject or wrap.
> 
> `httpUrl()` at [`prepare.ts:177`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/configuration/prepare.ts#L177) rejects credentials in the URL but accepts plaintext `http:`. It gates `policyStore.url` and `lock.configurationUrl`, the latter alongside `CEDARLING_LOCK_SSA_JWT`. Requiring `https:` outside loopback would match the algorithm-default reasoning above.
> 
> `withAuthorizationShortcuts` at [`client.ts:80`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/client/client.ts#L80) sets `denied: false` on the error branch, so `if (result.denied) reject()` fails open while `if (!result.allowed)` is safe. Since the shortcuts exist to encourage terse call sites, that asymmetry belongs in the type's doc comment.
> 
> `selectAuthorizationRequest` in [`authorization/dispatch.ts`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/authorization/dispatch.ts) reads `type` and `request` and silently ignores unknown envelope fields, while every other boundary in the SDK calls `rejectUnknown`.
> 
> SDK and WASM versions are permanently coupled: [`build-packages.yml:680`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L680) runs `npm version "${WASM_VERSION}"` and [`stage-release.mjs:110`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/scripts/stage-release.mjs#L110) pins exactly. A WASM patch therefore forces an SDK republish, and the published `cedarling_wasm` version stops matching the Rust crate. Confirming this is the intended long-term behaviour would be useful.
> 
> [`query.ts:94`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/logs/query.ts#L94) is unreachable. Given the `present === 0` check above it and the `id !== undefined` early return, `requestId` and `tag` cannot both be undefined at that point.
> 
> [`stage-release.mjs:120`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/scripts/stage-release.mjs#L120) spreads `...sourceManifest`, so all 10 `devDependencies` (electron, playwright, webpack, workerd) and 24 `test:*` scripts land in the published `package.json`. Consumers do not install devDeps, so the effect is cosmetic.
> 
> `test-config.json` declares [`expectedFailures: ["authorizeMultiIssuer"]`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/demos/cedarling-js-examples/common/test-config.json#L77-L79) described as a "known bug", echoed by a `console.log` in the demo's `init.js`. A linked tracking issue would be better than prose in a fixture.
> 
> The new [`cedarling_js_tests` job](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/test-cedarling.yml#L245) builds the WASM crate and installs Bun, Deno, and three Playwright browsers to run eight runtime suites, with no `timeout-minutes` and no `persist-credentials: false` on checkout. The latter is set on the new `build-packages.yml` job.
> 
> ## Release path
> [`build-packages.yml`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml) triggers on `workflow_run` and `workflow_dispatch` only, never on `pull_request`. The 257 lines of publish logic added here (version coordination, manifest staging, tarball verification, `npm publish`) have not executed. The green checks on this PR come from `test-cedarling.yml`. A `workflow_dispatch` dry run against nightly would validate it — and would settle the `npm ci` and npm-authentication questions above — before anything depends on it.
> 
> Note that the new SDK publish at [`build-packages.yml:750`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L750) fails loudly, whereas the existing WASM publish at [line 501](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L501) continues on error. Since `build_cedarling_js` depends on that job and then polls `npm view` for five minutes, a silently failed WASM publish surfaces here as a confusing timeout.
> 
> Minor: the new [`collect-cedarling-js-digests` tag step at :1977](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/.github/workflows/build-packages.yml#L1977) interpolates `inputs.target_tag` and `workflow_run.head_branch` directly into `run:`. This matches the existing pattern used by the other `collect-*` jobs in the file, so it is not a regression, but `build_cedarling_js` itself uses the safer `env:` form and the new job could match it.
> 
> CodeRabbit skipped this PR ("249 files exceed the limit of 150"). With 29,181 lines of committed lockfiles and an Electron demo carrying its own webpack configs, the PR exceeds the bots' limits. Splitting the SDK from the demos would restore that coverage.
> 
> ## Checked, no findings
> Publish security — injection and secret handling: actions are SHA-pinned, `harden-runner` present, `persist-credentials: false`, `github.repository` fork guard, `--provenance` with OIDC, SLSA provenance job, WASM package name and version asserted before publish. No injection or token-leak paths found. (Authentication is covered separately under Blocking.)
> 
> No committed secrets. The demo IdP generates its keypair at runtime with `generateKeyPairSync`. The Electron demo sets `contextIsolation: true`, `sandbox: true`, `nodeIntegration: false`, and routes external URLs through `shell.openExternal`.
> 
> Error model: [`errors.ts:229`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/errors/errors.ts#L229) produces frozen `Error` objects, strips URL credentials, and gates retained causes through a WeakSet allowlist. One trade-off: every message becomes a fixed generic string, so the underlying Cedar and WASM detail is discarded. This may make production debugging harder than expected; a debug flag preserving detail would address it.
> 
> Input validation is the strongest part of the SDK: `inspectOwnProperty` avoids invoking getters, `rejectUnknown` catches misspelled options deterministically, and values are snapshotted before crossing the WASM boundary. Lifecycle handling — in-flight accounting, a single memoized close promise, drain-before-shutdown — is careful. [`scripts/verify-consumer.mjs`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/scripts/verify-consumer.mjs) packs, installs offline into a scratch consumer, asserts the WASM dependency hoists rather than nests, and exercises both ESM and CJS entry points.
> 
> Archive bytes are copied during validation at [`prepare.ts:302`](https://github.com/JanssenProject/jans/blob/4ac3926c1a6aeb04fdea6ac6061aac963b52226b/jans-cedarling/bindings/cedarling_js/src/configuration/prepare.ts#L302), so the factory does not need to copy again. The SDK lockfile is committed so `npm ci` resolves in CI. The Cargo workspace exclusion is harmless, though `bindings/cedarling_js` contains no `Cargo.toml`, so the entry is a no-op.

Ack and working on these in child PR.

jans-cedarling/bindings/cedarling_js/src/client/types.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  /**
+   * Dispatches a discriminated authorization union to one named operation.
+   *
+   * @param request - Unsigned or multi-issuer authorization envelope.
+   * @returns The selected named operation's decision with dispatcher errors.
+   */
+  authorize(
+    request: AuthorizationRequest,
+  ): Promise<AuthorizationResult<CedarlingAuthorizationError>>;
```
olehbozhok: Instead of having `authorizeUnsigned` and `authorizeMultiIssuer` we have 3 other method that do the same, do we really need that? KISS & YAGNI

jans-cedarling/bindings/cedarling_js/src/client/types.ts
```diff
@@ -0,0 +1,135 @@
...
+   *
+   * @returns Success after shutdown, or a normalized lifecycle error.
+   *
+   * @example
+   * ```ts
+   * const closed = await client.close();
+   * if (!closed.ok) console.error(closed.error.code);
+   * ```
+   */
+  close(): Promise<Result<void, CedarlingLifecycleError>>;
```
olehbozhok: In WASM we have method `shut_down` I think better to have the same naming. To be consistent over different bindings.

jans-cedarling/bindings/cedarling_js/README.md
olehbozhok: OK, maybe at least we should have a method to build a config based on raw bootstrap properties; maybe for someone it will be useful.

jans-cedarling/bindings/cedarling_js/src/client/types.ts
olehbozhok: This method waits until all workers are finished

jans-cedarling/bindings/cedarling_js/README.md
```diff
@@ -0,0 +1,398 @@
...
+
+All authorization methods return a unified authorization result containing a `decision` (`true` for `Allow`, `false` for `Deny`), diagnostic reasoning, and an execution request ID.
+
+### Optional decision shortcuts
+
+The standard `ok`/`value`/`error` shape remains available and is recommended when you need complete explicit error handling. For concise authorization checks, all three authorization entry points also expose flat shortcuts:
+
+```ts
+const result = await client.authorizeUnsigned(request);
+const { ok, allowed, denied, err } = result;
```
olehbozhok: allowed and denied is the same bool value

jans-cedarling/bindings/cedarling_js/README.md
Dahkenangnon: Ok, thank . I think this is feasible.

jans-cedarling/bindings/cedarling_js/src/client/types.ts
Dahkenangnon: The three declarations are TypeScript overload signatures for one runtime dispatcher, but I agree that the dispatcher itself duplicates the two named authorization operations. Go and Python expose only unsigned and multi-issuer methods, and those names make the selected trust model explicit. I will remove the generic authorize() API and retain `authorizeUnsigned()` and `authorizeMultiIssuer()`.

jans-cedarling/bindings/cedarling_js/src/client/types.ts
Dahkenangnon: Given the explicit intention to have less difference on public api across bindings (python, go, js,etc) goal, renaming `close()` to `shutDown()` is reasonable and still idiomatic TypeScript.

So, I will rename it to `shutDown()` .

jans-cedarling/bindings/cedarling_js/README.md
Dahkenangnon: Agreed! I'll remove these flat

olehbozhok (dismissed):

moabu (dismissed):

olehbozhok:
This PR should be splited into a bunch of small PRs and merged step by step

PR #14636, opened 2026-07-27
feat(jans-cedarling): implement Sigstore/Cosign verification library  

### LINKED ISSUE #14465: feat(jans-cedarling): implement Sigstore/Cosign verification library
## Description

Implement a **standalone Sigstore/Cosign bundle verification library** for offline blob verification (equivalent to `cosign verify-blob`). The library must compile to both **native** (`x86_64`, `aarch64`) and **WASM** (`wasm32-unknown-unknown`) targets.

The verifier takes raw artifact bytes + a Sigstore bundle JSON and produces a verified identity, without any network calls during verification. All trust material (Fulcio roots, Rekor keys, CTFE keys) is provided at construction time.

## Motivation

- **Sigstore verification** is the de-facto standard for keyless software signing
- **Existing Rust library** (`sigstore-rs`) depends on `aws-lc-rs` (native C) and cannot compile to WASM
- **No WASM-compatible Sigstore verifier exists** in the Rust ecosystem
- A pure-Rust, WASM-compatible verifier unlocks browser, edge, and embedded use cases

## Key Design Constraint: WASM + Native

Any implementation **must work on `wasm32-unknown-unknown`**. This rules out `aws-lc-rs` (C library), `openssl`, `ring` (C + asm), `tough` (native TUF client), `webbrowser`, and `oci-client`.

### Dependency Compatibility Matrix

| Operation | sigstore-rs dep | WASM? | Pure Rust alternative |
|---|---|---|---|
| SHA-256 | `sha2` | ✓ | Same (already pure Rust) |
| ECDSA P-256 | `p256` + `ecdsa` | ✓ | Same (RustCrypto, pure Rust) |
| Cert chain validation | `rustls-webpki` + `aws-lc-rs` | ✗ | Custom path validation over `x509-parser` + `p256::ecdsa::VerifyingKey` |
| X.509 parsing | `x509-cert` | ✓ | `x509-parser` (zero-copy, pure Rust) |
| JSON canonicalization | `serde_json_canonicalizer` | ✓ | Same |
| Bundle JSON parsing | `serde_json` | ✓ | Same |
| TUF trust root fetch | `tough` (native) | ✗ | **Out of scope.** Caller provides `SigstoreTrustRootRaw` — embedded or fetched via own TUF client |
| SCT verification | via `aws-lc-rs` | ✗ | Extract SCT from x.509 extension, verify with `p256::ecdsa::VerifyingKey` |
| SET verification (Rekor) | TODO in sigstore-rs | ✓ | Pure Rust (`serde_json_canonicalizer` + SHA-256 + ECDSA) |
| OAuth/OIDC login | `webbrowser` + `openidconnect` | ✗ | Not needed (verify-only, no signing) |

### Strategy: Custom Implementation (not sigstore-rs fork)

Rather than forking `sigstore-rs` (pre-1.0, heavy dependency tree, ~50 deps), build a **minimal custom verifier**:

- **Only blob verification** (no OCI container/image signing, no `cosign sign`)
- **Offline-only** (SET-based offline verification per Sigstore spec, no Merkle inclusion proofs)
- **ECDSA P-256 only** — Fulcio root, leaf certs, Rekor key, CTFE keys all use ECDSA P-256 in production. RSA/Ed25519 added later if needed.
- **No network inside `verify()`** — trust root is caller's responsibility. No TUF client in this crate.

### sigstore-rs (`Verifier` — verifier.rs:127-223)

```rust
pub struct Verifier {
    rekor_config: RekorConfiguration,
    cert_pool: CertificatePool,          // Fulcio roots
    ctfe_keyring: Keyring,               // CTFE public keys
}

impl Verifier {
    pub fn new<R: TrustRoot>(rekor_config, trust_repo: R) -> Result<Self>;
    pub async fn verify_digest(&self, digest: Sha256, bundle: Bundle, policy: &P) -> VerificationResult;
    pub async fn verify<R: AsyncRead>(&self, input: R, bundle: Bundle, policy: &P) -> VerificationResult;
}
```

7-step flow: cert chain → SCT → OIDC identity → signature → Rekor consistency → SET → cert timestamp validity.
Performs offline SET-based verification; online Merkle inclusion is a separate code path.

**WASM blocker:** uses `aws-lc-rs` for cert chain and SCT, `tough` for TUF, `webbrowser` for OAuth, `oci-client` for containers. All native C.

### sigstore-js (`@sigstore/verify` — verifier.ts)

```typescript
class Verifier {
    constructor(trustMaterial: TrustMaterial, options?: VerifierOptions)
    verify(entity: SignedEntity, policy?: VerificationPolicy): Signer
}

// Entity = signature + key(cert|pubkey) + timestamps + tlogEntries
// TrustMaterial = certs + keyFinder + ctlogs + tlogs + timestampAuthorities
```

Modular package structure: `bundle/`, `key/`, `timestamp/`, `tlog/`, `trust/`. This modular decomposition is a good model for our crate.

### sigstore-go (`verifier.go`)

```go
type SignedEntity struct {
    Signature  SignatureContent
    SigningKey VerificationKey        // certificate or public key
    Timestamps []Timestamp
    TlogEntries []*TlogEntry
}

type TrustedMaterial struct {
    FulcioCertificateAuthorities []*x509.Certificate
    RekorLogs map[string]*TransparencyLog
    CTLogs    map[string]*TransparencyLog
}

func (v *SignedEntityVerifier) Verify(entity *SignedEntity, policy *Policy) (*VerificationResult, error)
```

## Proposed API

```rust
/// Raw PEM-encoded trust material. Caller provides these — no network in this crate.
///
/// Multiple entries per field support key rotation: verifier tries all roots
/// during chain building, accepts the one that validates the leaf cert.
/// Validity periods are extracted from the certs/keys themselves during parsing.
pub struct SigstoreTrustRootRaw {
    pub fulcio_root_certs: Vec<Vec<u8>>,           // PEM-encoded Fulcio root CAs
    pub fulcio_intermediate_certs: Vec<Vec<u8>>,   // PEM intermediates
    pub rekor_keys: Vec<Vec<u8>>,                  // PEM Rekor signing keys
    pub ctfe_keys: Vec<Vec<u8>>,                   // PEM CTFE public keys
}
```

```rust
/// Offline Sigstore blob verifier. No network calls during verify().
///
/// Hot-swap via `ArcSwap` at the caller level.
pub struct SigstoreBlobVerifier { .. }

impl SigstoreBlobVerifier {
    /// Construct from explicit trust root bytes.
    /// May fail if PEM/DER data is malformed.
    pub fn new(trust_root: SigstoreTrustRootRaw) -> Result<Self, SigstoreVerificationError>;

    /// Construct with public-good Sigstore keys embedded at compile time.
    /// Uses `include_bytes!` — zero network, zero fs.
    /// Panics only if embedded files are corrupted (compile-time bug, caught by tests).
    pub fn with_static_trust_root() -> Self;

    /// Verify that `artifact_bytes` was signed, producing `bundle_json`.
    pub fn verify(
        &self,
        artifact_bytes: &[u8],
        bundle_json: &[u8],
        policy: &VerificationPolicy,
    ) -> Result<VerifiedSignature, SigstoreVerificationError>;
}
```

```rust
/// Identity is MANDATORY. Nonexistent: you must always specify whom you trust.
pub struct VerificationPolicy {
    pub cert_identity: IdentityMatch,
    pub cert_issuer: String,   // exact match against 1.3.6.1.4.1.57264.1.8 extension
}

pub enum IdentityMatch {
    /// Exact SAN match.
    Exact(String),
    /// Regex against SAN. AUTO-anchored: internally wrapped to `\A(?:pattern)\z`
    /// to prevent partial-match attacks (e.g. `evil.com` won't match `not-evil.com.attacker.io`).
    Regex(String),
}
```

```rust
pub struct VerifiedSignature {
    pub subject_alternative_name: String,
    pub issuer: String,
    pub verified_at: i64,           // Rekor integratedTime (verified via SET)
}
```

### Error type

```rust
pub enum SigstoreVerificationError {
    BundleParsing { source: serde_json::Error },
    InvalidBundleFormat { reason: String },
    CertificateParsing { reason: String },
    CertificateChain { reason: String },
    SctVerification { reason: String },
    PolicyViolation { reason: String },
    SignatureMismatch { reason: String },
    RekorInconsistency { reason: String },
    SetVerification { reason: String },
    CertificateExpired { reason: String },
    UnsupportedAlgorithm { algorithm: String },
}
```

## Verification Algorithm (9 steps, offline)

**Critical ordering:** SET verification (step 3) authenticates `integratedTime` **before** it is used as a trusted timestamp in cert chain validation (step 4) and validity window check (step 6).

```
1. Parse bundle JSON → extract mediaType, cert, signature, tlog_entry, SET
2. Parse X.509 cert from bundle → extract pubkey, SAN, OIDC issuer ext, validity, SCT
3. SET verification → verify Rekor's signed entry timestamp (RFC 8785 canonicalization)
   This authenticates integratedTime — after this step, integratedTime is TRUSTED.
4. Cert chain validation → Fulcio root (timestamp-anchored on verified integratedTime)
   Must check: BasicConstraints CA, KeyUsage keyCertSign, EKU id-kp-codeSigning, pathLen.
5. SCT verification → verify against CTFE public key (key selected by integratedTime)
6. Cert validity window → not_before ≤ integratedTime ≤ not_after
7. OIDC identity check → SAN regex/Exact match + issuer exact match against policy
8. Artifact signature → SHA-256(artifact) verified against cert pubkey
9. Rekor entry consistency → cert + signature + artifact hash all match log body (CVE-2022-36056)
```

## Bundle Format Support

| mediaType | Support | Notes |
|---|---|---|
| `application/vnd.dev.sigstore.bundle.v0.3+json` | ✓ Required | Current bundle format |
| `MessageSignature` payload | ✓ Required | Simple blob signing (cosign sign-blob) |
| `DSSE` / in-toto payload | Optional (phase 2) | Attestations, not needed for initial release |

Following [Sigstore client spec §4.1.1](https://github.com/sigstore/architecture-docs/blob/main/client-spec.md#411-recommended-discovery-order-for-verification-materials), the *caller* is responsible for locating the bundle file (e.g., `{artifact}.sigstore.json`). The verifier only takes bytes.

## Dependencies

```toml
[dependencies]
# All WASM-compatible, pure Rust (RustCrypto)
sha2 = { version = "0.10", default-features = false }
p256 = { version = "0.13", default-features = false, features = ["ecdsa", "pem"] }
ecdsa = { version = "0.16", default-features = false, features = ["verifying"] }
x509-parser = { version = "0.17", default-features = false }
serde_json = { version = "1", default-features = false, features = ["std"] }
serde_json_canonicalizer = "0.3"    # PIN VERSION — mismatch breaks SET verification
der = { version = "0.7", default-features = false }
pem = { version = "3", default-features = false }
regex-lite = { version = "0.1", default-features = false }  # lightweight, no unicode bloat
thiserror = { version = "2", default-features = false }

# getrandom NOT needed — ECDSA verify does not use RNG.
# ring NOT needed — p256::ecdsa::VerifyingKey handles all ECDSA verification.

[dev-dependencies]
rcgen = { version = "0.14", default-features = false, features = ["pem", "p256"] }  # synthetic certs for unit tests
```

**No features needed.** Single `[dependencies]` block compiles identically for native and WASM.

`serde_json_canonicalizer` must be pinned to a specific version and tested against `cosign verify-blob --output-file` to ensure canonicalization byte-for-byte matches Rekor's RFC 8785 format.

## Acceptance Criteria

- [ ] `cosign sign-blob ./artifact --bundle artifact.sigstore.json` produces a verifiable bundle
- [ ] Verifier accepts valid bundle with identity matching policy
- [ ] Verifier rejects bundle signed by wrong identity (policy violation)
- [ ] Verifier rejects tampered artifact (signature mismatch)
- [ ] Verifier rejects certificate not chaining to Fulcio root (self-signed attack)
- [ ] Verifier rejects expired certificate (integratedTime outside validity window)
- [ ] Verifier rejects bundle where Rekor entry doesn't match cert/signature/hash (CVE-2022-36056)
- [ ] Verifier rejects bundle with invalid SET
- [ ] Verifier rejects bundle with unsupported algorithm
- [ ] `cargo build --target wasm32-unknown-unknown` succeeds
- [ ] `cargo build --target x86_64-unknown-linux-gnu` succeeds
- [ ] Unit test for each verification step (positive + negative)
- [ ] Integration test: `cosign sign-blob` → verify with `SigstoreBlobVerifier`
- [ ] Integration test: verify against `cosign` CLI output for cross-implementation parity
- [ ] Integration test: `serde_json_canonicalizer` output byte-for-byte identical to cosign's RFC 8785 canonicalization
- [ ] No network calls during `verify()` (offline verification)
- [ ] No `getrandom` / RNG dependency in dependency tree

### Phase 1: Core verifier (platform-agnostic)

- [ ] `bundle.rs`: bundle JSON deserialization types (`Bundle`, `TlogEntry`, `MediaType`, `MessageSignature`, `DsseEnvelope`)
- [ ] `crypto.rs`: ECDSA P-256 signature verification via `p256::ecdsa::VerifyingKey` (verify-only, no signing, no RNG)
- [ ] `cert.rs`: X.509 cert parsing from DER/PEM via `x509-parser`.
  - **Trust root certs** (at `new()` time): check `BasicConstraints CA:true`, `KeyUsage keyCertSign` — reject non-CA certs as invalid trust anchors
  - **Leaf cert** (from bundle, at `verify()` time): extract pubkey, SAN (URI + email), OIDC issuer extension (`1.3.6.1.4.1.57264.1.8`), validity (`not_before`, `not_after`), SCT extension
  - **Leaf cert checks:** `BasicConstraints CA:false` (must not be CA), `EKU id-kp-codeSigning` (1.3.6.1.5.5.7.3.3) — Fulcio issues only code-signing certs, reject if missing
  - **All certs:** validity check anchored on `integratedTime` (step 4/6 in algorithm, after SET verified)
- [ ] `chain.rs`: certificate chain validation (BasicConstraints/KeyUsage/EKU pre-validated by `cert.rs`):
  - Path building from leaf to trusted root
  - Timestamp anchoring using verified `integratedTime` (not current wall clock)
  - Check `pathLen` constraints on intermediate CAs
  - Each link: verify `issuer.sign(leaf_tbs)` via `p256::ecdsa::VerifyingKey`
  - Name constraints, policy constraints, CRL not required for Fulcio (short-lived certs, no revocation)
- [ ] `sct.rs`: SCT extraction from cert's x.509 extension and signature verification against CTFE keys (try all CTFE keys, accept any valid)
- [ ] `tlog.rs`: Rekor entry parsing from bundle, consistency check (cert + sig + hash match), SET verification — RFC 8785 canonicalization via `serde_json_canonicalizer` + ECDSA verify against Rekor key
- [ ] `verifier.rs`: `SigstoreBlobVerifier` orchestrating the 9-step flow with correct ordering (SET before cert chain)
- [ ] `policy.rs`: `VerificationPolicy` — identity matching with auto-anchored regex + exact issuer match. No `Option` — identity is mandatory.
- [ ] `error.rs`: `SigstoreVerificationError` enum with `Display` + `std::error::Error`
- [ ] `trust_root.rs`: `SigstoreTrustRootRaw` struct + PEM→DER conversion + `with_static_trust_root()`
- [ ] `trust/`: embedded static files — Fulcio root + intermediate (PEM), Rekor key (PEM), CTFE keys (PEM). Each with documented validity periods.
- [ ] `test_files/`: real `cosign sign-blob` bundles for integration tests (committed to repo, no cosign CLI at test time)
- [ ] `test_utils/`: rcgen-based synthetic cert generator — creates Fulcio-like CA + leaf certs with configurable OIDC issuer, SAN, SCT, validity for unit tests

### Phase 2: Tests

Two levels — synthetic (rcgen, offline, deterministic) and real bundles (cosign, integration).

#### Synthetic tests (rcgen — no network, pure Rust, WASM-safe)

Generate a test Fulcio-like CA + leaf cert via `rcgen` (pure Rust, WASM-compatible). Full control over cert fields — test every step in isolation with both positive and negative cases.

- [ ] **Cert chain:** valid chain (leaf → intermediate → root) passes; self-signed leaf rejected; chain with missing intermediate rejected; root not in trusted set rejected
- [ ] **BasicConstraints:** CA:true on intermediate/root accepted; CA:false on root rejected; leaf cert CA:false accepted; leaf cert CA:true rejected
- [ ] **EKU:** leaf cert with `id-kp-codeSigning` accepted; without it rejected
- [ ] **KeyUsage:** cert with `keyCertSign` accepted; without it rejected
- [ ] **Validity:** cert valid at `integratedTime` passes; cert expired before `integratedTime` rejected; cert not-yet-valid rejected
- [ ] **SCT:** valid SCT in cert extension passes; missing SCT rejected; SCT signed by wrong key rejected
- [ ] **SET:** Rekor SET over synthetic log entry passes; tampered SET rejected; SET signed by wrong key rejected
- [ ] **Rekor consistency:** entry body matches cert+sig+hash passes; body with wrong cert rejected; wrong hash rejected (CVE-2022-36056)
- [ ] **Policy matching:** exact SAN match passes; regex SAN with anchored pattern passes; `evil.com` does NOT match `not-evil.com.attacker.io`; wrong issuer rejected; empty issuer rejected
- [ ] **Signature:** valid ECDSA P-256 signature over SHA-256(artifact) passes; signature over wrong artifact rejected; zero-length signature rejected
- [ ] **with_static_trust_root():** never panics (embedded PEMs parse correctly)

#### Real bundle integration tests (cosign CLI)

Pre-generated bundles committed to `test_files/`. No cosign CLI needed at test time — tests read from disk.

- [ ] Verify `cosign sign-blob` bundle against public-good Sigstore with real cert chain
- [ ] Cross-implementation parity: `serde_json_canonicalizer` output byte-for-byte identical to cosign CLI RFC 8785 output
- [ ] Verify bundle signed with GitHub Actions OIDC
- [ ] Verify bundle signed with Google OIDC

#### CI

- [ ] `cargo build --target wasm32-unknown-unknown`
- [ ] `cargo test --target x86_64-unknown-linux-gnu`
- [ ] Verify dependency tree has no `getrandom` / RNG crates

### Phase 3: Polish

- [ ] Bundle format version negotiation (v0.1, v0.2, v0.3)
- [ ] DSSE / in-toto envelope support
- [ ] TSA (RFC 3161) timestamp verification for Rekor v2

## Out of Scope

- **TUF trust root fetching.** This crate is a verifier, not a TUF client. Caller fetches keys via their own TUF implementation (or embeds at compile time) and passes raw bytes to `SigstoreBlobVerifier::new()`. A proper TUF client must verify signed roles, threshold signatures, rollback protection — this is a separate crate.
- **Online verification (Merkle inclusion proofs).** SET-based offline verification is sufficient per the Sigstore spec. Online verification requires network calls to Rekor API, breaks the offline guarantee, and adds complexity without proportional security gain for blob verification.
- **OCI container/image signing.** This crate is a blob verifier only (`cosign verify-blob` equivalent). Container signing involves OCI registries, layers, and manifests — a separate concern.
- **Signing.** This crate is verify-only. The signer never holds a long-lived key in the keyless model — OIDC + Fulcio + CT log interaction is signing infrastructure, not verification.
- **RSA / Ed25519 support.** All production Sigstore keys (Fulcio root, leaf certs, Rekor, CTFE) use ECDSA P-256. RSA and Ed25519 can be added later without API breakage.

## Related

- [Sigstore Client Spec](https://github.com/sigstore/architecture-docs/blob/main/client-spec.md)
- [Cosign Signature Spec](https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md)
- [sigstore-rs](https://github.com/sigstore/sigstore-rs) — Rust reference (aws-lc-rs, not WASM-compatible)
- [sigstore-js `@sigstore/verify`](https://github.com/sigstore/sigstore-js/tree/main/packages/verify/src) — JS reference, modular architecture
- [sigstore-go](https://pkg.go.dev/github.com/sigstore/sigstore-go/pkg/verify) — Go reference
- [x509-parser](https://crates.io/crates/x509-parser) — pure-Rust X.509 parser
- [p256](https://crates.io/crates/p256) — pure-Rust ECDSA P-256 (via `ecdsa` feature)
- [serde_json_canonicalizer](https://crates.io/crates/serde_json_canonicalizer) — RFC 8785 JSON canonicalization
- [regex-lite](https://crates.io/crates/regex-lite) — lightweight regex without Unicode bloat for WASM
- [cosign sigstore deep analys.md](https://github.com/user-attachments/files/29506926/cosign.sigstore.deep.analys.md)

olehbozhok:
### Prepare

-------------------

#### Target issue
  
  
  
  
closes #14465

#### Implementation Details
  
New crate for offline, WASM-compatible Sigstore/Cosign bundle verification.
No network calls during `verify()` — all trust material is compiled in or
supplied by the caller.

**Architecture**

The 10-step verification pipeline runs in `verifier.rs` (the orchestrator).
Each step delegates to a dedicated module:

verifier → bundle(parse) → cert(extract) → tlog(SET)
        → chain(validate) → sct(CTFE) → crypto(signature)
        → policy(identity) → merkle(inclusion proof)

Module dependency tree:

      lib.rs (pub re-exports)
         │
┌────────┼────────┐
verifier  policy  trust_root
    │
    ├── bundle
    ├── cert ── chain ── crypto
    ├── crypto
    ├── sct ── crypto
    ├── tlog ── crypto
    └── merkle

All modules depend on `error.rs` (11-variant `thiserror` enum).

**Trust roots**

Production Fulcio/Rekor/CTFE keys embedded at compile time via
`include_bytes!`. `build.rs` validates every PEM: CA certs must have
`BasicConstraints CA:true` + `KeyUsage keyCertSign`, and their expiration
is checked. Two paths for callers:

- `SigstoreBlobVerifier::with_static_trust_root()` — embedded keys.
- `SigstoreBlobVerifier::new(trust_root_raw)` — caller-provided PEMs.

**Supported bundle formats**

Sigstore bundles v0.1–v0.3, both `messageSignature` and `dsseEnvelope`
media types. DSSE payloads are bound to the artifact via in-toto Statement
`subject[].digest.sha256` comparison (not just envelope-level PAE — full
subject binding). Unknown media types, managed-key bundles, and Rekor v2
proof-only bundles are explicitly rejected.

**Testing strategy — 66 tests, three layers**

1. **Unit tests** (58): each module tested in isolation. Synthetic certs/keys
   via `rcgen` (pure Rust, no OpenSSL). Negative tests assert exact error
   variant, not just `is_err()`.

2. **Real-bundle parity** (7): verifies a genuine public-good Sigstore v0.3
   bundle from the sigstore-conformance corpus against the embedded trust
   root. Committed negative fixtures: corrupted inclusion proof, invalid
   checkpoint signature, wrong checkpoint root hash, messageDigest mismatch
   — all must reject.

3. **Conformance scan** (1, opt-in): runs the full sigstore-conformance
   `bundle-verify` corpus when `SIGSTORE_CONFORMANCE_DIR` is set. Status:
   all hashedrekord positives pass, all negatives rejected, zero
   false-accepts.

**Key files for review**

| File | Lines | Role |
|---|---|---|
| `src/verifier.rs` | 989 | 10-step orchestrator + e2e tests |
| `src/chain.rs` | ~350 | DN-based path building, P-384 Fulcio chain |
| `src/tlog.rs` | ~250 | SET verify, body consistency (CVE-2022-36056) |
| `src/cert.rs` | 497 | X.509 parse via x509-parser, OIDC extensions |
| `src/merkle.rs` | ~100 | Offline RFC 6962 inclusion proof (Trillian fold) |
| `src/sct.rs` | 558 | RFC 6962 SCT verification, precert reconstruction |
| `src/bundle.rs` | ~300 | v0.1–v0.3 format enum, media type dispatch |

**Deliberately out of scope (do not flag)**

- TUF / `trusted_root.json` — PEM-only trust material.
- Rekor v2 / TSA timestamps — SET is always required.
- RSA / Ed25519 — only ECDSA P-256 (leaf/Rekor/CTFE) and P-384 (Fulcio CAs).
- Managed keys — keyless/certificate-based bundles only.
- Fulcio deprecated issuer OID `1.3.6.1.4.1.57264.1.1` — v2 only.
-------------------

## Summary by CodeRabbit

* **New Features**
  * Added offline, WASM-compatible Sigstore/Cosign blob signature verification with certificate-chain validation, SCT checks, Rekor SET authentication, optional Merkle inclusion-proof + checkpoint verification, and strict identity/issuer policy enforcement.
  * Supports both message-signature and DSSE bundle verification.
* **Documentation**
  * Added quick start plus detailed architecture and verification-algorithm references.
* **Tests**
  * Added unit/integration/real-bundle coverage, conformance scanning, and new fixture cases (including inclusion-proof and checkpoint failures).
* **Chores/Style**
  * Updated code style guidance to discourage panic-based runtime validation for input handling.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/f5ccd4e3-1072-4064-8014-ad7d8b4dbffb?refs=356bb94361b8f2028c5d0c366c88575326558f07&source=prChecksComment) |

jans-cedarling/sigstore-verifier/docs/cosign-keyless-verification-algorithm.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Bundle format
+
+| Case | Step | Expect |
+|---|---|---|
+| Unknown mediaType | 1 | REJECT |
+| Malformed JSON | 1 | REJECT |
+| Both `messageSignature` and `dsseEnvelope` present | 1 | REJECT (ambiguous) |
+| Neither present | 1 | REJECT |
+| `inclusionProof` present | 10 | REJECT if checkpoint is absent, unsigned by trusted Rekor key, root hash mismatches, or proof malformed |
```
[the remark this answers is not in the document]
olehbozhok: Already fixed

jans-cedarling/sigstore-verifier/src/verifier.rs
```diff
@@ -0,0 +1,1202 @@
...
+        // Step 3: SET verification — authenticate integratedTime
+        let tlog_entry =
+            parsed
+                .tlog_entry()
+                .ok_or_else(|| SigstoreVerificationError::InvalidBundleFormat {
+                    reason: "bundle has no tlog entries".into(),
+                })?;
+        // Bundle spec: media type v0.2+ requires an inclusion proof (with
+        // checkpoint). v0.1 predates that and may be SET-only.
+        if parsed.version()? >= crate::bundle::BundleVersion::Bundle0_2
```
tareknaser: If I understand this right, a bundle can get out of this check by rewriting its own `mediaType`.

Nothing signs that field (not the cert or the SET or the Rekor body) so if set it to `...version=0.1` then drop `inclusionProof` and this gate passes. Step 10 is `if let Some(proof)` so with the proof gone the Merkle check and the checkpoint check both don't run.

I see `v01_bundle_without_inclusion_proof_still_verifies` so I think the v0.1 case is intentional. What I am not sure about is that the version comes from whoever hands you the bundle so it isn't really a property of the bundle (anyone can flip it). Is that the intent?

jans-cedarling/sigstore-verifier/src/sct.rs
```diff
@@ -0,0 +1,603 @@
...
+    })?;
+
+    let mut any_key_id_matched = false;
+    for sct in &scts {
+        let signed_data = build_digitally_signed_data(sct, &issuer_key_hash, &precert_tbs)?;
+        let hash: [u8; 32] = Sha256::digest(&signed_data).into();
+
+        for key in ctfe_keys {
+            // Only try keys whose key ID matches the SCT's logID.
+            if crate::crypto::p256_key_id(&key.pubkey_bytes)? != sct.log_id {
```
tareknaser: The `?` here gives up on the whole SCT step at the first key that isn't a 65-byte point instead of skipping that key and trying the rest. I think that makes the outcome depend on list order. 

`verify_checkpoint` in `tlog.rs` and `verify_integrated_time` skip bad keys

jans-cedarling/sigstore-verifier/build.rs
```diff
@@ -0,0 +1,117 @@
...
+    let not_after = tbs.validity.not_after.timestamp();
+    let now = Utc::now().timestamp();
+    assert!(
+        not_after >= now,
+        "{filename}: certificate expired at UNIX {not_after} (now: {now}). \
+         Update the trust root PEM files from the Sigstore TUF repository."
+    );
+}
+
+fn validate_public_key(pem_bytes: &[u8], filename: &str) {
```
tareknaser, line 103: This parses the SPKI and prints the OID but I don't see it assert anything about it. So for `rekor.pem`, `ctfe.pem`, and `ctfe_2021.pem`, the only thing checked is that the base64 decodes and the DER parses.

seems like a few places lean on this being a real check. The module doc says a corrupt PEM fails the build and that this "guarantees that `with_static_trust_root()` can `unwrap()` safely at runtime". 
`verifier.rs` has `.expect("trust root keys validated at build time")`

jans-cedarling/sigstore-verifier/tests/conformance_scan.rs
```diff
@@ -0,0 +1,139 @@
...
+use std::path::Path;
+
+use sigstore_verifier::{IdentityMatch, SigstoreBlobVerifier, VerificationPolicy};
+
+const DEFAULT_IDENTITY: &str = "https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon/.github/workflows/extremely-dangerous-oidc-beacon.yml@refs/heads/main";
+const DEFAULT_ISSUER: &str = "https://token.actions.githubusercontent.com";
+
+#[test]
+fn scan_conformance_bundle_verify() {
+    let Ok(root) = std::env::var("SIGSTORE_CONFORMANCE_DIR") else {
```
tareknaser, line 28: `test-cedarling.yml` runs `cargo test --locked --workspace --exclude cedarling_pg` so it does execute.
also nothing in `.github/workflows/` sets `SIGSTORE_CONFORMANCE_DIR`

not sure if it's checked in another workflow maybe?

tareknaser (commented):
Thank you. The PR is very dense and there's a lot to go through. I'll do another review later on

jans-cedarling/sigstore-verifier/src/chain.rs
```diff
@@ -0,0 +1,297 @@
...
+            r.subject_dn == current.issuer_dn && verify_cert_signature(current, r).is_ok()
+        }) {
+            root.validate_ca()?;
+            root.check_validity(integrated_time)?;
+            if leaf_issuer.is_none() {
+                leaf_issuer = Some(root.clone());
+            }
+            return leaf_issuer.ok_or_else(|| SigstoreVerificationError::CertificateChain {
+                reason: "leaf issuer not found on chain path".into(),
+            });
```
dagregi: This part is unreachable since `leaf_issuer` is set to Some right above it

jans-cedarling/sigstore-verifier/src/bundle.rs
```diff
@@ -0,0 +1,298 @@
...
+
+        /// The signatures within the envelope.
+        signatures: Vec<DsseSignature>,
+    },
+}
+
+/// A message digest within a `MessageSignature`.
+#[derive(Debug, Clone, Deserialize)]
+pub(crate) struct MessageDigest {
+    /// The hex-encoded digest value.
```
dagregi: This doc comment is wrong it's base64 encoded not hex

jans-cedarling/sigstore-verifier/tests/conformance_scan.rs
olehbozhok: I added a CI test to check that it is used

jans-cedarling/sigstore-verifier/src/verifier.rs
```diff
@@ -0,0 +1,1647 @@
...
+        // `version()` reads the bundle's self-declared, unsigned `mediaType`
+        // field, so a producer can freely relabel a v0.2+ bundle as v0.1 to
+        // dodge this gate. That doesn't weaken the trust anchor: the SET
+        // verified just below is a Rekor-signed cryptographic proof the
+        // producer cannot forge, so a "downgraded" bundle still needs a
+        // genuine Rekor signature to pass. Skipping the inclusion-proof gate
+        // only forgoes the extra offline Merkle/checkpoint consistency
+        // check, not authentication itself — matching upstream cosign/
+        // sigstore-go, which also treats v0.1 SET-only bundles as valid.
+        if parsed.version()? >= crate::bundle::BundleVersion::Bundle0_2
```
haileyesus2433: This gate reads `mediaType`, which is an unsigned string inside the bundle, so the bundle itself
decides whether the inclusion proof and checkpoint are checked. I relabelled the committed v0.3
fixtures as `version=0.1` and deleted `inclusionProof`, and all three negative fixtures
(`inclusion-proof-corrupted-hash`, `invalid-checkpoint-signature`, `checkpoint-wrong-roothash`)
verified successfully. The comment above argues this is safe because the SET is unforgeable, but
that assumes the Rekor key is honest  the inclusion proof against a signed checkpoint is exactly
the control that survives a compromised or split-view log, and the SET and checkpoint are signed by
the same key. The decision needs to come from the caller, not the bundle: either always require an
inclusion proof, or add an explicit `VerificationPolicy` opt-in for SET-only v0.1.

jans-cedarling/sigstore-verifier/tests/real_bundle.rs
```diff
@@ -0,0 +1,140 @@
...
+        &tampered,
+        &beacon_policy(),
+        |e| matches!(e, SigstoreVerificationError::SignatureMismatch { .. }),
+        "SignatureMismatch for a modified artifact",
+    );
+}
+
+// Real sigstore-conformance negative fixtures — each corrupts one part of the
+// transparency-log evidence; all must be rejected for the *right* reason.
+
```
haileyesus2433, line 136: Related to the `verifier.rs` comment each of these fixtures still verifies after a one-field
downgrade, so as written they prove the corrupted-proof path is rejected only when nobody edits the
media type. Once the gate is fixed, please add the downgraded variant of each fixture as its own
must-reject case so the bypass can't come back.

jans-cedarling/sigstore-verifier/src/bundle.rs
```diff
@@ -0,0 +1,306 @@
...
+        match &self.0.content {
+            BundleContent::MessageSignature { signature, .. } => Some(signature.as_str()),
+            BundleContent::DsseEnvelope { signatures, .. } => {
+                signatures.first().map(|s| s.sig.as_str())
+            },
+        }
+    }
+
+    /// Returns the tlog entry for Rekor verification.
+    #[must_use]
```
haileyesus2433, line 335: `.first()` is the only read of `tlog_entries` anywhere in the crate, so steps 3, 9 and 10 all run
against entry `[0]` and everything after it is parsed and ignored. I appended a second entry with a
garbage `canonicalizedBody` and an `integratedTime` in year 5138 and the bundle still verified. The
issue spec requires every entry to be verified, and the practical problem is that a bundle this
crate has marked "verified" can carry attacker-controlled log entries onward to whoever stores or
re-publishes it. Either verify all entries or reject `tlog_entries.len() != 1`  silently ignoring
them is the worst of the three options.

jans-cedarling/sigstore-verifier/src/bundle.rs
```diff
@@ -0,0 +1,306 @@
...
+    /// The bundle media type (e.g., `application/vnd.dev.sigstore.bundle.v0.3+json`).
+    #[serde(rename = "mediaType")]
+    pub(crate) media_type: String,
+
+    /// The verification material (certificate + tlog entries).
+    #[serde(rename = "verificationMaterial")]
+    pub(crate) verification_material: VerificationMaterial,
+
+    /// The signed content.
+    #[serde(flatten)]
```
haileyesus2433: A bundle carrying both `messageSignature` and `dsseEnvelope` is accepted, and serde just picks one
I added a `dsseEnvelope` to a valid `messageSignature` bundle and it took the DSSE path, so an
attacker-added field decided which content got authenticated. It fails closed today because
`verify_body_consistency` dispatches on the log body's `kind` and catches the mismatch, but that's
incidental rather than structural, and it goes live the moment the two paths diverge or a third
content type is added. Deserialize into one `Option` per variant and reject unless exactly one is
present.

jans-cedarling/sigstore-verifier/src/policy.rs
```diff
@@ -0,0 +1,292 @@
...
+    /// How to match the certificate's Subject Alternative Name.
+    pub cert_identity: IdentityMatch,
+
+    /// The required OIDC issuer (exact match against OID 1.3.6.1.4.1.57264.1.8).
+    pub cert_issuer: String,
+}
+
+/// How to match the certificate identity (SAN).
+#[derive(Debug, Clone)]
+pub enum IdentityMatch {
```
haileyesus2433, line 26: `IdentityMatch::CompiledRegex` can't be fixed at the call site Same issue as `full_match`, but worse the caller hands over an already-compiled unanchored `Regex`,so there's no way to repair it later. This wants a constructor-validated newtype that can only be built through `compile_anchored` suggestion

jans-cedarling/sigstore-verifier/src/cert.rs
```diff
@@ -0,0 +1,623 @@
...
+            });
+        }
+        Ok(())
+    }
+}
+
+/// Simple PEM-to-DER conversion without relying on the `pem` crate.
+///
+/// Extracts the base64 content between `-----BEGIN ...-----` and `-----END ...-----`.
+pub(crate) fn parse_pem_to_der(pem_bytes: &[u8]) -> Option<Vec<u8>> {
```
haileyesus2433, line 452: this returns `Some([])` on non-PEM input and drops all but the first block

With no `-----BEGIN` line, `in_body` never flips, `b64` stays empty, and `decode("")` returns
`Ok(vec![])`, so `parse_pem_to_der(b"not a pem")` is `Some([])` rather than `None`. That makes the
`unwrap_or_else` DER fallbacks at `tlog.rs`  unreachable for any valid-UTF-8 input  documented raw-DER support works only because binary DER usually fails `from_utf8` first.

haileyesus2433: Also, `break` at the first `-----END` means a concatenated PEM yields exactly one certificate,
so someone configuring `SigstoreTrustRootRaw` with the standard multi-cert `fulcio.crt.pem` silently
gets one root and no error. Truncated PEM (no `-----END` at all) is accepted too.

jans-cedarling/sigstore-verifier/src/trust_root.rs
```diff
@@ -0,0 +1,258 @@
...
+            .map(|pem| Cert::from_pem(pem))
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let all_fulcio_intermediates: Vec<Cert> = self
+            .fulcio_intermediate_certs
+            .iter()
+            .map(|pem| Cert::from_pem(pem))
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let now = chrono::Utc::now().timestamp();
```
haileyesus2433, line 96: trust anchors are filtered on wall-clock time

Every other time comparison in the crate is anchored on the SET-verified `integratedTime` precisely
so verification doesn't depend on the verifier's clock, and this is the one place that isn't  for
the trust anchor itself. Once a Fulcio root passes `not_after`, every historical bundle it anchored
becomes unverifiable, even though `chain.rs` already checks the root against `integrated_time`
correctly and is just made unreachable by this filter. It also matters for the WASM target under
`wasmbind`, `Utc::now()` lowers to a `js_sys::Date` call, so a host reporting epoch 0 drops every
root and surfaces as a confusing "no trusted path" error  and it pulls `chrono`, `js-sys` and
`wasm-bindgen` into the crate. Dropping this call removes all three dependencies and the invariant violation at once.

jans-cedarling/sigstore-verifier/src/tlog.rs
```diff
@@ -0,0 +1,1146 @@
...
+/// <origin>
+/// <tree_size>
+/// <base64 root hash>
+///
+/// — <key_name> <base64: keyhint[4] || ecdsa_signature>
+/// ```
+/// The signed bytes are the body lines (origin, size, root hash) each ending in
+/// `\n` — up to but excluding the blank line before the signature. The keyhint
+/// is the first 4 bytes of `SHA-256(SubjectPublicKeyInfo DER)` of the Rekor key.
+pub(crate) fn verify_checkpoint(
```
haileyesus2433, line 451: the checkpoint isn't bound to the log identity or to the entry's `logId`

`origin` is parsed and discarded, and key selection loops over every trusted Rekor key matching on a
4-byte keyhint. `verify_integrated_time` at `verifier.rs` does the right thing  it filters by
full SHA-256 SPKI ID against `tlog_entry.log_id.key_id`, with a comment saying it's so a bundle
can't pick which key verifies it and that wasn't carried down here.

jans-cedarling/sigstore-verifier/build.rs
```diff
@@ -0,0 +1,155 @@
...
+        not_after >= now,
+        "{filename}: certificate expired at UNIX {not_after} (now: {now}). \
+         Update the trust root PEM files from the Sigstore TUF repository."
+    );
+
+    // Fail early, well before actual expiry, so rotation happens on a planned
+    // schedule instead of as an emergency once the cert has already expired.
+    let seconds_left = not_after - now;
+    let warn_window_seconds = EXPIRY_WARN_WINDOW_DAYS * 24 * 60 * 60;
+    assert!(
```
haileyesus2433, line 90: a hard `assert!` on wall-clock time makes the build self-destruct

This means the crate stops compiling on a future date with no code change, every downstream consumer of `jans-cedarling` breaks at once, and previously released tags become un-buildable reproducing an old release would need a `build.rs` patch or a faked clock. 
It's also cached `rerun-if-changed=src/trust/` means this time-dependent check only re-runs when the PEM files change, so the failure lands at an unpredictable moment rather than on a knowable date.

.github/workflows/test-sigstore-verifier-conformance.yml
```diff
@@ -0,0 +1,49 @@
+name: sigstore-verifier Conformance Scan
+
```
haileyesus2433, line 2: the WASM target has no CI coverage `cargo build --target wasm32-unknown-unknown` is in the issue's acceptance criteria, but `sigstore-verifier` is never built for wasm anywhere in CI.

jans-cedarling/sigstore-verifier/Cargo.toml
```diff
@@ -0,0 +1,55 @@
...
+chrono = { workspace = true }
+
+p256 = { version = "0.13", default-features = false, features = ["ecdsa"] }
+# Fulcio root + intermediate CAs are P-384 (signing leaves with ecdsa-with-SHA384),
+# so chain-link signature verification needs P-384. Leaf/Rekor/CTFE keys are P-256.
+p384 = { version = "0.13", default-features = false, features = ["ecdsa"] }
+ecdsa = { version = "0.16", default-features = false, features = ["der", "verifying"] }
+x509-parser = { version = "0.18", default-features = false }
+serde_json_canonicalizer = "0.3"
+der = { version = "0.8", default-features = false, features = ["oid"] }
```
haileyesus2433: declared but never used There isn't a single reference to `der` in `src/` or `build.rs`;

haileyesus2433: declared but never used There isn't a single reference to `der` in `src/` or `build.rs`;

jans-cedarling/sigstore-verifier/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+
+pub use error::SigstoreVerificationError;
+pub use policy::{IdentityMatch, VerificationPolicy};
+pub use trust_root::SigstoreTrustRootRaw;
+pub use verifier::{SigstoreBlobVerifier, VerifiedSignature};
+
+/// Re-exported so callers can construct [`IdentityMatch::CompiledRegex`]
+/// without adding their own `regex-lite` dependency (and risking a version
+/// mismatch with the one this crate was built against).
+pub use regex_lite;
```
haileyesus2433, line 57: Any semver bump of `regex-lite` becomes a breaking change here. The reason given (letting callers
build `CompiledRegex` without a version mismatch) is real, but the anchored-regex newtype suggested
on `policy.rs` removes the need to expose the foreign type at all, which solves both.

haileyesus2433 (changes requested):

jans-cedarling/sigstore-verifier/Cargo.toml
olehbozhok: resolved

olehbozhok: resolved

jans-cedarling/sigstore-verifier/src/cert.rs
olehbozhok: resolved

jans-cedarling/sigstore-verifier/src/bundle.rs
olehbozhok: resolved

jans-cedarling/sigstore-verifier/src/bundle.rs
olehbozhok: resolved

jans-cedarling/sigstore-verifier/src/verifier.rs
olehbozhok: resolved

jans-cedarling/sigstore-verifier/src/tlog.rs
olehbozhok: resolved

.github/workflows/test-sigstore-verifier-conformance.yml
olehbozhok: resolved

jans-cedarling/sigstore-verifier/src/lib.rs
olehbozhok: The semver point is fair — `pub use regex_lite;` does put a foreign type on the public surface.

The proposed remedy no longer applies though: following your `policy.rs` thread I checked what upstream does, and nobody auto-anchors. sigstore-go's `SubjectAlternativeNameMatcher` uses `Regexp.MatchString` and documents *"regexp matching is not anchored by default; use `^...$` if you intend to match the entire SAN value"*; cosign's `CheckCertificatePolicy` is the same; sigstore-python has no regex at all. Anchoring is now removed from this crate, so there's no invariant left for an anchored-regex newtype to enforce.

Keeping the re-export for now: so there's nothing for a semver bump to break — and the re-export is what lets callers configure their own `Regex` (`RegexBuilder` has `case_insensitive`, `multi_line`, `size_limit`), which a newtype taking `&str` would remove. Worth revisiting when the crate is published and there's a real consumer to design against.

jans-cedarling/sigstore-verifier/src/trust_root.rs
olehbozhok: Both observations are correct. `chain.rs` already checks the root against `integrated_time` (`chain.rs:58`, alongside the leaf at `:37` and intermediates at `:93`), so this filter rejects nothing the chain walk wouldn't — it only removes anchors, and it does pull `chrono` → `js-sys`/`wasm-bindgen` on wasm32 (confirmed via `cargo tree --target wasm32-unknown-unknown`).

Keeping it as a deliberate choice: once an embedded CA is past its own `not_after`, we'd rather stop trusting it outright than keep accepting bundles anchored to it. The accepted cost is exactly what you describe — historical bundles that were legitimately signed inside that root's validity window become unverifiable. That's a fail-closed stance on expired trust anchors, not a claim that the filter adds a control.

The wasm side now has CI coverage: this PR adds a `wasm32-unknown-unknown` build job, so the `js-sys` dependency can't change unobserved.


We will try to fix the problem in the future before certificate expiration

jans-cedarling/sigstore-verifier/build.rs
olehbozhok: It is fine and designed to be like that. We need to add new certificates at that time.

olehbozhok:
@coderabbitai  do you have any suggestions, or can you approve?

dagregi (dismissed):

haileyesus2433 (dismissed):
LGTM

olehbozhok:
@coderabbitai approve

.github/workflows/test-sigstore-verifier-conformance.yml
```diff
@@ -0,0 +1,82 @@
+name: sigstore-verifier Conformance Scan and WASM Build
```
moabu: Following the new conventions please clearly name the workflow. It should be double quoted with its domain forexample “Test: ……."

moabu (changes requested):

dagregi (approved):

.github/workflows/test-sigstore-verifier-conformance.yml
olehbozhok: resolved

moabu (approved):

PR #14653, opened 2026-07-28
fix(jans-fido2): capture ip address, user agent and session id in passkey metrics

### LINKED ISSUE #14650: fix(jans-fido2): Passkey metrics entries never record ipAddress, userAgent or deviceInfo

The FIDO2 passkey metrics entries returned by `GET /jans-fido2/restv1/metrics/entries` never contain `ipAddress`. The same is
true for `userAgent`, `deviceInfo`, `sessionId` and `metricType` — the values are not merely absent from the API response, they are never written to the database. Confirmed by exporting the `jansFido2MetricsEntry` table directly: the `jansFido2MetricsIpAddress`, `jansFido2MetricsUserAgent`, `jansFido2MetricsDeviceInfo`, `jansFido2MetricsSessionId` and `jansFido2MetricsType` columns are empty on every row, while `userId`, `username`, `operationType`, `status`, `authenticatorType` and `nodeId` are populated normally.

This makes the entries unusable for the auditing and per-origin analysis they are meant to support, and it silently contradicts the telemetry documentation, which already tells operators these responses may contain "IP address, user-agent, session ID".

There are two independent causes, both silent:

1. `AttestationService` and `AssertionService` are `@ApplicationScoped` CDI beans that declare `@Context private HttpServletRequest  
    httpRequest`. `@Context` is only processed by RESTEasy for JAX-RS components, so the field is never injected and stays null. The 
    same null request is also handed to custom interception scripts via `ExternalFido2Context`.
2. Even with a valid request, `MetricService` reads it inside  `CompletableFuture.runAsync(...)`, i.e. off the request thread, where a 
    request-scoped proxy is no longer resolvable.

Both failures are swallowed by a `log.debug`, so nothing appears in the logs at default log level.

Passkey metrics entries should record the request context of the operation:

- `ipAddress` — proxy headers first, socket remote address as fallback
- `userAgent` — the `User-Agent` header
- `deviceInfo` — browser / OS / device type, when `fido2DeviceInfoCollection` is enabled
- `sessionId` — the session the operation belongs to
- `metricType` — the metric name of the event, e.g. `fido2_registration_success`

The request must be read synchronously, on the request thread, and the resulting values carried into the asynchronous write.

- Injecting `@Context HttpServletRequest` in the controllers (which are genuine JAX-RS resources, where `@Context` works) and 
  threading it down. Correct, but changes the public signatures of `AttestationService.options/verify` and `AssertionService.options/generateOptions/verify`.
- Making the metrics write synchronous. Rejected — it would put persistence latency on the registration and authentication hot path.
- Leaving the async read and retrying on failure. Rejected — the request is genuinely gone by then; no retry can recover it.

**Additional context**

Affects every deployment with `fido2MetricsEnabled=true`. Existing rows cannot be backfilled; only entries written after the fix will carry the data.

Reproducer, no authenticator needed — the options call alone writes an `ATTEMPT` entry:

imran-ishaq:
### Prepare

-------------------

#### Target issue

closes #14650 

#### Implementation Details

FIDO2 metrics entries were persisted with empty `ipAddress`, `userAgent`, `deviceInfo`, `sessionId` and `metricType`. Two independent causes, both silently swallowed by a `log.debug`:

1. **The request was null.** `AttestationService` and `AssertionService` are `@ApplicationScoped` CDI beans declaring 
    `@Context private HttpServletRequest`.  `@Context` is only honoured for JAX-RS components, so the field was never 
      injected.
3. **The request was read off the request thread.** `MetricService` read it inside `CompletableFuture.runAsync(...)`, where a 
      request-scoped proxy is no longer resolvable.

Cause 1 is what the persisted data actually shows: had the request been non-null but throwing, the `deviceInfo` catch would have fallen back to `createMinimalDeviceInfo()` and stored `"Unknown"` values. The column is blank, so the extraction block never ran at all.

Changes:

- **`MetricService`** — added an immutable `RequestSnapshot` and `snapshotRequest()`,  invoked *before* `CompletableFuture.runAsync` in `recordRegistrationMetrics` and  `recordAuthenticationMetrics`. `createMetricsData` no longer accepts an `HttpServletRequest` at all, so an off-thread read cannot be reintroduced.
- **`AttestationService` / `AssertionService`** — `@Context` → `@Inject` on the `HttpServletRequest` field. `weld-web` is on the runtime classpath and its `HttpServletRequestBean.getScope()` returns `RequestScoped`, so Weld injects a client proxy that resolves per-call on the request thread. `httpResponse` intentionally keeps `@Context` — `HttpServletResponse` is not a Weld built-in bean. This also fixes the null request previously handed to custom scripts via `ExternalFido2Context`.
- **`sessionId`** — resolved from the auth server's `session_id` cookie, falling back to  the servlet session. It was never set anywhere in production code before.
- **`metricType`** — set from `Fido2MetricType.getMetricName()` (e.g. `fido2_registration_success`), including on the fallback path, and mapped onto the  entry in `Fido2MetricsService.setOptionalFields`.

`applicationType` is deliberately left unset, per the existing comment — it is always `FIDO2` and redundant.

`server-fips/` contains no Java sources (POM-only repackaging), so there is nothing to mirror; it was rebuilt and passes.

Note: existing rows are not backfilled — only entries written after deployment carry the data.

-------------------

Three tests added to `MetricServiceTest`:

- `testRequestIsReadOnCallingThread` — records which thread dereferences the request and asserts it is the caller's. Verified 
   as a genuine regression test by stashing the fix:  it fails with the read landing on `ForkJoinPool-1-worker-4`.
- `testInactiveRequestScopeDoesNotPropagate` — a proxy throwing outside an active request degrades to an empty 
   snapshot instead of breaking the operation.
- `testNullRequestIsTolerated` — callers with no request at all. Full suite: 238 tests, 0 failures. Full `jans-fido2` build including `server-fips` passes.

Docs: `docs/janssen-server/fido/passkey-telemetry.md` gains a "Request context on raw entries" table covering the six fields and their sources, plus a warning that `ipAddress` is only as trustworthy as the deployment's `X-Forwarded-For` handling.

## Summary by CodeRabbit

* **New Features**
  * Passkey telemetry now captures metric type and richer per-entry request context (IP address, user agent, session ID, node ID, and optional device information).

* **Bug Fixes**
  * Improved request option validation so missing payloads return HTTP 400.
  * Enhanced session ID extraction by preferring the `session_id` cookie (with fallback).
  * Strengthened IP validation, including rejecting IPv4 inputs with non-standard digits.

* **Documentation**
  * Updated passkey telemetry “How it works” with clearer request-context field descriptions and reverse-proxy guidance.
  * Clarified behavior when device-info collection is disabled.

* **Tests**
  * Expanded coverage to ensure request data is read on the request thread and metrics degrade safely when request context is unavailable.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/b278d62d-19a0-493b-b83b-5668fa8c9a1f?refs=e58513387e35d2e9fb502267aa3ade556b658bbf&source=prChecksComment) |

yurem (approved):

Issue #14437, opened 2026-06-24
feat(jans-cedarling): add policy store validation command


olehbozhok:
## Summary

Add a `validate` subcommand to `cedarling_cli` that validates a Cedarling policy store
on three levels — parse, schema, and metadata — without starting the authorization
engine. Works with all policy store sources: local files, directories, `.cjar` archives,
and URLs.

## Motivation

Policy errors discovered at runtime cause authorization failures (deny-by-default =
user blocked) or silent misconfigurations. Validation at CI/CD time catches syntax
errors, schema mismatches, missing entity types, and invalid metadata before deployment.

## CLI

```bash
cedarling_cli validate --policy-store ./policies/                  # directory
cedarling_cli validate --policy-store ./policies.yaml              # YAML file
cedarling_cli validate --policy-store ./policy_store.cjar          # .cjar archive
cedarling_cli validate --policy-store https://.../policies.cjar    # .cjar via URL
```

All sources supported by `PolicyStoreSource`: `FileJson`, `FileYaml`, `CjarFile`,
`CjarUrl`, `Directory`, `Uri` (auto-detect).

## Validation levels

1. **Parse** — Cedar policy syntax via `PolicyParser`, with file/line/column error location
2. **Schema validate** — policies against the Cedar schema via `cedar_policy::Validator`
3. **Metadata validate** — policy store structure via `MetadataValidator`

## Output

```
validating ./policies/
  parse ...................... ok
  schema ..................... ok
  metadata ................... ok
validation passed

# On failure:
validating ./policies/
  parse ...................... ok
  schema ..................... FAIL
    policy allow_read.cedar:12:5: entity type 'Document' not found in schema
  metadata ................... ok
validation failed: 1 error(s)
```

Exit 0 on success, non-zero on failure. Suitable for CI gates:
```bash
cedarling_cli validate --policy-store ./policies && git push
```

## What exists

- `PolicyStoreSource` + `load_policy_store()` — loading from all source types
- `PolicyParser` — Cedar parsing with location-aware errors
- `cedar_policy::Validator::validate()` — schema validation
- `MetadataValidator` — policy store metadata validation

## What to do

1. Subcommand `validate` in `cedarling_cli` binary
2. `--policy-store` accepts a path or URL (reuses existing `PolicyStoreSource` parsing)
3. Load the policy store via the same loader used at bootstrap
4. Run all three validation levels, collect errors with file/line/column
5. Print structured output, exit 0 or non-zero

## Key files

[...]

PR #14673, opened 2026-07-31
feat(jans-cedarling): add policy testing framework with validation

### LINKED ISSUE #14436: feat(jans-cedarling): add policy testing framework
## Summary

Add a `cedarling_cli` binary with a `test` subcommand that runs YAML-defined test suites
against a Cedarling policy store. Unlike `cedar run-tests` (which tests pure Cedar
policies against raw Cedar entities), this tool tests the Cedarling-specific layer:
policy store YAML format, `EntityData`, and bootstrap configuration.

## Motivation

There is currently no way for a policy author to test Cedarling policies locally without
writing a full application. This closes the policy lifecycle gap: author → test locally →
deploy. Combined with `validate` (#14437), this gives policy authors the same quality gates
application code already has.

## CLI

```bash
# Basic usage
cedarling_cli test \
  --policy-store ./policies.yaml \
  --test-file ./tests.yaml

# With full bootstrap config
cedarling_cli test \
  --config ./cedarling.yaml \
  --test-file ./tests.yaml
```

Bootstrap parameters are set via:
- Environment variables (`CEDARLING_POLICY_STORE_URI`, `CEDARLING_LOG_TYPE`, etc.)
- CLI flags (`--policy-store`, `--log-type`, etc.)
- Config file (`--config`)

## Test file format (YAML)

```yaml
tests:
  - name: "teacher can read document"
    request:
      principal:
        type: "Jans::User"
        id: "alice"
        attributes: { role: ["Teacher"] }
      action: 'Jans::Action::"Read"'
      resource:
        type: "Jans::Document"
        id: "doc1"
      context: {}
    result:
      decision: Allow
      reason_ids: ["allow_public_read"]   # optional
      num_errors: 0                       # optional
```

Each test verifies:
- `decision` — Allow or Deny (required)
- `reason_ids` — which policies determined the decision (optional)
- `num_errors` — expected number of evaluation errors (optional)

## Output

```
running 3 test(s)
  test teacher can read document ... ok
  test student cannot read document ... ok
  test admin can edit ... fail: expected Deny, got Allow
results: 2 passed, 1 failed
```

## Coverage report

After all tests pass, compare all policy IDs in the store against the union of all
`reason_ids` encountered:

```
Coverage: 3/4 policies triggered (75%)
  ✓ allow_public_read
  ✓ deny_restricted
  ✓ allow_teacher_edit
  ✗ dead_policy_never_used
```

## What to do

1. Create new `cedarling-cli` crate with `[[bin]]` target, depending on `cedarling` as library
2. Bootstrap Cedarling from env vars + CLI flags + config file
3. Parse YAML test files, bootstrap Cedarling for each test suite
4. Call `authorize_unsigned` for each test case, compare results
5. Print pass/fail with colors, exit code
6. Compute and print coverage report

## Additional subcommands (same binary)

- `cedarling_cli authorize` — single authorize_unsigned call for debugging
- `cedarling_cli validate` — validate policy store (see #14437 )

## Key files

- `cedarling-cli/` (new crate: `Cargo.toml`, `src/main.rs`)
- `cedarling/src/authz/authorize_result.rs`

haileyesus2433:
### Prepare

-------------------

#### Target issue
  
  
  
  
closes #14436 #14437 

#### Implementation Details
  

-------------------

## Summary by CodeRabbit

* **New Features**
  * Added a Cedarling CLI for local policy testing, authorization checks, policy validation, and coverage reporting.
  * Added configurable policy stores, JSON request attributes, YAML test definitions, and Cjar URL support.
  * Exposed policy metadata and policy-store validation through the Cedarling API.
  * Added Cedar version validation for policy stores.
  * Improved schema loading when running in non-strict mode.

* **Documentation**
  * Added CLI installation, usage, configuration, test format, coverage, and exit-code guidance.

* **Tests**
  * Added coverage for authorization tests, validation outcomes, configuration failures, and metadata errors.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/ce5be154-9a3c-4749-b921-4dbffa15b234?refs=f242b79c9087ebbb37bbd2f15ea9123d7619ff92&source=prChecksComment) |

jans-cedarling/cedarling/src/lib.rs
```diff
@@ -147,6 +149,159 @@ pub struct Cedarling {
...
+    /// Intended for tooling that needs to enumerate the full policy set
+    /// (coverage reports, dashboards). Ordering is unspecified.
+    #[must_use]
+    pub fn all_policy_metadata(&self) -> Vec<PolicyMetadata> {
+        self.authz.load().all_policy_metadata()
+    }
+
+    /// Run parse / schema / metadata validation against a policy-store source
+    /// without initializing the authorization engine. Never returns Err for
+    /// validation failures — those land in the returned report; only
```
dagregi: Is this doc comment trailing off mid sentence at the end with the 'only'?

jans-cedarling/cedarling/src/lib.rs
```diff
@@ -147,6 +149,159 @@ pub struct Cedarling {
...
+    #[allow(clippy::too_many_lines)]
+    pub async fn validate_policy_store(
+        config: &PolicyStoreConfig,
+    ) -> Result<ValidationReport, ValidateInfraError> {
+        // 1. Build a fresh HttpClient
+        let http_client = crate::http::HttpClient::new(crate::http::HttpClientConfig::default())?;
+
+        // 2. Call load_policy_store
+        let load_result =
+            crate::init::policy_store::load_policy_store(config, &http_client, false).await;
```
dagregi: Is there a reason this is hardcoded to false instead of strict_schema_validation from the bootstrap config?

dagregi (commented):
It would be nice to add example commands in the readme for users to test it out or maybe add it in the docs in a separate PR. Also in my opinion we should use just cedarling instead of cedarling_cli for the binary name

jans-cedarling/README.md
```diff
@@ -208,6 +208,10 @@ The python bindings for `Cedarling` is located in the `bindings/cedarling_python
 
 Or you can find readme by clicking [here](bindings/cedarling_python/README.md).
 
+## Cedarling CLI
+
+The `cedarling-cli` is a local test runner and debug utility for Cedarling policy stores. It allows policy authors to validate and test policies (with coverage reporting) without running a full application. You can find its documentation [here](cedarling-cli/README.md).
```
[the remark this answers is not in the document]
haileyesus2433: addressed in https://github.com/JanssenProject/jans/pull/14673/commits/f68c4b37e84c71bc4f3166002ed876704cfcfb07

jans-cedarling/cedarling/src/tests/validate.rs
```diff
@@ -0,0 +1,159 @@
...
+    let config: crate::BootstrapConfig = raw_config.try_into().expect("should parse");
+    let report = Cedarling::validate_policy_store(&config.policy_store_config)
+        .await
+        .expect("infra layer ok");
+
+    assert!(report.is_ok());
+    assert!(report.parse.is_ok_or_skipped());
+    assert!(matches!(report.schema, LevelResult::Skipped { .. }));
+    assert!(report.metadata.is_ok_or_skipped());
+}
```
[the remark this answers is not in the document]
haileyesus2433: addressed in https://github.com/JanssenProject/jans/pull/14673/commits/f68c4b37e84c71bc4f3166002ed876704cfcfb07

jans-cedarling/cedarling/src/lib.rs
```diff
@@ -147,6 +149,159 @@ pub struct Cedarling {
...
+                                line: None,
+                                column: None,
+                                message: e.to_string(),
+                            }],
+                        },
+                    }
+                } else {
+                    LevelResult::Ok
+                };
+
```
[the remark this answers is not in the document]
haileyesus2433: addressed in https://github.com/JanssenProject/jans/pull/14673/commits/f68c4b37e84c71bc4f3166002ed876704cfcfb07

jans-cedarling/cedarling/src/lib.rs
```diff
@@ -147,6 +149,159 @@ pub struct Cedarling {
...
+    /// Run parse / schema / metadata validation against a policy-store source
+    /// without initializing the authorization engine. Never returns Err for
+    /// validation failures — those land in the returned report; only
+    #[allow(clippy::too_many_lines)]
+    pub async fn validate_policy_store(
+        config: &PolicyStoreConfig,
+    ) -> Result<ValidationReport, ValidateInfraError> {
+        // 1. Build a fresh HttpClient
+        let http_client = crate::http::HttpClient::new(crate::http::HttpClientConfig::default())?;
```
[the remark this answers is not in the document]
haileyesus2433: addressed in https://github.com/JanssenProject/jans/pull/14673/commits/f68c4b37e84c71bc4f3166002ed876704cfcfb07

jans-cedarling/cedarling-cli/src/test/runner.rs
```diff
@@ -0,0 +1,171 @@
+use anyhow::{bail, Context, Result};
```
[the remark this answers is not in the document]
haileyesus2433: addressed in https://github.com/JanssenProject/jans/pull/14673/commits/f68c4b37e84c71bc4f3166002ed876704cfcfb07

jans-cedarling/cedarling/src/authz/mod.rs
```diff
@@ -1592,4 +1623,41 @@ mod tests {
...
+            ..Default::default()
+        };
+
+        let config: BootstrapConfig = raw_config.try_into().expect("should parse config");
+        let cedarling = match Cedarling::new(&config).await {
+            Ok(c) => c,
+            Err(e) => {
+                // policy-store_ok.yaml contains real external endpoints (like test.jans.org).
+                // If it fails due to network/timeout errors, gracefully skip to avoid flaky CI.
+                println!("Skipping test due to initialization error (likely network timeout): {e}");
```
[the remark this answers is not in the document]
haileyesus2433: this is fine since it wont be used in wasm, its a separate binary

jans-cedarling/cedarling/src/tests/validate.rs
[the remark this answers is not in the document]
haileyesus2433: check here https://github.com/JanssenProject/jans/pull/14673/commits/481fd1a68d636c8f5306acde07ceacdc2f97d2b6

jans-cedarling/cedarling/src/lib.rs
haileyesus2433: It is intentionally hardcoded to false because we want load_policy_store to return the parse LoadedPolicyStore regardless of whether the schema validates successfully. If we passed strict_schema_validation from the bootstrap config, load_policy_store would abort internally upon encountering a schema error and return a PolicyStoreLoadError. That would prevent us from extracting the granular, line-by-line Cedar validation diagnostic messages we capture further down. By bypassing the loader's strict validation, we separate the parse success from the schema failure, which lets us cleanly populate the .parse and .schema sections of the ValidationReport independently!

jans-cedarling/cedarling/src/lib.rs
haileyesus2433: The sentence continues on the next line! It reads: ...only returns Err for underlying I/O or network failures. It was removed while fixing the 100 character limit now its just wrapped a bit awkwardly.

olehbozhok:
why do we have changes in `cedarling_pg` and other not related folders?

dagregi:
> why do we have changes in `cedarling_pg` and other not related folders?

It looks like `cargo fmt` was run, so most of it maybe formatting fixes?

olehbozhok:
> > why do we have changes in `cedarling_pg` and other not related folders?
> 
> It looks like `cargo fmt` was run, so most of it maybe formatting fixes?

Yeah, but also after running `cargo fmt` locally, I can see changes, but shouldn't.
Anyway, better to commit only the related part of the code.

dagregi:
> > > why do we have changes in `cedarling_pg` and other not related folders?
> > 
> > 
> > It looks like `cargo fmt` was run, so most of it maybe formatting fixes?
> 
> Yeah, but also after running `cargo fmt` locally, I can see changes, but shouldn't. Anyway, better to commit only the related part of the code.

yeah if all the changes not related to this pr are all formatting fixes then it should probably be done separately if needed

jans-cedarling/cedarling/src/common/policy_store/legacy_store/mod.rs
```diff
@@ -544,6 +544,7 @@ impl From<LegacyPolicyStore> for super::PolicyStore {
 #[derive(Debug, Clone)]
 #[cfg_attr(test, derive(PartialEq))]
 pub(crate) struct LegacyAgamaPolicyStore {
+    pub cedar_version: String,
```
olehbozhok, line 547: I think better to have strong typed variant. If need we can have type that wraps semver (with own parsing)

jans-cedarling/cedarling/src/common/policy_store/legacy_store/mod.rs
```diff
@@ -562,7 +563,16 @@ impl<'de> Deserialize<'de> for LegacyAgamaPolicyStore {
             de::Error::custom("missing required field 'policy_stores' in policy store")
         })?;
 
+        let cedar_version = match obj.get("cedar_version") {
+            Some(v) => v
+                .as_str()
+                .ok_or_else(|| de::Error::custom("'cedar_version' must be a string if present"))?
+                .to_string(),
+            None => "v4.0.0".to_string(),
```
olehbozhok: I'm not sure if we need `v` in the version. We should comunicate with AgamaLab team

olehbozhok:
`cedar_version` on `LegacyAgamaPolicyStore` ([legacy_store/mod.rs:547](jans-cedarling/cedarling/src/common/policy_store/legacy_store/mod.rs#L547)) is format-validated but never actually used for anything.

Trace:
- `validate_legacy_store` ([validator.rs:37-51](jans-cedarling/cedarling/src/common/policy_store/validator.rs#L37-L51)) checks the format (strips `v` prefix, `semver::Version::parse`), called once from `process_legacy_agama_store` at load time ([init/policy_store.rs:205](jans-cedarling/cedarling/src/init/policy_store.rs#L205)).
- The result is discarded — `extract_first_policy_store` sets `metadata: None` ([init/policy_store.rs:75](jans-cedarling/cedarling/src/init/policy_store.rs#L75)), so `cedar_version` never reaches the runtime `PolicyStoreMetadata`.
- The actual cedar-language compat gate (`metadata.is_compatible_with_cedar(...)` vs `cedar_policy::get_lang_version()`, [lib.rs:763-787](jans-cedarling/cedarling/src/lib.rs#L763-L787)) only runs for non-legacy stores, since it needs `metadata: Some(...)`.
- The public `Cedarling::validate_policy_store()` API also doesn't re-check it for legacy sources — `validate_legacy_metadata` ([validator.rs:221-228](jans-cedarling/cedarling/src/common/policy_store/validator.rs#L221-L228)) checks `PolicyStoreInfo.version` (a different field), not `cedar_version`.

Net effect: legacy `cedar_version` is parsed, format-gated, then thrown away. It never gates anything.

Two ways to make it meaningful instead of just deleting it (it's harmless as a garbage-input filter, but its stated purpose — declaring which cedar version legacy policies target — isn't realized):

1(prefer). Wire it through: set `metadata: Some(...)` in `extract_first_policy_store` using the legacy `cedar_version`, so the real compat check in `lib.rs` runs for legacy stores too.
2. Leave it as a pure format-gate, but that's worth calling out explicitly (e.g. doc comment) so it's not mistaken for an actual compat mechanism.

resolved in https://github.com/JanssenProject/jans/pull/14673/commits/07744bef25e93187941e23d96a0f1378cbb8492b i went with the first suggestion

jans-cedarling/cedarling/src/init/policy_store.rs
```diff
@@ -209,6 +198,21 @@ pub(crate) async fn load_policy_store(
     Ok(loaded)
 }
 
+fn process_legacy_agama_store(
+    agama_policy_store: &LegacyAgamaPolicyStore,
+    strict_schema_validation: bool,
+) -> Result<LoadedPolicyStore, PolicyStoreLoadError> {
+    crate::common::policy_store::validator::MetadataValidator::validate_legacy_store(
```
olehbozhok: Why we can't import validator on top of the file?

jans-cedarling/cedarling/src/init/policy_store.rs
```diff
@@ -334,7 +339,22 @@ async fn load_policy_store_from_cjar_file(
 
     let loaded = loader::load_policy_store_archive(path, strict_schema_validation)
         .await
-        .map_err(|e| PolicyStoreLoadError::Archive(format!("Failed to load from archive: {e}")))?;
+        .map_err(|e| match e {
+            crate::common::policy_store::errors::PolicyStoreError::Validation(ve) => {
```
olehbozhok: same `crate::common::policy_store::errors::` looks ugly

jans-cedarling/cedarling/src/init/policy_store.rs
```diff
@@ -43,6 +43,9 @@ pub enum PolicyStoreLoadError {
     /// Failed to load policy store from directory.
     #[error("Failed to load policy store from directory: {0}")]
     Directory(String),
+    /// Policy store validation error.
+    #[error("Policy store validation error: {0}")]
+    Validation(String),
```
olehbozhok: Why we use string instead of strong typed error?

jans-cedarling/cedarling/src/init/policy_store.rs
olehbozhok: Maybe it worst to move to method?

jans-cedarling/cedarling/src/init/policy_store.rs
```diff
@@ -501,7 +549,12 @@ fn load_policy_store_from_archive_bytes(
     // Load from bytes (works in both native and WASM)
     let loaded =
         loader::load_policy_store_archive_bytes(bytes, strict_schema_validation).map_err(|e| {
-            PolicyStoreLoadError::Archive(format!("Failed to load from archive bytes: {e}"))
+            match e {
+                crate::common::policy_store::errors::PolicyStoreError::Validation(ve) => {
+                    PolicyStoreLoadError::Validation(ve.to_string())
+                },
+                _ => PolicyStoreLoadError::Archive(e.to_string()),
```
olehbozhok: I belive that in `Archive` we can store `PolicyStoreError` type instead of stirring

haileyesus2433:
> > > why do we have changes in `cedarling_pg` and other not related folders?
> > 
> > 
> > It looks like `cargo fmt` was run, so most of it maybe formatting fixes?
> 
> Yeah, but also after running `cargo fmt` locally, I can see changes, but shouldn't. Anyway, better to commit only the related part of the code.

they are not formating changes only i also added descriptive messages to the error's it was being flagged by coderabbit alot

jans-cedarling/cedarling/src/common/policy_store/legacy_store/mod.rs
haileyesus2433: cedar_version is kept as a generic String at this layer is to ensure we don't fail too eagerly during serde deserialization. If it were a strongly-typed semver::Version parsing wrapper, providing a malformed version (like v4.x or invalid-format) would immediately trigger a JSON Parse error.  By keeping it a string during serialization, we let serde succeed and then its validate it later using `MetadataValidator::validate_cedar_version().` This allows us to give policy authors a very detailed, granular validation error in the .metadata section of their diagnostic report rather than a generic JSON parse failure

jans-cedarling/cedarling/src/init/policy_store.rs
haileyesus2433: Same reason as above

jans-cedarling/cedarling/src/init/policy_store.rs
haileyesus2433: PolicyStoreLoadError is a public enum exposed to users of the cedarling crate, whereas PolicyStoreError (and ValidationError) are  kept pub(crate) to avoid leaking our internal loader/validator types into the public API. If we try to wrap them, the compiler forces us to make them pub too, which i think is best to avoid

jans-cedarling/cedarling/src/init/policy_store.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14673/commits/07744bef25e93187941e23d96a0f1378cbb8492b

jans-cedarling/cedarling/src/init/policy_store.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14673/commits/07744bef25e93187941e23d96a0f1378cbb8492b

jans-cedarling/cedarling/src/common/policy_store/legacy_store/mod.rs
haileyesus2433: removed in https://github.com/JanssenProject/jans/pull/14673/commits/bd872409ccb564dcbb1c7b2fa9ee725ff2719e23 , i will add it back if thats what the agama team is using

haileyesus2433:
> It would be nice to add example commands in the readme for users to test it out or maybe add it in the docs in a separate PR. Also in my opinion we should use just cedarling instead of cedarling_cli for the binary name

yeah i agree `cedarling validate` has more nice ring to it renamed and added examples in here https://github.com/JanssenProject/jans/pull/14673/commits/ef9fb668718120a7cdec0df7ef6ba0d05b418418

olehbozhok:
## Review

Nice addition — closes the author → test → deploy loop for policy authors without needing a full app harness. Structure of `cedarling-cli` (cli/config/authorize/test/validate modules) is clean and easy to follow.

### Strengths

- Clear separation of concerns: `cli.rs` (arg parsing), `config.rs` (bootstrap resolution), `authorize.rs`, `validate.rs`, `test/` (runner + spec).
- Config precedence (file → env → CLI flags) is documented in the README and matches the implementation in `resolve_bootstrap`.
- Integration tests exercise the actual compiled binary via `assert_cmd`, covering pass/fail for `test` and `validate`, plus a missing-config error case — good, since these paths depend on real exit codes.
- Exit code contract (`0`/`1`/`2`) is documented and consistent across subcommands — useful for CI/scripting.
- No panics in the request path; errors flow through `anyhow::Context` and exit cleanly via `unwrap_or_exit`.
- Coverage report (triggered/untriggered policy IDs + warning on unknown reason_ids) is a nice touch for policy authors.

### Concerns

1. **New public API surface on `Cedarling`**: `all_policy_metadata()` and `validate_policy_store()` (in `cedarling/src/lib.rs`) are added as public methods purely to support this CLI tool, but they become part of the stable public API for *every* consumer (Python bindings, UniFFI, etc.), not just `cedarling-cli`. Suggest gating these behind a feature flag (e.g. `tooling`), default-off, enabled only by `cedarling-cli`'s dependency declaration — keeps the primary authorization API surface minimal for other consumers.

2. **Missing test coverage for `authorize` subcommand**: `tests/integration_tests.rs` covers `test` and `validate` but not `authorize`, despite it having its own argument validation logic (e.g. `principal_type`/`principal_id` pairing, JSON parsing of `--principal-attrs`/`--resource-attrs`/`--context`). Worth adding at least one happy-path and one malformed-JSON case.

3. **Minor formatting nit**: [`validate.rs`](jans-cedarling/cedarling-cli/src/validate.rs) has an extra-indented `.await` after the multiline `Cedarling::validate_policy_store(...)` call that doesn't match `rustfmt` output — worth running `cargo fmt` once more before merge.

### Suggestion

Consider feature-gating the two new `Cedarling` methods so this PR doesn't expand the stable public API surface for unrelated consumers (bindings) as a side effect of adding a CLI tool.

dagregi (dismissed):
Nicely done

jans-cedarling/cedarling-cli/src/config.rs
```diff
@@ -0,0 +1,87 @@
...
+        // Clear all conflicting sources before overriding
+        raw_config.local_policy_store = None;
+        raw_config.policy_store_uri = None;
+        raw_config.policy_store_local_fn = None;
+        raw_config.policy_store_cjar_url = None;
+
+        let s = store.to_string_lossy().to_string();
+        if s.starts_with("http://") || s.starts_with("https://") {
+            raw_config.policy_store_uri = Some(s);
+        } else if s.starts_with("cjar://") {
```
olehbozhok: Why we have custom URI schema? Is it used anywhere?

jans-cedarling/cedarling-cli/src/cli.rs
```diff
@@ -0,0 +1,88 @@
...
+    #[arg(long, env = "CEDARLING_CONFIG")]
+    pub config: Option<PathBuf>,
+
+    /// Path or URL to the policy store (overrides config)
+    #[arg(long)]
+    pub policy_store: Option<PathBuf>,
+
+    /// Logger type (e.g., `std_out`, `memory`, `off`)
+    #[arg(long, env = "CEDARLING_LOG_TYPE")]
+    pub log_type: Option<String>,
```
olehbozhok: `log_type` and `log_level` here are `Option<String>` on `CommonArgs`, but both correspond to real enums (`LoggerType`, `LogLevel`). Right now the string is validated twice, at two different points:

1. Never at parse time — clap accepts any string for these flags.
2. In `resolve_bootstrap` (config.rs:63-75), where each one is normalized (`to_lowercase`/`to_uppercase`), serialized back to a JSON string, and fed through `serde_json::from_str` just to reuse the enum's `Deserialize` impl and produce an error message.

That round trip only exists because neither enum implements `FromStr`. If we add it:

#[derive(Debug, thiserror::Error)]
#[error("invalid log type: {0}")]
pub struct ParseLoggerTypeError(String);

impl FromStr for LoggerType {
    type Err = ParseLoggerTypeError;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s.to_lowercase().as_str() {
            "off" => Ok(LoggerType::Off),
            "memory" => Ok(LoggerType::Memory),
            "std_out" | "stdout" => Ok(LoggerType::StdOut),
            _ => Err(ParseLoggerTypeError(s.to_string())),
        }
    }
}


// same idea for LogLevel (FATAL/ERROR/WARN/INFO/DEBUG/TRACE)

then two things change:

- `CommonArgs` can hold the real types directly:

  pub log_type: Option<LoggerType>,
  pub log_level: Option<LogLevel>,

  clap uses `FromStr` automatically for non-`String`/`PathBuf` fields, so invalid values get rejected right at argument parsing, with an error that names the flag — instead of surfacing later out of `resolve_bootstrap`.

- The two blocks in `config.rs` collapse to plain assignment, no more serde_json/anyhow dance:

  if let Some(log_type) = args.log_type {
      raw_config.log_type = log_type;
  }
  if let Some(log_level) = args.log_level {
      raw_config.log_level = log_level;
  }

Net effect: validation happens earlier (fail fast on bad CLI input/env var), `config.rs` loses ~12 lines of round-trip conversion code, and the doc comment on `resolve_bootstrap` claiming "Panics if JSON serialization... fails" goes away too, since there's no more serialization step to panic on.

`application_name` and `policy_store` don't have this problem — they're genuinely string/path values, no enum backing them, so `Option<String>`/`Option<PathBuf>` stays as-is for those.

jans-cedarling/cedarling-cli/src/authorize.rs
```diff
@@ -0,0 +1,84 @@
...
+            };
+            Some(EntityData {
+                cedar_mapping: CedarEntityMapping {
+                    entity_type: typ,
+                    id,
+                },
+                attributes: attrs,
+            })
+        },
+        _ => None,
```
olehbozhok: should we warn user or throw an error?

jans-cedarling/cedarling-cli/src/main.rs
```diff
@@ -0,0 +1,63 @@
+// This software is available under the Apache-2.0 license.
+// See https://www.apache.org/licenses/LICENSE-2.0.txt for full text.
+//
+// Copyright (c) 2024, Gluu, Inc.
+
+use cedarling_cli::cli::{Cli, Command};
+use cedarling_cli::{authorize, config, test, validate};
+use clap::Parser;
+
+fn unwrap_or_exit<T, E: std::fmt::Debug>(result: Result<T, E>) -> T {
```
olehbozhok: Name misleading — no "unwrap" involved (no Option/Result::unwrap() call, no panic on Err). It print err + process::exit(2) on Err path. Rust convention "unwrap" implies panic-on-None/Err.

olehbozhok:
I run the agent in another mode, and here is the result:

# cedarling-cli — Code Review Notes

Strategic/long-term-maintenance review of the new `cedarling-cli` crate (implements
[#14436](https://github.com/JanssenProject/jans/issues/14436) and
[#14437](https://github.com/JanssenProject/jans/issues/14437)).
Findings below are grouped by severity. File:line references point at the reviewed revision.

## Blockers (recommend fixing before merge)

### 1. `src/test/spec.rs` — no `#[serde(deny_unknown_fields)]`
A typo'd optional field (e.g. `raeson_ids` instead of `reason_ids`) parses successfully,
the field silently becomes `None`, and the test reports "ok" while checking nothing.
Verified empirically against a real YAML sample. Required fields already fail loudly
(`missing field`), so the current safety net only exists by accident of which fields
happen to be required.

**Fix:** add `#[serde(deny_unknown_fields)]` to `TestExpected`, `TestRequest`, `TestEntity`.

### 2. `src/test/spec.rs` — `probe_typo_field` test asserts nothing
```rust
match parsed {
    Ok(v) => eprintln!("PROBE_TYPO_OK: {:?}", v),
    Err(e) => eprintln!("PROBE_TYPO_ERR: {}", e),
}
```
Passes regardless of outcome. Reads like leftover exploratory/debug code. Misleading —
`cargo test` reports it as a passing test.

**Fix:** delete it, or replace with a real `assert!`/`expect_err` per the intended behavior.

### 3. Zero integration-test coverage for the `authorize` subcommand
`tests/integration_tests.rs` covers `test` and `validate` only. `authorize` — the
subcommand issue #14436 frames as the primary ad-hoc/CI debug tool — has no black-box
coverage at all: no allow case, no deny case, no malformed-JSON case, no exit-code check.

**Fix:** add at least: allow (exit 0), deny (exit 1), malformed `--context` JSON (exit 2).

### 4. `src/cli.rs` — `--principal-attrs` has no `requires` constraint
`principal_type`/`principal_id` correctly `requires` each other, but `principal_attrs`
does not require either. If a user passes `--principal-attrs '{...}'` alone, clap accepts
it, and `authorize.rs` silently drops the attrs (only read inside the `(Some, Some)` match
arm) — user believes their attributes were sent; they were not.

**Fix:** add `requires = "principal_id"` to `principal_attrs`, or validate explicitly in
`authorize.rs` and error if `principal_attrs` is `Some` while `principal` ends up `None`.

## Strategic (won't hurt today, will hurt in a year)

### 5. `src/validate.rs` — a fully-skipped validation run reports "validation passed"
`ValidationReport::is_ok()` treats `Skipped` as passing. A CI pipeline relying on
`cedarling_cli validate && git push` can go green while validating nothing, with no
signal in the summary line distinguishing "verified" from "mostly skipped."

**Fix:** surface skip count in the summary (`"validation passed (1 level skipped)"`), and/or
add a `--strict` flag that turns skips into failures for CI use.

### 6. Triple duplication of `authorize` args across three files
`cli.rs::Command::Authorize { .. }` → hand-copied field-by-field in `main.rs` →
`authorize::AuthorizeArgs`. Every new flag requires a coordinated edit in 3 places.

**Fix:** derive `clap::Args` on `AuthorizeArgs` directly, use `#[command(flatten)]` /
a tuple variant (`Command::Authorize(AuthorizeArgs)`) to collapse to one source of truth.

### 7. `main.rs::unwrap_or_exit` — misleading name
Doesn't unwrap/panic; prints the error and calls `std::process::exit(2)`. A reader
familiar with `Result::unwrap()` semantics will expect unwind/panic behavior, not a hard
process exit. Rename to e.g. `exit_on_err` or `print_and_exit`.

### 8. `src/config.rs` — case-sensitive config-file extension sniffing
`ext == "yaml" || ext == "yml"` (no `.to_lowercase()`), unlike the equivalent check in
`cedarling/src/bootstrap_config/decode.rs` which does lowercase first. A file named
`Config.YAML` silently falls into the JSON branch and fails with a confusing
"failed to parse config as JSON" error.

**Fix:** lowercase `ext` before comparing, matching the existing library convention.

### 9. `src/config.rs` — no unit tests for the CLI/env/file precedence-merge logic
`resolve_bootstrap` has non-trivial branching (file vs env, YAML/JSON sniffing, policy-store
scheme detection, enum-from-string overrides) with zero test coverage in this file. No
regression net for future contributors adding field #8.

### 10. `src/test/runner.rs` — `break` on first missing `reason_id`
If 3 expected reason_ids are missing, only the first is reported. Forces a slow
fix-one/rerun/repeat loop for policy authors. One-line fix: collect all missing ids into
a single message instead of breaking on the first.

### 11. `src/authorize.rs` — duplicated JSON-parse-or-default pattern (x3)
Same `if let Some(a) = args.X { serde_json::from_str(&a)... } else { default }` shape
repeated for `principal_attrs`, `resource_attrs`, `context`. Also `EntityData` /
`CedarEntityMapping` construction duplicated for principal vs resource — `test/spec.rs`
already has the right idiom via `impl From<TestEntity> for EntityData`, not reused here.

**Fix:** extract a shared `parse_attrs(...)` helper and/or a `build_entity(...)` helper.

## Minor / polish

- `--no-color` doesn't honor the `NO_COLOR` env-var convention (no-color.org).
- `--log-level` help text omits `FATAL`/`TRACE` (both valid `LogLevel` variants).
- `validate.rs` dot-padding assumes level names < 24 chars (`saturating_sub`); a longer
  future level name silently misaligns output instead of erroring.
- Test fixtures (`test_store.yaml`, `tests.yaml`, etc.) live flat in `tests/`, not
  `tests/fixtures/` — fine now, will get cluttered as scenarios grow.
- No `--format json` escape hatch for `validate`/`test` output — underlying report types
  are already `Serialize`, so this would be close to free to add later; worth a follow-up
  issue rather than blocking this PR.
- `config.rs`: `log_type` and `log_level` overrides use two different mechanisms
  (JSON round-trip vs a to-be-added `FromStr`) for the same "enum from CLI string"
  problem — asymmetry future contributors may copy incorrectly.
- `lib.rs` exposes a fully `pub` module tree with no doc comment stating whether that's
  intentional API surface or just plumbing for `main.rs`.

## What's good (worth preserving, not regressing)

- Exit-code convention (0 success / 1 command-level failure / 2 infra error) is
  consistent across all three subcommands and actually tested end-to-end for `validate`.
- `validate.rs::describe_source` match has no wildcard arm — a new `PolicyStoreSource`
  variant fails to compile here rather than silently mishandling it.
- `config.rs`'s mutually-exclusive policy-store-source clearing before override (lines
  ~46-49) correctly matches the downstream validation in `decode.rs` — the hardest part
  of that function to get right, and it's right.
- Coverage-report sorting (`triggered`/`untriggered`/`extra_ids`) keeps CLI output
  deterministic across runs — good for anyone diffing CI logs.


resolved in https://github.com/JanssenProject/jans/pull/14673/commits/a9bc7617de11178d8e5d27e67b4c5cfa455c12f3

jans-cedarling/cedarling-cli/src/config.rs
haileyesus2433: the pseudo-schema is actually just a CLI-level affordance we use to explicitly force a URL to be processed as a CJAR archive (PolicyStoreSource::CjarUrl instead of PolicyStoreSource::Uri). Since the CLI multiplexes all of these into a single policy-store flag (which can take a local filepath, an http(s):// URL, etc.), the pseudochema was an easy way to disambiguate the  intent without having to add a completely separate --policy-store-cjar-url CLI flag

jans-cedarling/cedarling-cli/src/cli.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14673/commits/536a777df4301284dd3205541bcf34b60c2a244a

jans-cedarling/cedarling-cli/src/authorize.rs
haileyesus2433: resolved in https://github.com/JanssenProject/jans/pull/14673/commits/265bf392c9e077ddd4212d8a083dbebd0b3cf325

jans-cedarling/cedarling-cli/src/main.rs
haileyesus2433: renamed in https://github.com/JanssenProject/jans/pull/14673/commits/8e0daf3d694dba63b23f0d125e4a8592eb132e53

haileyesus2433:
@coderabbitai approve

dagregi (dismissed):

olehbozhok:
@haileyesus2433  you ignored suggestion 1 from https://github.com/JanssenProject/jans/pull/14673#issuecomment-5193227686

jans-cedarling/cedarling-cli/src/test/runner.rs
```diff
@@ -0,0 +1,195 @@
...
+                    if expected_err_count != actual_err_count {
+                        test_passed = false;
+                        fail_reasons.push(format!(
+                            "expected {expected_err_count} errors, got {actual_err_count}"
+                        ));
+                    }
+                }
+
+                if test_passed {
+                    println!("{}", "ok".green());
```
olehbozhok, line 116: looks like you didn't implement respect of CLI flag `no_color`

jans-cedarling/cedarling-cli/src/config.rs
olehbozhok: I feel like we don't need to add something artificial, only what works in the CLI

jans-cedarling/cedarling-cli/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "cedarling-cli"
+version = "0.0.0"
+edition = "2024"
+publish = false
+license = "Apache-2.0"
+
+[[bin]]
+name = "cedarling"
```
olehbozhok: maybe better to call `cedarling_cli` or `cedarling-cli`?

haileyesus2433: i think cedarling is better it was cedarling-cli before, there was even a discusssion in here https://github.com/JanssenProject/jans/pull/14673#issuecomment-5189598356

jans-cedarling/cedarling-cli/src/test/runner.rs
haileyesus2433: The --no-color flag is  handled globally in main.rs via colored::control::set_override(false). This tells the colored crate to automatically strip ANSI escapes from all .green() and .red() calls across the entire codebase when the flag is passed

olehbozhok: I see, ok, thanks!

jans-cedarling/cedarling-cli/Cargo.toml
olehbozhok: I think it leads to misleading. Better to have an unambiguous name

haileyesus2433: ok i will revert it to cedarling-cli

haileyesus2433: renamed in here https://github.com/JanssenProject/jans/pull/14673/commits/5ff67ed77805977bc0b86b6811e6ab4b1e5657b3

olehbozhok (approved):

jans-cedarling/cedarling-cli/Cargo.toml
dagregi: In my opinion running `cedarling test --policy-store ./my_store.yaml my_tests.yaml` is better than using `cedarling-cli test --policy-store ./my_store.yaml my_tests.yaml`

haileyesus2433:
@coderabbitai approve

jans-cedarling/cedarling-cli/Cargo.toml
dagregi: And if we are going to revert back to cedarling-cli the readme should be updated too

Issue #14803, opened 2026-08-19
feat(jans-auth-server): populate forwarded client-context headers at the FIDO2 call sites


imran-ishaq:
**Is your feature request related to a problem? Please describe.**
[#14801](https://github.com/JanssenProject/jans/issues/14801) makes the FIDO2 server accept X-Jans-Client-IP and X-Jans-Client-User-Agent and prefer them over getRemoteAddr() / User-Agent when recording metrics. That change is inert on its own: nothing sends the headers, so ipAddress stays the Auth Server's own address and userAgent stays Apache-HttpClient/4.5.14 (Java/17.0.20) until the callers are updated.

This issue covers the sending half. It is separate from [#14801](https://github.com/JanssenProject/jans/issues/14801) because the callers live in modules that build against the published jans-fido2-client artifact — they cannot compile against the new overloads until [#14801](https://github.com/JanssenProject/jans/issues/14801) has merged and republished, so the two cannot land in one PR.

**Describe the solution you'd like**
Pass the browser's IP and user agent on every Fido2ClientFactory proxy call that has the end user's request in hand, using the overloads added in [#14801](https://github.com/JanssenProject/jans/issues/14801):

1. jans-auth-server/server/src/main/webapp/auth/fido2/passkeys.xhtml and its backing bean — the path the Passkey Security Monitor's data actually comes from, so this is the one that visibly fixes the dashboard.
2. jans-casa/agama/project/lib/io/jans/casa/authn/FidoValidator.java — three call sites (lines 31, 54, 76).
3. docs/script-catalog/person_authentication/fido2-external-authenticator/Fido2ExternalAuthenticator.py — roughly ten authenticate / verify call sites across assertion and attestation.

Where a caller has no access to the originating HttpServletRequest, it should send nothing rather than a placeholder — [#14801](https://github.com/JanssenProject/jans/issues/14801) falls back to current behaviour when the headers are absent, and a fabricated value is worse than a missing one.

**Describe alternatives you've considered**
- Fold this into [#14801](https://github.com/JanssenProject/jans/issues/14801). Not possible in one PR: cross-module CI builds each module against published sibling jars, so jans-auth-server and jans-casa would be compiling against a jans-fido2-client that does not yet have the overloads.
- Update only passkeys.xhtml and stop there. It is the highest-value call site and would fix the dashboard, but it would leave Casa and the person-auth script silently recording the Auth Server as the client — the same wrong data, now inconsistently.
[...]

PR #14805, opened 2026-08-19
feat(jans-fido2): honour forwarded end-user IP and user agent in passkey metrics

### LINKED ISSUE #14801: feat(jans-fido2): propagate end-user IP and user agent to FIDO2 so metrics describe the user, not the calling service

Every `ipAddress`, `userAgent`, `browser`, `os` and `deviceType` value in `jansFido2MetricsEntry` describes the Auth Server, not the end user. Sample rows from `admin-ui-test.gluu.org`:

| Field | Recorded value | Should be |
| --- | --- | --- |
| `ipAddress` | the VM's own address | the browser's public IP |
| `userAgent` | `Apache-HttpClient/4.5.14 (Java/17.0.20)` | the real browser UA |
| `browser` / `os` | `Unknown` / `Unknown` | e.g. `Safari` / `macOS` |
| `deviceType` | `DESKTOP` on every row | `MOBILE` / `DESKTOP` as applicable |

The cause is architectural: the browser never talks to the FIDO2 server. It calls the Auth Server, which relays to `/attestation/*` and `/assertion/*` through `Fido2ClientFactory` — a plain RESTEasy/Apache HTTP client with no header propagation (`jans-fido2/client/.../Fido2ClientFactory.java:77-88`). The proxy interfaces it builds declare no header parameters at all (`AssertionService.java:24-38`, `AttestationService.java`). So by the time a request reaches `MetricService`, the only client it can observe is the Auth Server itself.

`MetricService.extractIpAddress` (`MetricService.java:730-771`) does the right thing for a browser-facing service — it checks `X-Forwarded-For` plus eight other proxy headers before falling back to `getRemoteAddr()`. Nothing sets them here, because the caller is a service-to-service client, not a reverse proxy.

The `deviceType` case is worse than merely missing. `DeviceInfoExtractor.determineDeviceType` (`DeviceInfoExtractor.java:171-181`) matches `mobile`/`android`/`iphone` → `MOBILE`, `tablet`/`ipad` → `TABLET`, and **everything else falls through to `DESKTOP`**. There is no unknown branch. So every service-to-service call is silently recorded as a desktop, and the resulting chart looks plausible rather than empty — the failure mode is fabricated data, not absent data.

Downstream consumers of these fields are therefore wrong today: `deviceTypes`, `browsers` and `operatingSystems` in `/metrics/analytics` (`Fido2MetricsService.java:437-470`), `DEVICE_TYPES` in the aggregation rollup (`:758-764`), and the device/IP panels in the Flex Admin UI Passkey Security Monitor (GluuFederation/flex#2905). The "Suspicious IPs" graph had to be dropped from that dashboard because the IP column carries no usable signal.

Not affected, and safe to keep graphing: usernames, statuses, counts, durations, and `authenticatorType` (platform / cross-platform) — the last of these is read from the credential itself (`AttestationService.java:323-344`, `AssertionService.java:423`), not from the user agent.

Pass the end user's connection context across the Auth Server → FIDO2 boundary and prefer it when present.

1. Add dedicated headers to the client proxy interfaces — `X-Jans-Client-IP` and `X-Jans-Client-User-Agent` — as `@HeaderParam` arguments on `AssertionService.authenticate/verify` and the `AttestationService` equivalents, with `Fido2ClientFactory` unchanged.
2. In `MetricService.extractIpAddress` and `DeviceInfoExtractor.extractDeviceInfo`, prefer the new headers over `X-Forwarded-For`/`getRemoteAddr()`/`User-Agent`, and only trust them from a configured set of callers — the spoofing caveat already documented at `MetricService.java:717-726` applies with more force once a header is authoritative.
3. Independently of the above, change `determineDeviceType` to return `UNKNOWN` when the user agent matches no known pattern, instead of defaulting to `DESKTOP`. Absence should read as absence. This is worth doing on its own even if the propagation design changes, since it converts a wrong chart into an honestly empty one.

Backwards compatibility: headers are optional, so existing callers keep working and simply continue producing the current values.

- **Do nothing and drop the affected graphs.** This is what the Admin UI did as a stopgap — Suspicious IPs was replaced with a per-user attack graph. It works, but it permanently forfeits geo-anomaly and impossible-travel detection, which are the main security value of passkey metrics.
- **Have the browser POST device context straight to FIDO2.** Requires a new unauthenticated public endpoint plus CORS, and the payload would be entirely client-controlled — strictly worse than a trusted server-side hop.
- **Look the context up from the session.** FIDO2 already receives `sessionId`, and the Auth Server has recorded the IP and UA against that session. No new headers, and not spoofable by the client. Rejected as the primary approach because it couples FIDO2 to Auth Server session storage and yields nothing for direct API callers, but it is a reasonable fallback for the cases where no header is supplied.
- **Reuse `X-Forwarded-For` rather than a dedicated header.** Zero server-side change, since `extractIpAddress` already reads it. Rejected because it conflates a genuine proxy hop with an application-level relay, and would make a real reverse-proxy deployment ambiguous.

**Additional context**

- Same trust boundary as [#14744](https://github.com/JanssenProject/jans/issues/14744), and the two should be designed together. That issue needs the browser's error name and elapsed time to reach FIDO2, and rejects a direct browser→FIDO2 call in favour of relaying through the Authorization Server; this issue needs the browser's IP and user agent to reach FIDO2 over the same hop. Both are "client context the FIDO2 server structurally cannot observe", and a single decision on how the Auth Server forwards that context should settle both. They differ in one respect worth keeping straight: [#14744](https://github.com/JanssenProject/jans/issues/14744) relays a report the client volunteers after a ceremony ends, so it is untrusted annotation, whereas the IP and user agent here are observed by the Auth Server on the live request and are as trustworthy as the Auth Server's own view of the client.
- Found by @arnab-dutta while validating the Passkey Security Monitor dashboard (GluuFederation/flex#2905) against live data on `admin-ui-test.gluu.org`.
- Until this lands, treat `ipAddress` and the device/browser/OS breakdowns as unusable, and note that `deviceType` is actively misleading rather than blank.
- Needs a design call on header naming and the trusted-caller list before implementation.
- Scope note: this spans `jans-fido2` (the client proxy interfaces, `MetricService`, `DeviceInfoExtractor`) and the callers that hold the browser's request — the person-authentication script, Casa, and `jans-auth-server`'s `passkeys.xhtml`. Step 4 (`determineDeviceType` returning `UNKNOWN` instead of defaulting to `DESKTOP`) is self-contained inside `jans-fido2` and needs none of the above design decision; it can be split out and landed first if that is preferable.

imran-ishaq:
### Prepare

-------------------

#### Target issue
Server half only. Populating the headers at the call sites — the person-authentication script, Casa
and `passkeys.xhtml` — is #14803, and cannot be in this PR: those modules build against the
published `jans-fido2-client` artifact, so they cannot compile against the new overloads until this
merges and republishes.
  closes #14801 

#### Implementation Details
  
The FIDO2 endpoints are never called by the browser. The Authorization Server, Casa and the
person-authentication script relay to them over `Fido2ClientFactory`, which propagates no headers, so
the only client the server can observe is the calling service. That is why `ipAddress` is the VM's
own address and `userAgent` is `Apache-HttpClient/4.5.14 (Java/17.0.20)` on every row.

**Carrying the end user's details across the hop.** `ClientContextHeaders` (new, in `model` so client
and server share one definition) names `X-Jans-Client-IP` and `X-Jans-Client-User-Agent`.
`AssertionService` and `AttestationService` gain **overloads** taking them as `@HeaderParam`. The
existing signatures are untouched deliberately: changing them would force Casa, the person
authentication script and the jans-chip demo to move in lockstep with this merge, whereas overloads
let each migrate on its own schedule. Passing null omits the header.

**Deciding whether to believe them.** A header is a claim, so `MetricService` honours these only from
a caller listed in the new `fido2TrustedClientContextSources`, which is **empty by default** — the
headers are ignored until an operator names a caller. A forwarded address is still run through
`isValidIpAddress` before use, and a malformed one falls back to the request's own address rather
than being stored. `"*"` trusts any caller, for deployments where callers have no stable address.

**Stopping the device breakdown being fabricated.** `DeviceInfoExtractor.determineDeviceType`
previously matched mobile and tablet markers and let *everything else* fall through to `DESKTOP`,
with no unknown branch — so every service-to-service call was recorded as a desktop and the chart
looked plausible rather than empty. It now tests for desktop markers explicitly and reports `UNKNOWN`
otherwise. The mobile test still runs first, so Android's shared Linux marker does not reclassify
phones as desktops. This is independent of the trust question above and worth having regardless.

There is also a new `extractDeviceInfo(String)` overload, since the user agent to parse is now often
not the one on the request.

-------------------

**Tests** — full `jans-fido2` suite green, 411 tests, 14 of them new:

- `DeviceInfoExtractorTest` (new, 9 tests) — the service-to-service user agent now reporting
  `UNKNOWN` instead of `DESKTOP`, real desktop user agents still `DESKTOP`, Android not
  reclassified, null user agent leaving fields unset.
- `MetricServiceTest` (5 added) — forwarded headers honoured from a trusted caller, ignored when
  unconfigured, ignored from an untrusted caller, wildcard trust, and a malformed forwarded address
  falling back to the request's own.

**Static analysis** — SpotBugs run locally on `model`, `client` and `server`. This change adds two
findings, `EI_EXPOSE_REP` / `EI_EXPOSE_REP2` on the new `List` getter and setter, which is the same
finding every other collection property in `AppConfiguration` already carries; left consistent with
the surrounding bean rather than making one property defensively copy. The module has 121
pre-existing findings of that class and the plugin is declared only in `pluginManagement`, so it does
not gate the build.

**Docs** — updated in a separate `docs:` commit: `passkey-telemetry.md` (a "Whose device is recorded"
section explaining the relay, the new property and the `UNKNOWN` device type),
`fido2-server-properties-config.md` and the generated `fido2-properties.md`.

### DIFF
--- a/docs/janssen-server/fido/fido2-server-properties-config.md
+++ b/docs/janssen-server/fido/fido2-server-properties-config.md
@@ -32,6 +32,7 @@ The following properties represent the dynamic configuration for the Janssen FID
 | `fido2MetricsAggregationInterval` | `60` | Interval in minutes driving the passkey metrics aggregation scheduler (default `60` = hourly). |
 | `fido2MetricsRetentionDays` | `90` | Retention period in days for passkey metric entries and aggregations before automatic cleanup. |
 | `fido2DeviceInfoCollection` | `true` | Whether device info (browser, OS, device type) is collected and stored with passkey metrics. |
+| `fido2TrustedClientContextSources` | `[]` | Callers whose `X-Jans-Client-IP` and `X-Jans-Client-User-Agent` headers are honoured, listed by remote address. Empty (the default) ignores the headers. See [Passkey Telemetry & Metrics](passkey-telemetry.md). |
 | `fido2ErrorCategorization` | `true` | Whether passkey operation failures are categorized for the error-analysis endpoint. |
 | `fido2PerformanceMetrics` | `true` | Whether passkey operation durations are tracked for performance analytics. |
 | `fido2Configuration` | Object | Nested object containing FIDO2 protocol-specific details (see structure below). |
--- a/docs/janssen-server/fido/passkey-telemetry.md
+++ b/docs/janssen-server/fido/passkey-telemetry.md
@@ -139,9 +139,9 @@ Beyond the outcome itself, each raw entry records where the operation came from:
 
 | Field | Source |
 |---|---|
-| `ipAddress` | First valid address from `X-Forwarded-For` and the other common proxy headers, otherwise the socket remote address. |
-| `userAgent` | The `User-Agent` request header, up to 512 characters. |
-| `deviceInfo` | Browser, OS and device type parsed from the user agent, plus a copy of the user agent itself. The only field `fido2DeviceInfoCollection` suppresses — every other field here is written regardless. |
+| `ipAddress` | The `X-Jans-Client-IP` header when it comes from a trusted caller, otherwise the first valid address from `X-Forwarded-For` and the other common proxy headers, otherwise the socket remote address. |
+| `userAgent` | The `X-Jans-Client-User-Agent` header when it comes from a trusted caller, otherwise the `User-Agent` request header. Up to 512 characters. |
+| `deviceInfo` | Browser, OS and device type parsed from the user agent above, plus a copy of the user agent itself. The only field `fido2DeviceInfoCollection` suppresses — every other field here is written regardless. |
 | `sessionId` | The `session_id` cookie set by the Authorization Server, falling back to the servlet session when one exists. Empty for requests that carry neither. |
 | `metricType` | The metric name of the event, e.g. `fido2_registration_success`. |
 | `nodeId` | Identifier of the cluster node that served the request. |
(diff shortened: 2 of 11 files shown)


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/cda2924b-0aaa-45ec-86c8-496a4ffd44b0?refs=f2931def0531928fe691912a280f4838bbb65981&source=prChecksComment) |

PR #14833, opened 2026-08-20
feat(jans-cedarling): qualify JavaScript SDK distribution

Dahkenangnon:
### Prepare

  -------------------

  ### Description

  #### Target issue

  Part of #14582

  This PR completes JavaScript SDK distribution and runtime qualification. The retained demo integration will be submitted separately, so this PR should not close the parent issue.

  #### Implementation Details

  - Adds CommonJS support alongside the existing ESM package root without introducing a new public runtime entry or changing the SDK API.
  - Adds focused build cleanup and coordinated private package staging.
  - Verifies explicit staged-version handling and installed ESM/CommonJS package behavior through a clean consumer.
  - Exercises both ESM and CommonJS package resolution using real authorization.
  - Reuses a portable QUnit contract subset across Node.js, Bun, Deno, Chromium, Firefox, and WebKit.
  - Adds focused coverage for:
    - multi-issuer request validation and input detachment;
    - initialization failure recovery and generated-protocol validation;
    - idempotent shutdown and closed-client behavior across every public service;
    - valid RS256 authorization and tampered-signature rejection.
  - Extends the existing `cedarling_js_tests` CI job with:
    - Node.js unit and contract checks;
    - installed ESM/CommonJS package verification;
    - Bun and Deno portable contracts;
    - Chromium, Firefox, and WebKit browser contracts.
  - Rewrites the package README as consumer-only documentation.
  - Adds a separate maintainer guide for source builds, testing, CI, packaging, and internal architecture.
  - Keeps registry publication, credentials, provenance, and release promotion outside this change.
  - Does not change the existing public authorization methods or production `src/**` implementation.

  -------------------

  - [x] Relevant documentation has been updated if any (i.e. user guides, installation and configuration guides,
  technical design docs etc)

  Please check the below before submitting your PR. The PR will not be merged if there are no commits that start with
  `docs:` to indicate documentation changes or if the below checklist is not selected.

## Summary by CodeRabbit

- **New Features**
  - Added CommonJS and edge-compatible package entry points alongside ESM.
  - Expanded support across Node.js, Bun, Deno, Chromium, Firefox, WebKit, and Firefox ESR.
  - Added signed JWT validation, multi-issuer authorization, and embedded WebAssembly support.
  - Improved lifecycle handling with idempotent shutdown and consistent client-closed errors.

- **Bug Fixes**
  - Strengthened policy-source and context-store TTL validation.
  - Improved compatibility with cross-realm byte arrays and URL-like values.

- **Documentation**
  - Added comprehensive SDK usage, configuration, security, and maintainer guidance.

- **Testing**
  - Expanded coverage for initialization, packaging, consumer compatibility, portable runtimes, and browsers.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/31301cc8-d8cf-457a-a915-d6ab29e1681b?refs=8ee2c1c1ce200f2036ac37d37fbba731914ca52c&source=prChecksComment) |

Dahkenangnon:
@coderabbitai Please can you as last commit, please review and resolve previous comments if they are effectively fixed with the last commits and approve the PR if everything is Ok. Unless, provide other clear, verifiable feedback

olehbozhok:
## What this PR does

Promotes `cedarling_js` from a Node-only, private, single-command package into a full distribution-qualification setup:

- **CommonJS output** — esbuild bundles `dist/entries/node.js` → `dist/cjs/node.cjs`; `require` conditions added to the export map.
- **Cross-runtime qualification** — new `portable` contract group (4 suites, no Node-specific fixtures) executed under Bun, Deno, and Chromium/Firefox/WebKit via Playwright + an ephemeral static server.
- **Packaging** — `clean.mjs`, `stage-packages.mjs` (coordinated exact-version SDK + wasm tarballs, exact-dependency enforcement), `verify-consumer.mjs` (offline install into a temp consumer, real authorization through both ESM and CJS, duplicate-wasm-install check).
- **CI** — Bun/Deno/Playwright steps in `cedarling_js_tests`.
- **Docs** — README becomes the consumer guide; new `docs/readme.md` is the maintainer guide.
- **Tests** — lifecycle, RS256 signature validation, `authorizeMultiIssuer` unit, engine-initialization boundary.

Direction and structure look good: portable suites are shared rather than duplicated per runtime, the staging script refuses non-exact versions, and `verify-consumer` exercises a genuinely installed package instead of the source tree. Comments below are about robustness of the gate itself.

---

### 1. CI steps depend on implicit state from earlier steps

`.github/workflows/test-cedarling.yml`:

- `Verify staged ... packages` runs `node scripts/verify-consumer.mjs` directly — it needs `dist/`, produced by the preceding `npm run test:prepare`.
- `Run ... browser contracts` runs `npm run test:browser:run` — it needs `.test-dist/runners/browser.js`, produced by the `test:prepare` inside the *Deno* step.

Reordering, removing, or making the Deno step conditional silently breaks the browser step. Suggest `npm run package:verify` and `npm run test:browser` (both self-preparing), or one explicit prepare step with the rest documented as dependent.

### 2. Declared minimum Node is never tested

`engines.node` is `>=20.19`, CI runs only `node-version: '22.x'`. For a PR whose purpose is distribution qualification, the floor should be in the matrix (20.19 + 22). The CJS bundle and `createRequire`/dynamic-`import()` path are exactly where a 20.x/22.x difference would show up.

### 3. Staged manifest has no legacy resolution entry points

`stagedSdkManifest` carries `exports` only — no top-level `main` or `types`. Consumers on `moduleResolution: node`/`node10`, or older bundlers, resolve nothing and get no types. Either add `main`/`types` pointing at the CJS/`.d.ts`, or state the Node16/bundler resolution requirement explicitly in the README.

### 4. Two assertions do not test what they appear to test

**`tests/contract/signature-validation.test.ts`**

```ts
const tampered = await authorize(tamperSignature(token));
assert.true(
  !tampered.ok || !tampered.value.decision,
  "the tampered signature cannot authorize",
);
```

This passes for *any* failure — wrong mapping name, issuer server hiccup, a policy that never permits. It cannot distinguish "signature rejected" from "test broken". Assert the specific outcome (expected `error.code`, or `ok` with `decision === false` plus a signature-related diagnostic).

Same file: `assert.true(valid.ok && valid.value.decision, ...)` discards `valid.error` on failure — worth surfacing the code in the message.

**`tests/unit/engine-initialization.test.ts`**

```ts
assert.false(JSON.stringify(first).includes(secret));
```

`createSdkError` builds errors via `Object.assign(new Error(message), {...})` (`src/errors/errors.ts:135`), so `message` is **non-enumerable** and never appears in `JSON.stringify`. The raw cause is held in a `WeakMap`, also invisible. The assertion therefore cannot fail for the leak it is guarding against. Assert against `first.message`, `String(first)`, and the absence of `cause` explicitly. (The README itself notes that standard `Error` properties are not all enumerable — the test contradicts that note.)

### 5. CJS bundle: real constraint worth documenting

`src/engine/node.ts` loads generated glue via `await import(pathToFileURL(...).href)`. esbuild keeps that as a live `import()` in the CJS output (default `target: esnext`), so it works on Node. It breaks if a downstream tool re-transpiles the shipped `.cjs` to an older CJS target and rewrites `import()` → `require()` of an ESM module. Worth one line in the maintainer guide.

Related: `--log-override:empty-import-meta=silent` is currently safe only because of the `typeof __filename !== "undefined"` guard at `src/engine/node.ts:37`. Silencing the whole rule means a future `import.meta` use degrades to `{}` with no warning. A scoped comment next to the flag, or a build-time assertion that `dist/cjs/node.cjs` contains no `import_meta`, would keep that guarantee.

Minor: `--external:@janssenproject/cedarling_wasm` is close to a no-op here — the Node entry has no static import of the generated package (the re-export is type-only). Harmless, but it reads as load-bearing.

### 6. Browser runner robustness

`tests/runners/browser-page.ts`:

- `expect(diagnostics).toEqual([])` fails on **any** console error, including browser-emitted network errors. A stray `/favicon.ico` request against the fixture server hits the 404 branch; if the browser surfaces that as a console error, the assertion fails for reasons unrelated to the SDK. Cheap hardening: serve `204` for `/favicon.ico` and restrict `diagnostics` to `pageerror` plus console errors originating from the page.
- The wasm fetch works because the bundle is emitted flat at `.test-dist/.build/run-browser.js` and served at `/`, so wasm-bindgen's `new URL('cedarling_wasm_bg.wasm', import.meta.url)` lands on `/cedarling_wasm_bg.wasm`. Moving the esbuild `--outfile` into a subdirectory breaks it with a 404 and no obvious cause. A comment at the server, or serving the wasm from the same relative path as the script, would make the coupling explicit.
- `waitForFunction` at 90s inside a 120s test timeout: WebKit on a cold CI runner compiling the module is the slow case here. Watch for flakes.

### 7. `npm run check` no longer matches the CI gate

`check` = typecheck + prepare + node unit + node contract + consumer verify. Bun, Deno, and browsers exist only in CI. Contributors get green locally and red on the PR. `docs/readme.md` says this, the README does not — worth a line there too, or a `check:all` target that runs everything.

### 8. README documents an install that cannot succeed

The README now opens with `npm install @janssenproject/cedarling`, but the package is `private: true` and the staged manifests deliberately stay private. Anyone following the quick start fails at step one. Add a "not yet published" banner.

Also missing for the browser story the PR is qualifying: consumers bundling for the browser must arrange for `cedarling_wasm_bg.wasm` to be emitted as an asset next to their bundle. That is the single most likely integration failure, and it is documented nowhere.

### 9. Dependency pinning is now inconsistent

`@playwright/test: "^1.61.0"` and `esbuild: "^0.28.1"` use carets; every pre-existing devDep is exact (`typescript 7.0.2`, `qunit 2.26.0`, `@types/node 24.13.3`). `npm ci` protects CI via the lockfile, but a lock refresh silently changes the browser builds the qualification gate runs against. Pin both.

### 10. Job cost

`test:prepare` (clean + two full `tsc` passes) runs three times per job — once in the Node step, again inside `test:portable:bun`, again inside `test:portable:deno`. Plus `playwright install --with-deps chromium firefox webkit` on every run with no browser cache. This lands on every PR touching cedarling. Caching `~/.cache/ms-playwright` keyed on the resolved Playwright version, and preparing once, would recover most of it.

### Nits

- `tests/unit/index.ts`: `engine-initialization.test.js` appended after `options.test.js`, breaking the alphabetical order the rest of the list (and the correctly-placed `authorize-multi-issuer`) follows.
- `verify-consumer.mjs` uses `npm install --offline` with a fresh cache. Correct today only because the generated wasm package has no dependencies; adding one turns this into a confusing offline-resolution failure. A comment would save the next person the debug session.

moabu (dismissed):

Dahkenangnon:
Hi @olehbozhok .

Thank you for the detailed review. I verified every item against the SDK, its
installed package, and the relevant runtime restrictions. Items 5A–5C, 6B,
8B, and Nit B shared one root cause: the published root still depended on
locating generated glue and a separate WASM file. I resolved that boundary
instead of documenting workarounds: root ESM and CommonJS are now
self-contained, while `./edge` supplies the static precompiled-module import
required by restricted edge hosts. The public API is unchanged.

## 1. Implicit CI preparation — accepted

[CI][workflow] now prepares production, test, and browser artifacts once;
Node, package verification, Bun, Deno, Playwright, and Firefox ESR consume that
same output. Local npm commands remain self-preparing. This intentionally uses
one job: GitHub documents that later steps can access changes in
`GITHUB_WORKSPACE`, whereas each hosted job starts on a fresh runner.

References: [GitHub runner workspaces][github-workspace] and
[workflow jobs][github-jobs].

## 2. Runtime versions — accepted

The [manifest][manifest] now requires Node `>=22`. CI runs the full gate on
Node 22 and installed ESM/CommonJS consumers on Node 22, 24, and 26. This
matches Node's current table: 22/24 are LTS, 26 is Current, and 20 is EOL.
Deno runs its `lts` and `latest` channels without
`--unstable-sloppy-imports`; Deno documents that flag as unstable support for
extensionless imports. Bun is pinned to stable 1.3.13 because Bun publishes
stable/canary channels, not a Node-like LTS policy. Browser coverage uses the
Chromium/Firefox/WebKit builds coupled to pinned Playwright 1.62.1 plus Firefox
ESR; Playwright WebKit is not described as Safari.

References: [Node releases][node-releases], [Deno channels][deno-releases],
[Bun channels][bun-stable], and [Playwright browsers][playwright-browsers].

## 3. Legacy resolution — partially accepted

I added top-level `types` to the source and staged manifests; package
verification confirms that the declaration is present. I did not add `main`:
supported Node versions
understand `exports`, which represents the ESM/CommonJS conditions and
`./edge`; `main` represents only one entry. The documented TypeScript boundary
is `node16`, `nodenext`, or `bundler`; TypeScript states that `node10` does not
support package `exports` and should no longer be used.

Code: [manifest][manifest], [staging][stage], [consumer verification][verify].
References: [Node package entry points][node-exports] and
[TypeScript module resolution][typescript-resolution].

## 4. Test assertions — accepted

- **4A — tampered signature:** the test now requires
  `AUTHORIZATION_FAILED` from `authorizeMultiIssuer`, while the valid-token
  control proves that issuer, mapping, request, and policy setup work.
- **4B — valid-token diagnostics:** a failed positive control now reports its
  stable Cedarling error code and stops the tampering check.
- **4C — redaction:** the test retains the JSON check and also checks
  `message`, `String(error)`, `details`, and absence of a public `cause`.

Code: [signature contract][signature-test] and
[initialization test][initialization-test].

## 5. CommonJS and generated-package loading — accepted

- **5A — dynamic import:** the old Node path loader was removed. CommonJS is
  built from the self-contained root ESM, so neither published root artifact
  dynamically imports generated glue.
- **5B — suppressed `import.meta`:** the suppression is gone. The build rejects
  root output containing `import.meta`, Node built-ins, generated-package
  imports, or any remaining output import.
- **5C — redundant external:** the external flag is gone. The generated package
  is an exact dev dependency and installed-consumer verification rejects its
  presence at runtime.

The generated package remains a build input produced by the existing
`wasm-pack --target web` command; no Rust source change or extra generated
target is needed. esbuild's documented binary loader embeds the WASM bytes as
a runtime `Uint8Array`.

Code: [build][build], [embedded loader][embedded], [verification][verify].
References: [wasm-bindgen deployment targets][wasm-bindgen] and
[esbuild's binary loader][esbuild-binary].

## 6. Browser robustness — accepted

- **6A — favicon:** the fixture returns `204` for `/favicon.ico` and still
  records every console and page error.
- **6B — asset coupling:** the root browser build embeds WASM, and the browser
  contract asserts zero `.wasm` requests. The former output-path coupling no
  longer exists.
- **6C — timeout:** the result wait is 120 seconds inside a 150-second test,
  leaving bounded time for navigation, assertions, and cleanup.

`./edge` intentionally differs: it retains exactly one static
`.wasm?module` import and no byte-compilation path. Cloudflare documents that
such imports produce `WebAssembly.Module`; Vercel requires imported WASM rather
than runtime compilation from bytes. The build also fails if the expected
wasm-bindgen glue pattern changes; Cloudflare explicitly documents that this
glue may need patching for Workers.

Code: [browser runner][browser-runner], [Playwright config][playwright-config],
[build][build], [edge loader][edge]. References:
[Cloudflare WASM modules][cloudflare-modules],
[Cloudflare wasm-bindgen guidance][cloudflare-rust], and
[Vercel Edge restrictions][vercel-edge].

## 7. Local versus CI gate — accepted

`npm run check` remains the Node/package gate; `npm run check:all` is the
explicit full local Bun, Deno, and browser gate. Both are documented in the
[maintainer guide][maintainer]; the [consumer README][consumer-readme] remains
usage-focused.

## 8. Consumer installation and WASM handling — accepted

- **8A — unavailable registry install:** the non-working registry command was
  removed; installation guidance will return when a supported channel exists.
- **8B — browser WASM:** root consumers no longer copy, locate, or configure an
  internal WASM asset. The edge package contains the asset required by its
  host-managed static import.

Installed-package probes completed real initialization, authorization, and
shutdown after esbuild 0.28.2, Vite 8.2.2/Rolldown, webpack 5.109.2, and
Next.js 16.3.2/Turbopack builds. These are recorded qualification results, not
a claim about every bundler; Vite confirms that Vite 8 uses Rolldown, so this is
not standalone Rollup coverage.

Code: [consumer README][consumer-readme], [maintainer guide][maintainer],
[build][build]. Reference: [Vite 8/Rolldown][vite-rolldown].

## 9. Dependency pinning — accepted

Playwright `1.62.1` and esbuild `0.28.2` are exact in the
[manifest][manifest] and [lockfile][lockfile]. CI uses `npm ci`, which npm
documents as a
frozen install that fails when the manifest and lockfile disagree.

Reference: [`npm ci`][npm-ci].

## 10. CI cost — accepted

The workflow now prepares shared outputs once. I did not cache Playwright
browsers because Playwright explicitly advises against it when restoration
cost is comparable to downloading and Linux system dependencies remain
uncacheable.

Reference: [Playwright CI guidance][playwright-ci].

## Nits — accepted

- **Nit A:** the complete [unit-suite list][unit-index] is alphabetical.
- **Nit B:** the old two-package offline assumption no longer exists. The
  generated package is build-only; clean-cache verification installs only the
  SDK and rejects generated-package leakage.

## Qualification boundary

The candidate gates cover Node, Bun, Deno, Playwright browsers, and Firefox
ESR. Electron gets no separate support claim or CI matrix in this PR: it has no
additional SDK entry, and Electron documents its main process as Node.js and
its renderer as web-standard code.

The edge entry passed local Workerd and a [live Cloudflare authorization
request][cloudflare-live]. Next.js production build and local Vercel Edge
authorization passed, but live Vercel deployment did not: its reported 1.73 MB
compressed function exceeds Hobby's 1 MB limit. Vercel documents 1 MB Hobby,
2 MB Pro, and 4 MB Enterprise limits, so the result is recorded as local-only,
not completed live qualification.

Code: [maintainer qualification matrix][maintainer]. References:
[Electron process model][electron], [Cloudflare WASM bundling][cloudflare-wasm],
and [Vercel limits][vercel-limits].


[workflow]: https://github.com/JanssenProject/jans/blob/jans-cedarling-14582-js-sdk-qualification/.github/workflows/test-cedarling.yml
[manifest]: https://github.com/JanssenProject/jans/blob/jans-cedarling-14582-js-sdk-qualification/jans-cedarling/bindings/cedarling_js/package.json
[lockfile]: https://github.com/JanssenProject/jans/blob/jans-cedarling-14582-js-sdk-qualification/jans-cedarling/bindings/cedarling_js/package-lock.json
[consumer-readme]: https://github.com/JanssenProject/jans/blob/jans-cedarling-14582-js-sdk-qualification/jans-cedarling/bindings/cedarling_js/README.md
[maintainer]: https://github.com/JanssenProject/jans/blob/jans-cedarling-14582-js-sdk-qualification/jans-cedarling/bindings/cedarling_js/docs/readme.md
[build]: https://github.com/JanssenProject/jans/blob/jans-cedarling-14582-js-sdk-qualification/jans-cedarling/bindings/cedarling_js/scripts/build.mjs
[stage]: https://github.com/JanssenProject/jans/blob/jans-cedarling-14582-js-sdk-qualification/jans-cedarling/bindings/cedarling_js/scripts/stage-packages.mjs
[verify]: https://github.com/JanssenProject/jans/blob/jans-cedarling-14582-js-sdk-qualification/jans-cedarling/bindings/cedarling_js/scripts/verify-consumer.mjs
[embedded]: https://github.com/JanssenProject/jans/blob/jans-cedarling-14582-js-sdk-qualification/jans-cedarling/bindings/cedarling_js/src/engine/embedded.ts
[edge]: https://github.com/JanssenProject/jans/blob/jans-cedarling-14582-js-sdk-qualification/jans-cedarling/bindings/cedarling_js/src/engine/edge.ts
[signature-test]: https://github.com/JanssenProject/jans/blob/jans-cedarling-14582-js-sdk-qualification/jans-cedarling/bindings/cedarling_js/tests/contract/signature-validation.test.ts
[initialization-test]: https://github.com/JanssenProject/jans/blob/jans-cedarling-14582-js-sdk-qualification/jans-cedarling/bindings/cedarling_js/tests/unit/engine-initialization.test.ts
[browser-runner]: https://github.com/JanssenProject/jans/blob/jans-cedarling-14582-js-sdk-qualification/jans-cedarling/bindings/cedarling_js/tests/runners/browser-page.ts
[playwright-config]: https://github.com/JanssenProject/jans/blob/jans-cedarling-14582-js-sdk-qualification/jans-cedarling/bindings/cedarling_js/tests/playwright.config.ts
[unit-index]: https://github.com/JanssenProject/jans/blob/jans-cedarling-14582-js-sdk-qualification/jans-cedarling/bindings/cedarling_js/tests/unit/index.ts

[github-workspace]: https://docs.github.com/en/actions/reference/runners/github-hosted-runners#file-systems
[github-jobs]: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobs
[node-releases]: https://nodejs.org/en/about/previous-releases
[deno-releases]: https://docs.deno.com/runtime/fundamentals/stability_and_releases/
[bun-stable]: https://bun.sh/docs/guides/util/upgrade
[playwright-browsers]: https://playwright.dev/docs/browsers
[node-exports]: https://nodejs.org/api/packages.html#package-entry-points
[typescript-resolution]: https://www.typescriptlang.org/docs/handbook/modules/reference.html
[wasm-bindgen]: https://wasm-bindgen.github.io/wasm-bindgen/reference/deployment.html
[esbuild-binary]: https://esbuild.github.io/content-types/#binary
[cloudflare-modules]: https://developers.cloudflare.com/workers/vite-plugin/reference/non-javascript-modules/
[cloudflare-rust]: https://developers.cloudflare.com/workers/languages/rust/
[vercel-edge]: https://vercel.com/docs/functions/runtimes/edge#unsupported-apis
[vite-rolldown]: https://vite.dev/blog/announcing-vite8
[npm-ci]: https://docs.npmjs.com/cli/commands/npm-ci/
[playwright-ci]: https://playwright.dev/docs/ci#caching-browsers
[electron]: https://www.electronjs.org/docs/latest/tutorial/process-model
[cloudflare-wasm]: https://developers.cloudflare.com/workers/runtime-apis/webassembly/javascript/#bundling
[cloudflare-live]: https://cedarling-sdk-loading-lab.dah-kenangnon.workers.dev/
[vercel-limits]: https://vercel.com/docs/functions/runtimes/edge#limits

olehbozhok:
## Review — items to fix

Verified by building and packing the SDK locally (wasm-pack 0.14.0 + 0.15.0, Node 20, TypeScript 5.9.3).

### Reproduce the baseline

```bash
cd jans-cedarling/bindings/cedarling_wasm
wasm-pack build --release --locked --target web --scope janssenproject
cd ../cedarling_js
npm ci --ignore-scripts
npm run build
```

---

### 1. CommonJS type resolution is broken (blocking)

`package.json` serves one ESM-flavoured `types` for the `require` condition, so the README's own CJS example fails to type-check.

**Reproduce**

```bash
node scripts/stage-packages.mjs --output /tmp/stage
mkdir -p /tmp/cjs && cd /tmp/cjs
echo '{"name":"c","private":true,"type":"commonjs"}' > package.json
echo '{"compilerOptions":{"module":"node16","moduleResolution":"node16","strict":true,"noEmit":true}}' > tsconfig.json
echo 'import { createCedarling } from "@janssenproject/cedarling";' > index.ts
npm i --no-package-lock /tmp/stage/*.tgz typescript@5.9.3
./node_modules/.bin/tsc --noEmit
```

```
index.ts(1,33): error TS1479: The current file is a CommonJS module whose imports
will produce 'require' calls; however, the referenced file is an ECMAScript module
and cannot be imported with 'require'.
```

Runtime is fine (`require()` returns `createCedarling`); only types break. ESM types pass.

**Fix.** Copying `index.d.ts` to `index.d.cts` does *not* work — the declaration re-exports from ESM-flavoured `.d.ts` files and produces 8× TS1479. Emit a CJS-flavoured declaration tree in its own directory instead:

In `scripts/build.mjs`, after `tsc`:
- copy the `.d.ts` tree to `dist/cjs/`
- write `dist/cjs/package.json` containing `{"type":"commonjs"}`

In `package.json`:

```jsonc
"exports": {
  ".": {
    "import":  { "types": "./dist/index.d.ts",     "default": "./dist/index.js"  },
    "require": { "types": "./dist/cjs/index.d.ts", "default": "./dist/index.cjs" },
    "default": "./dist/index.js"
  },
  "./edge": {
    "types": "./dist/index.d.ts",
    "import": "./dist/edge/index.js",
    "default": "./dist/edge/index.js"
  }
}
```

Verified: CJS types pass, ESM types pass, runtime unchanged.

**Also fix the gate.** `node scripts/verify-consumer.mjs` exits 0 with the bug present — it only executes, never type-checks. Add a `tsc --noEmit` pass over a CJS and an ESM consumer (or `@arethetypeswrong/cli` against the tarball).

---

### 2. Publish job bypasses the packaging scripts added in this PR

`build_cedarling_js` in `.github/workflows/build-packages.yml` builds its tarball with `cp -r dist publish/dist` plus a hand-written `jq` manifest. It never calls `stage-packages.mjs` or `verify-consumer.mjs`, and never loads the tarball it publishes — unlike the sibling `build_cedarling_wasm`, which has a "Smoke-test WASM Node.js package" step.

Consequences:
- `pack()`'s five required-file assertions, the `node_modules`-leak check, and `assertExactDependencies` only run against a tarball CI discards.
- The `jq` selector omits `dependencies` unconditionally. Add a runtime dependency and `npm run package:verify` still passes while the published package ships without it.

**Fix.** Publish the artifact that was verified:

```bash
node scripts/stage-packages.mjs --version "$VERSION_TO_PUBLISH" --output publish
```

Minimum acceptable: add a smoke-test step matching the wasm job.

---

### 3. `dist` ships intermediate `tsc` output with unresolvable specifiers

**Reproduce**

```bash
grep -rn "cedarling:wasm-bytes\|@janssenproject/cedarling_wasm" dist --include=*.js | grep -v '^dist/index'
```

```
dist/engine/generated-loader.js:1  from "@janssenproject/cedarling_wasm"        # devDependency
dist/engine/embedded.js:1          import wasmBytes from "cedarling:wasm-bytes" # virtual specifier
dist/engine/edge.js:1              import wasmModule from "./cedarling_wasm_bg.wasm?module"  # no such sibling file
```

All three are in the tarball (139 entries). `build.mjs`'s `assertTextExcludes` guards only the three bundles.

The `exports` map blocks well-behaved resolvers, so nothing breaks today — but SBOM/SCA scanners see a phantom dependency, and every shipped `.d.ts.map` / `.js.map` points at `../src/**`, which is not in the tarball.

Size is not the issue: 16 MB of the 18 MB unpacked tarball is the intended surface (two 5.6 MB bundles + 4.1 MB edge WASM); leftovers are 1.2 MB. The `.d.ts` tree (140 KB) must stay — `index.d.ts` imports from it.

**Fix.** In `build.mjs`, after bundling, delete the unused `.js` files and all `.map` files from `dist` (keep the `.d.ts` tree). Drop `declarationMap` / `sourceMap` from the production build config.

---

### 4. `./edge` subpath is never resolved by any test

`verify-consumer.mjs:83-84` only `access()`es `dist/edge/index.js` by raw path; the generated consumer imports the root only. Drop or mistype the `./edge` key and every assertion still passes.

**Fix.** Add to the consumer script:

```js
import.meta.resolve("@janssenproject/cedarling/edge");
```

Resolution is verifiable under Node; execution is not (`?module` is a Workers convention — importing it gives `ERR_UNKNOWN_FILE_EXTENSION`, as expected).

---

### 5. `verify-consumer.mjs` version assertion cannot fail

`sourceManifest` is read from `package.json` (line 31); `stage-packages.mjs` is spawned without `--version` (lines 44-48), so it reads the same file. `installed.version !== sourceManifest.version` compares a value to itself. Same for `installed.private !== true`, which compares against a literal `stage-packages.mjs:109` hardcodes.

No SDK-vs-`cedarling_wasm` version comparison exists anywhere, so the PR's "exact SDK/generated-package version coordination" claim is unsupported. The published version is computed independently by CI from `npm dist-tag ls`, so `0.0.0-nightly` is the one version that never ships.

**Fix.** Pass an explicit `--version` distinct from the manifest (this also covers the documented `--version` path, currently untested), or drop the claim from the description.

---

### Minor

- `.github/workflows/build-packages.yml:664` — `steps.publish.outputs.version` is written and never read. Every nightly asset is named `cedarling_js_0.0.0_pkg.tar.gz` and `--clobber`s the previous one, so signed assets cannot be correlated to an npm version without unpacking.
- `build-packages.yml:641` — `cargo install wasm-pack --locked` is unpinned and unchecksummed in a job that cosign-signs and SLSA-attests its output. `test-cedarling.yml:289-295` pins 0.14.0 with `sha256sum --check --strict`; reuse it. (Not a correctness risk — see below.)
- `scripts/clean.mjs:9` — the `build` target removes a root-level `.build` that nothing creates; the browser bundle goes to `.test-dist/.build`.
- `scripts/stage-packages.mjs:104-120` — `optionalDependencies` and `peerDependencies` are never copied into the staged manifest, so `assertExactDependencies` can only inspect empty objects for those sections.
- `src/configuration/bootstrap.ts:184-187` — setting only `contextStore.defaultTtlSeconds` above 3600 now fails with `INPUT_CONFLICT` against the implicit default max. Deliberate and tested, but it is a behaviour break undocumented in the description.

### Checked, no issue

- wasm-pack 0.14.0 vs 0.15.0 produce **byte-identical** glue and `.wasm` (glue comes from wasm-bindgen 0.2.127 pinned in `Cargo.lock` under `--locked`). No glue-drift risk from an unpinned wasm-pack.
- `build.mjs`'s realm-check snippet occurs exactly once in the generated glue, so `String.replace` first-match is safe.
- `initSync` returns `instance.exports`, which carries `.memory` — `hasGeneratedModuleOutput` passes.
- Job permissions, nightly version bump under `set -euo pipefail`, provenance subject consistency, lockfile sync, `--ignore-scripts` with esbuild/Playwright, Node 22→24→26 step ordering, `setup-firefox` output name, job timeout.
- Cross-realm `copyUint8Array` / `httpUrl` rewrites: no value newly accepted that should not be; the Proxy case is a bug fix. Only shift: a detached-buffer `Uint8Array` now reports `INPUT_INVALID_TYPE` instead of `INPUT_OUT_OF_RANGE`.
- All 13 operation strings in `lifecycle.test.ts` match what the closed path reports; `shutDown()` genuinely memoises its promise.

moabu (dismissed):

duttarnab (commented):
1. Add a section to README.md or docs/readme.md describing the purpose and usage of all scripts defined in package.json.
2. Add an example to the README demonstrating how bootstrapProperties (Cedarling bootstrap properties) are used to initialize Cedarling.
3. Add a cedarling_js/examples folder containing sample applications for React.js, Node.js, etc., demonstrating how to use @janssenproject/cedarling. The examples should use a policy store with the .cjar extension.

Dahkenangnon:
Documentation in README.md and docs/readme.md are updated.

Example app cannot be in this PR, it will be part of a follow up PR.

duttarnab (changes requested):
1. The complete error message is not displayed. On testing unsigned authz getting error --- `Cedarling authorization failed` but the cause of error is not displayed.

e.g. `Error: request validation error: principal type `AGT::User` is not valid for `AGT::Action::"ReadResource"``

Dahkenangnon:
Thanks for raising this. The SDK intentionally keeps raw error details out of the stable public error
message because they can contain sensitive values. They are available for local debugging through an explicit opt-in
and `error.cause`.

  ```ts
  const initialized = await createCedarling({
    applicationName: "task-api",
    policyStore: {
      type: "inline",
      document: policyStoreDocument,
    },

    // Local debugging only: raw engine errors may contain sensitive data.
    debug: { dangerouslyExposeRawErrors: true },
  });

  if (!initialized.ok) {
    throw initialized.error;
  }

  const result = await initialized.value.authorizeUnsigned(request);

  if (!result.ok) {
    console.error(result.error.message); // Stable SDK message.
    console.error(
      result.error.cause instanceof Error
        ? result.error.cause.message
        : result.error.cause,
    ); // Full Cedarling/engine errors details.
  }
```
  cause is deliberately non-enumerable, so JSON.stringify(error) and object spreading do not disclose it accidentally.
  I will keep this option disabled in production.

I just push a new commit which  add this usage example to the README for discoverability.

**Review by duttarnab — COMMENTED** [review]
1. Following syntax is throwing error
```
  const result = await createCedarling(
        {
          bootstrapProperties: {
            CEDARLING_APPLICATION_NAME: "task-api",
            CEDARLING_LOG_TYPE: "memory",
            CEDARLING_LOG_LEVEL: "debug",
            CEDARLING_LOG_TTL: "100"
          },
          policyStore: { type: "archive", bytes: policyArchiveBytes },
        });
```

error: A Cedarling input field is not supported.

I wanted to use the bootstrapProperties along with the policy-store in byte-array format.

2. I understand in below syntax, cedarling_js checks every field name and value, then translates it into the engine's own settings. For e.g. `logging: { type: "memory", level: "info" }` ==> `CEDARLING_LOG_TYPE: "memory", CEDARLING_LOG_LEVEL: "debug"` . We need a section in README describing each allowed attributes in the typed mode createCedarling input.

```
createCedarling({
  applicationName: "task-api",
  logging: { type: "memory", level: "info" },
  policyStore: { ... },
})
```

**Comment by Dahkenangnon** [comment]
I agree that bootstrapProperties is integral to Cedarling and remains fully supported as the raw, cross-binding configuration path.

  However, raw bootstrap configuration must be complete: it required `CEDARLING_APPLICATION_NAME` and one Cedarling policy-store property, such as `CEDARLING_POLICY_STORE_URI`. `policyStore: { type: "archive", bytes }` is instead a JavaScript/WASM-specific in-memory archive mechanism. Mixing both would create ambiguous policy-source precedence; the WASM archive initializer discards bootstrap policy-store fields in favor of the supplied bytes.

  I therefore propose keeping the two modes mutually exclusive:
```ts
  const result = await createCedarling({
    bootstrapProperties: {
      CEDARLING_APPLICATION_NAME: "task-api",
      CEDARLING_POLICY_STORE_URI:
        "https://your-host.example/policy-store.cjar",
      CEDARLING_LOG_TYPE: "memory",
      CEDARLING_LOG_LEVEL: "DEBUG",
      CEDARLING_LOG_TTL: "100",
    },
  });
```

  For an in-memory Cedar Archive, the typed JavaScript configuration is the supported path:

```ts
  const result = await createCedarling({
    applicationName: "task-api",
    logging: {
      type: "memory",
      level: "debug",
      ttlSeconds: 100,
    },
    policyStore: {
      type: "archive",
      bytes: policyArchiveBytes,
    },
  });
``` 
The bootstrap-property contract remains the source of truth: Cedarling properties reference (https://docs.jans.io/stable/cedarling/reference/cedarling-properties/).

Dahkenangnon:
About your point 2, the readme already has it; see `Configuration reference` section

duttarnab:
> About your point 2, the readme already has it; see `Configuration reference` section

how would a user know that we can define `ttlSeconds` attribute for logging
```
logging: {
      type: "memory",
      level: "debug",
      ttlSeconds: 100,
    }
```

Dahkenangnon:
> > About your point 2, the readme already has it; see `Configuration reference` section
> 
> how would a user know that we can define `ttlSeconds` attribute for logging
> 
> ```
> logging: {
>       type: "memory",
>       level: "debug",
>       ttlSeconds: 100,
>     }
> ```

I will add now, expandable section for each top object properties .

olehbozhok:
> 1. Following syntax is throwing error
> 
> ```
>   const result = await createCedarling(
>         {
>           bootstrapProperties: {
>             CEDARLING_APPLICATION_NAME: "task-api",
>             CEDARLING_LOG_TYPE: "memory",
>             CEDARLING_LOG_LEVEL: "debug",
>             CEDARLING_LOG_TTL: "100"
>           },
>           policyStore: { type: "archive", bytes: policyArchiveBytes },
>         });
> ```
> 
> error: A Cedarling input field is not supported.
> 
> I wanted to use the bootstrapProperties along with the policy-store in byte-array format.
> 
> 2. I understand in below syntax, cedarling_js checks every field name and value, then translates it into the engine's own settings. For e.g. `logging: { type: "memory", level: "info" }` ==> `CEDARLING_LOG_TYPE: "memory", CEDARLING_LOG_LEVEL: "debug"` . We need a section in README describing each allowed attributes in the typed mode createCedarling input.
> 
> ```
> createCedarling({
>   applicationName: "task-api",
>   logging: { type: "memory", level: "info" },
>   policyStore: { ... },
> })
> ```

I agree with the idea that the interface for bootstrap properties should be unified across all bindings. I'm not sure if we need to parse fields to build bootstrap properties that will be passed to the Rust WASM, because it adds maintenance cost and is just harder to keep consistent with the codebase. 
And Rust code will check it anyway. Rust code should be the only point of truth.

Dahkenangnon:
> > About your point 2, the readme already has it; see `Configuration reference` section
> 
> how would a user know that we can define `ttlSeconds` attribute for logging
> 
> ```
> logging: {
>       type: "memory",
>       level: "debug",
>       ttlSeconds: 100,
>     }
> ```


Expandable sections are added to cover all deep options: https://github.com/JanssenProject/jans/pull/14833/changes/8ee2c1c1ce200f2036ac37d37fbba731914ca52c

PR #14893, opened 2026-08-28
fix(jans-fido2): report uncomputable metrics rates as null instead of 0.0

### LINKED ISSUE #14831: fix(jans-fido2): metrics publish unknown values as if they were known
**Describe the bug**

Two places emit numbers that were never observed, with nothing distinguishing them from real ones.

**1. The no-`ATTEMPT` fallback publishes certainty.** `:549-559` emits `completionRate: 1.0`,
`dropOffRate: 0.0` and `abandonmentRate: 0.0` when the window holds no `ATTEMPT` rows. The comment on
the last one says the rate "is not computable from this data", and then publishes `0.0`. On a dashboard
that is indistinguishable from a genuine 100% completion rate with no abandonment. The empty-dataset
branch below has the same problem.

**2. Rate scaling silently rewrites `successRate`.** `:532-536` scales `successRate` and `failureRate`
down whenever `SUCCESS + FAILURE > ATTEMPT`, so the total stays within `[0,1]`. The published value was
never observed and nothing marks it as adjusted.

**To Reproduce**

1. Query a window containing only completed ceremonies and no `ATTEMPT` rows — e.g. any range
   predating the change that began writing attempt entries.
2. `GET /metrics/analytics/errors` → `completionRate: 1.0`, `dropOffRate: 0.0`,
   `abandonmentRate: 0.0`, none of which were measured.

**Expected behavior**

Rates that cannot be computed omitted, or returned as null, rather than as `0.0` / `1.0`. Observed
rates reported as observed, with a note when the underlying data is inconsistent — the file already has
a `rejectionRateNote` pattern for exactly this.

**Additional context**

Parent issue #14828 . **Changes the response shape** — consumers currently receive a number in these
fields and would begin receiving null or no key. Needs agreement with @arnab-dutta and @Faisal Siddique
on how the Passkey Security Monitor should render an unknown rate, since "unknown" and "zero" must not
look alike there either.

imran-ishaq:
Prepare

- [x] Read PR guidelines (https://github.com/JanssenProject/jans/blob/main/docs/CONTRIBUTING.md#prs)
- [x] Read license information (https://github.com/JanssenProject/jans/blob/main/LICENSE)

---

Description

Target issue

closes #14831

Implementation Details

/fido2/metrics/analytics/errors published two kinds of number that were never observed, with nothing distinguishing them from measured ones.

1. The no-ATTEMPT fallback published certainty. A window holding no ATTEMPT rows emitted completionRate: 1.0, dropOffRate: 0.0 and abandonmentRate: 0.0 — the comment on the last one said the rate "is not computable from this data" and then published 0.0. On a dashboard that is indistinguishable from a genuine flawless window. The empty-dataset branch did the same with zeros.

2. Rate scaling silently rewrote successRate. Whenever SUCCESS + FAILURE > ATTEMPT, both successRate and failureRate were scaled down so the two summed inside [0,1]. The published figure was not the observed one and nothing marked it as adjusted.

The rate block is now one helper, putCeremonyRates, that reports what was measured and says so where it could not measure:

- No ATTEMPT rows in range — completionRate, dropOffRate and abandonmentRate are null; they have no denominator. successRate/failureRate are still reported against the completions, and the note states that this is a different denominator from the usual one.
- Empty range — all five rates are null.
- Completions outnumber recorded starts — successRate and failureRate stand as observed and are never rescaled. completionRate is capped at 1.0; dropOffRate, which is inferred from the residual, is null, because there is no residual left to infer anyone dropping off from.
- Abandonments outnumber recorded starts — abandonmentRate is still capped at 1.0, but the cap is now stated rather than applied silently.

A new rateNote field carries the explanation, following the rejectionRateNote pattern the file already uses for the attestation breakdown. It is absent when every rate computed normally, so a client can treat its presence as the signal that something needs explaining.

Impact on existing behaviour

- Response shape: every rate key is still present. The controller serializes through a default ObjectMapper (inclusion ALWAYS), so an uncomputable rate arrives as "completionRate": null, not a missing key. The type widens from number to number-or-null; the swagger schema is updated with nullable: true.
- A healthy window is unchanged. completionRate is now totalCompleted / totalStarted rather than successRate + failureRate — the same value. Only the anomalous branches behave differently.
- successRate can now exceed 1.0 in a window whose completions belong to starts recorded outside it. That is the observed figure; the rateNote says so. This is the change most likely to surprise a UI that renders the value as a percentage bar.
- Consumers: the config-api plugin passes the payload through as JsonNode, so nothing there binds to a primitive. The in-repo Java reader (Fido2AnalyticsService) already casts to Double and null-guards before comparing, so it degrades to "no insight" rather than throwing. There are no Python, TypeScript or Go consumers of these fields in this repo.
- Aggregations are untouched — calculateAggregation and rateDenominator are unchanged, and persisted aggregation rows keep their shape.

@arnab-dutta @faisalsiddique — the Passkey Security Monitor has to render an unknown rate as unknown. "Unknown" and "zero" must not look alike there either, which is the whole point of the change; the rateNote gives the UI the text to show alongside it.

---
Test and Document the changes

## Summary by CodeRabbit

* **New Features**
  * FIDO2 ceremony analytics now provide more accurate success, failure, completion, drop-off, and abandonment rates.
  * Rates without a valid denominator are reported as unknown instead of zero.
  * Analytics responses include an explanatory note when rates are unknown, capped, or use an alternate denominator.

* **Documentation**
  * Updated Passkey Telemetry guidance and API documentation to explain rate calculations and how unknown values should be displayed.

### DIFF
--- a/docs/janssen-server/fido/passkey-telemetry.md
+++ b/docs/janssen-server/fido/passkey-telemetry.md
@@ -119,6 +119,19 @@ observed to have lapsed. In multi-node deployments the sweep is not coordinated 
 `abandonmentRate` is approximate — an exact count is available by querying `jansStatus = 'abandoned'`
 directly within the retention window.
 
+!!! note "An unknown rate is `null`, not zero"
+    A rate here is a ratio against the `ATTEMPT` count. A range can hold terminal entries whose
+    `ATTEMPT` falls outside it, or predate attempt tracking entirely, and then a rate has no
+    denominator to be computed from. Those rates are reported as `null` and never as `0.0` or `1.0`,
+    which on a dashboard would be indistinguishable from a range that really was measured at a
+    flawless completion rate with no abandonment.
+
+    Whenever any rate in the response is `null`, capped at `1.0`, or measured against a different
+    denominator, a `rateNote` field is present saying which and why; it is absent when everything was
+    computed as normal. `successRate` and `failureRate` are reported as observed and are never
+    rescaled, so in a range whose completions outnumber its recorded starts they can sum above 1.0 —
+    the `rateNote` says so. Render an unknown rate as unknown; it must not look like a zero.
+
 !!! warning "A failed fingerprint is never a `FAILURE`"
     With platform authenticators such as Touch ID, Face ID or Windows Hello, user verification
     happens **inside the authenticator**. A wrong fingerprint causes the operating system to retry
--- a/jans-fido2/docs/jansFido2Swagger.yaml
+++ b/jans-fido2/docs/jansFido2Swagger.yaml
@@ -1419,22 +1419,30 @@ components:
           description: Count of errors by message.
         successRate:
           type: number
-          description: Ratio of started operations that succeeded (0.0–1.0). Based on ATTEMPT count as denominator.
+          nullable: true
+          description: Ratio of started operations that succeeded. Based on ATTEMPT count as denominator, and reported as observed, so it can exceed 1.0 when completions in the range belong to starts outside it. Null when nothing was recorded. Where no ATTEMPT entries exist at all it is the share of completed ceremonies that succeeded instead — see rateNote.
         failureRate:
           type: number
-          description: Ratio of started operations that failed (0.0–1.0). Based on ATTEMPT count as denominator.
+          nullable: true
+          description: Ratio of started operations that failed. Same denominator and same caveats as successRate.
         completionRate:
           type: number
-          description: Ratio of started operations that completed (success or failure). Equals successRate + failureRate (0.0–1.0).
+          nullable: true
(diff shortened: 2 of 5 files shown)


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/8b63c51e-2fd0-426f-bb90-f679002ac9af?refs=20d5b6dd72dfc692074a6b24746f93ac21b1a0b2&source=prChecksComment) |

yurem (approved):

yurem (approved):

PR #14900, opened 2026-08-30
feat(jans-cedarling)!: package portable Cedarling WASM binding

### LINKED ISSUE #14899: feat(jans-cedarling): package Cedarling WASM for JavaScript runtimes
## 1. Description of the Problem

1. The ignored `cedarling_wasm/pkg/` output from `wasm-pack --target web` is a generated input, not a consumer-ready package for browser ESM, Node ESM, CommonJS, and static edge WASM loading.
2. Generated public method names are Rust snake_case. JavaScript consumers need camelCase method names without changing Cedarling-defined bootstrap-property or returned-data keys.
3. Browser, Node, CommonJS, and edge environments load the same WASM asset differently. Consumers must not locate, fetch, read, or compile that asset.
4. Generated objects own WASM resources. Consumers should receive ordinary JavaScript values without calling generated initialization functions or generated `free()`.
5. The package must ship one canonical raw WASM asset rather than separate runtime-specific generated WASM packages.

### 2.1. Package boundary

1. Build one consumer-ready JavaScript package in `jans-cedarling/bindings/cedarling_wasm/js/`.
2. Use one `wasm-pack build --target web` binding as the input for browser ESM, Node ESM, CommonJS, and explicit edge ESM package entries.
3. Add `wasm-bindgen` `js_name` attributes to public Rust exports that need camelCase generated JavaScript names.
4. Preserve generated Cedarling operations and their defined data semantics. The package layer owns only runtime loading, initialization, and generated-resource cleanup.
5. Initialize WASM internally when an exported Cedarling operation needs it,transfer resource-owning generated results into ordinary JavaScript values,and release generated resources internally.

### 2.2. Files and ownership

```text
jans-cedarling/bindings/cedarling_wasm/
  pkg/{cedarling_wasm.js, cedarling_wasm_bg.wasm, cedarling_wasm.d.ts}
                                                generated, ignored build input
  js/{                                     handwritten, tracked package project
    package.json, package-lock.json, README.md,
    src/{browser.mjs, node.mjs, edge.mjs, runtime.mjs},
    scripts/{build.mjs, verify-package.mjs},
    tests/{browser.mjs, edge.mjs},
    dist/{wasm/, esm/, cjs/, edge/, types/} generated, ignored publishable output
  }
```

1. `pkg/` exists because `wasm-pack` produces the only generated binding input used by the package build. It remains ignored and is never published.
2. `js/package.json` and `package-lock.json` define reproducible package metadata, exports, dependency resolution, and publication contents.
3. `js/README.md` is the npm consumer document and is published with the package.
4. `src/browser.mjs`, `src/node.mjs`, and `src/edge.mjs` are necessary because browser, Node, and edge environments resolve the same WASM asset differently.
5. `src/runtime.mjs` contains the only shared handwritten runtime behavior: lazy initialization and generated-resource release.
6. `scripts/build.mjs` assembles `dist/` from `pkg/`; `scripts/verify-package.mjs` verifies the packed artifact rather than source files.
7. `tests/browser.mjs` and `tests/edge.mjs` qualify the loading models that Node cannot execute directly.
8. `dist/` is generated and ignored but is the only publishable runtime output. 

### 3.1. Package entries and exports

1. Generate root `pkg/` once with `wasm-pack build --target web`.
2. Build browser ESM, Node ESM, Node CommonJS, and explicit edge ESM entries from that same generated glue and WASM input.
3. Use a bundler-compatible asset path for browser ESM, a package-relative loader for Node ESM and CommonJS, and a static WASM module import for the edge entry.
4. Export browser ESM, Node `import`, Node `require`, and `./edge` explicitly from the package manifest.
5. Copy generated declarations into ESM and CommonJS declaration trees during the build so TypeScript resolves both module systems.

### 3.2. Initialization and lifecycle

1. Memoize generated module initialization per runtime entry and clear a failed initialization attempt so a later exported operation can retry.
2. Transfer generated resource-owning results to ordinary JavaScript values without changing their Cedarling-defined fields or meaning.

### 3.3. Single WASM asset

1. Copy `pkg/cedarling_wasm_bg.wasm` once to `js/dist/wasm/cedarling_wasm_bg.wasm` and verify the copied bytes.
2. Require every final runtime entry to resolve that canonical asset through its runtime-specific loading mechanism.
3. Assert that the packed artifact contains exactly one raw `.wasm` file and no generated `pkg/` directory.
4. Keep final runtime entries independent of a second generated package dependency.

### 3.4. Qualification and CI

1. Pack the package and test isolated installed ESM and CommonJS consumers.
2. Type-check both installed module systems and execute representative exported Cedarling operations.
3. Qualify browser and edge asset loading against the packed runtime entries.
4. Add Bun and Deno qualification when those runtimes are claimed as supported.
5. Make CI build and qualify the same packed artifact that publication uses.

## Updates: backward-compatible generated initialization and lifecycle
0. Keep the snake_case naming until the team identify and accept to migrate all impacted internal consumer app like Tarp extension, meanwhile, don't add the proposed `js_name` binding option from rust side
1. Preserve the complete generated initialization surface for existing consumers:
   - default `initWasm(input?)`;
   - named `initSync(input)`;
   - named `init(...)` and `init_from_archive_bytes(...)`.

Tarp already imports the default initializer and named `init` from `@janssenproject/cedarling_wasm` ([source](https://github.com/JanssenProject/jans/blob/f2491f4f47ae2ee7e7c37191afad2a9e748278d6/demos/janssen-tarp/browser-extension/src/features/cedarling/components/UnsignedAuthzForm.tsx#L4)) and calls `await initWasm()` before `init(...)` ([source](https://github.com/JanssenProject/jans/blob/f2491f4f47ae2ee7e7c37191afad2a9e748278d6/demos/janssen-tarp/browser-extension/src/features/cedarling/components/UnsignedAuthzForm.tsx#L42-L58)). The multi-issuer flow uses the same pattern ([source](https://github.com/JanssenProject/jans/blob/f2491f4f47ae2ee7e7c37191afad2a9e748278d6/demos/janssen-tarp/browser-extension/src/features/cedarling/components/MultiIssuerAuthzForm.tsx#L67-L86)).
2. New consumers must be able to call `init(...)` or `init_from_archive_bytes(...)` directly. Those APIs load and initialize the canonical WASM asset internally. Legacy explicit initialization and automatic initialization must share one per-entry initialized state, so either path is safe and does not initialize the generated module twice.

Dahkenangnon:
### Prepare

  - [x] Read PR guidelines

  ———

#### Target issue

closes #14899

#### Implementation Details
Adds the publishable @janssenproject/cedarling_wasm JavaScript package, assembled from one generated web-target WASM input and shipped with exactly one WASM asset.

The package provides automatic initialization for browser, Node.js ESM, Node.js CommonJS, and edge consumers. It also provides ESM-only ./manual and ./wasm entries for bundlers that must emit the WASM asset themselves.
The generated wasm-bindgen API and types remain public. Public JavaScript service methods use camelCase while generated result data retains its generated field names. Explicit initialization and generated free() APIs remain available for compatibility.

Adds runtime, browser, edge, manual-bundler, ESM, CommonJS, and TypeScript consumer qualification; dedicated cedarling_wasm_test CI; release packaging support; and consumer/maintainer documentation covering configuration, authorization, context data, logs, issuers, annotations, and lifecycle.

#### BREAKING CHANGE

Public callable JavaScript APIs now use camelCase. Existing consumers must update the following calls when upgrading; these renames do not change their arguments or returned data shapes.

<details>
<summary>Complete old-name → new-name mapping</summary>

| Previous name | New name |
|---|---|
| `init_from_archive_bytes` | `initFromArchiveBytes` |
| `Cedarling.new_from_map` | `Cedarling.newFromMap` |
| `authorize_unsigned` | `authorizeUnsigned` |
| `authorize_multi_issuer` | `authorizeMultiIssuer` |
| `authorize_unsigned_batch` | `authorizeUnsignedBatch` |
| `authorize_multi_issuer_batch` | `authorizeMultiIssuerBatch` |
| `annotations_map` | `annotationsMap` |
| `annotation_values` | `annotationValues` |
| `annotations_by_policy` | `annotationsByPolicy` |
| `pop_logs` | `popLogs` |
| `get_log_by_id` | `getLogById` |
| `get_log_ids` | `getLogIds` |
| `get_logs_by_tag` | `getLogsByTag` |
| `get_logs_by_request_id` | `getLogsByRequestId` |
| `get_logs_by_request_id_and_tag` | `getLogsByRequestIdAndTag` |
| `shut_down` | `shutDown` |
| `push_data_ctx` | `pushDataCtx` |
| `get_data_ctx` | `getDataCtx` |
| `get_data_entry_ctx` | `getDataEntryCtx` |
| `remove_data_ctx` | `removeDataCtx` |
| `clear_data_ctx` | `clearDataCtx` |
| `list_data_ctx` | `listDataCtx` |
| `get_stats_ctx` | `getStatsCtx` |
| `is_trusted_issuer_loaded_by_name` | `isTrustedIssuerLoadedByName` |
| `is_trusted_issuer_loaded_by_iss` | `isTrustedIssuerLoadedByIss` |
| `total_issuers` | `totalIssuers` |
| `loaded_trusted_issuers_count` | `loadedTrustedIssuersCount` |
| `loaded_trusted_issuer_ids` | `loadedTrustedIssuerIds` |
| `failed_trusted_issuer_ids` | `failedTrustedIssuerIds` |
| `json_string` | `jsonString` |

`initFromArchiveBytes` is a module export; `newFromMap` is a static `Cedarling` method. The remaining service methods belong to `Cedarling` instances.

The `jsonString()` rename applies to **AuthorizeResult, MultiIssuerAuthorizeResult, DataEntry, and DataStoreStats**.

</details>

Canonical bootstrap-property keys, request JSON, serialized JSON, and generated data/getter properties—including `request_id`, `data_type`, `is_ok`, `item_index`, and `batch_id`—remain unchanged. Explicit initialization and resource-release APIs (`initWasm`, `initSync`, and `free()`) remain available.

  ———
- [x] Relevant documentation has been updated if any (i.e. user guides, installation and configuration guides,technical design docs etc)

  Please check the below before submitting your PR. The PR will not be merged if there are no commits that start with
  docs: to indicate documentation changes or if the below checklist is not selected.

  - [ ] I confirm that there is no impact on the docs due to the code changes in this PR.

## Summary by CodeRabbit

- **New Features**
  - Added the `@janssenproject/cedarling_wasm` package with browser, Node.js, edge, CommonJS, and manual integration options.
  - Added initialization helpers, archive-based loading, batch authorization, multi-issuer authorization, context management, logging, and trusted-issuer APIs.
  - JavaScript-callable APIs now use camelCase naming.

- **Documentation**
  - Added package usage, integration, API, migration, and maintainer guidance.

- **Bug Fixes**
  - Updated examples and benchmarks to use the current API names.

- **Tests**
  - Added browser coverage across Chromium, Firefox, and WebKit, plus Node.js version matrix testing.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/671cd6e2-188c-437f-8e53-08aac8c98a0d?refs=f2283787f773c43976472a34e81e12af4d693053&source=prChecksComment) |

jans-cedarling/bindings/cedarling_wasm/js/scripts/build.mjs
```diff
@@ -0,0 +1,361 @@
...
+      () => ({ path: generatedGluePath, namespace: "cedarling-generated-glue" }),
+    );
+    esbuild.onLoad({ filter: /.*/, namespace: "cedarling-generated-glue" }, async ({ path }) => {
+      const source = await readFile(path, "utf8");
+      if (!source.includes(realmSensitiveModuleCheck)) {
+        throw new Error("Generated glue changed its WebAssembly.Module check");
+      }
+      let contents = source.replace(
+          realmSensitiveModuleCheck,
+          realmNeutralModuleCheck,
```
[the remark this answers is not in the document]
Dahkenangnon: Already addressed in the current implementation: `realmNeutralModuleCheck` retains the generated fallback and assigns `module = new WebAssembly.Module(module)` when `WebAssembly.Module.exports(module)` rejects a BufferSource. `npm run check` passes.

.github/workflows/build-packages.yml
```diff
@@ -547,10 +540,13 @@ jobs:
...
-        tar -czvf cedarling_wasm_"${TAG}"_pkg.tar.gz -C pkg .
-        cosign sign-blob --yes --bundle cedarling_wasm_"${TAG}"_pkg.tar.gz.bundle cedarling_wasm_"${TAG}"_pkg.tar.gz
-        gh release upload "${VERSION}" *.tar.gz *.bundle --clobber || {
+        TARBALL="cedarling_wasm_${TAG}_pkg.tar.gz"
+        if [ ! -f "${TARBALL}" ]; then
+          echo "::error::Published package artifact ${TARBALL} not found"
+          exit 1
+        fi
+        cosign sign-blob --yes --bundle "${TARBALL}.bundle" "${TARBALL}"
+        gh release upload "${VERSION}" "${TARBALL}" "${TARBALL}.bundle" --clobber || {
```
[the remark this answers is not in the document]
Dahkenangnon: I verified this handler is inherited unchanged from current main: https://github.com/JanssenProject/jans/blob/2a7abba68e6/.github/workflows/build-packages.yml#L549-L551. This PR changes Cedarling WASM package assembly and publication only, so I am preserving the established release-upload policy.

jans-cedarling/bindings/cedarling_wasm/src/lib.rs
```diff
@@ -68,8 +68,23 @@ impl From<cedarling::MultiIssuerAuthorizeResult> for MultiIssuerAuthorizeResult
     }
 }
 
-/// Create a new instance of the Cedarling application.
-/// This function can take as config parameter the eather `Map` other `Object`
+/// Creates a Cedarling application from bootstrap properties.
+///
+/// # Arguments
+///
+/// * `config` - A JavaScript `Map` or plain object of Cedarling bootstrap-property names and values.
```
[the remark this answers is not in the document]
Dahkenangnon: Fixed in the pending update: both Rustdoc lines now comply with the repository 100-character width. `cargo fmt --check` and the full `npm run check` package gate pass.

jans-cedarling/bindings/cedarling_wasm/src/lib.rs
```diff
@@ -319,16 +380,20 @@ impl Cedarling {
...
     /// ```
+    #[wasm_bindgen(js_name = annotationValues)]
     pub fn annotation_values(&self, policy_ids: Vec<String>, key: &str) -> Vec<String> {
         let ids: Vec<PolicyId> = policy_ids.iter().map(PolicyId::new).collect();
         self.instance.annotation_values(ids.iter(), key)
     }
 
     /// Return the annotations of each given policy, grouped by policy ID
-    /// the loss-free companion to `annotations_map`. Unknown policy IDs are
+    /// the loss-free companion to `annotationsMap`. Unknown policy IDs are
```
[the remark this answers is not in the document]
Dahkenangnon: Fixed in the pending update: the documentation now uses a sentence boundary before “It is the loss-free companion”. The full `npm run check` package gate passes.

jans-cedarling/bindings/cedarling_wasm/src/lib.rs
```diff
@@ -382,8 +479,18 @@ impl Cedarling {
         result
     }
 
-    /// Get logs by tag, like `log_kind` or `log level`.
-    /// Tag can be `log_kind`, `log_level`.
+    /// Returns retained memory logs matching an indexed value.
+    ///
+    /// # Arguments
+    ///
+    /// * `tag` - A log kind (`"System"`, `"Decision"`, or `"Metric"`) or a system-log level such as `"DEBUG"`.
```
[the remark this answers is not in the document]
Dahkenangnon: Fixed in the pending update: both log-tag descriptions are wrapped below the configured 100-character width. I also audited every branch-added Rustdoc line; no remaining added line exceeds that limit. `cargo fmt --check` and `npm run check` pass.

jans-cedarling/bindings/cedarling_wasm/js/README.md
```diff
@@ -0,0 +1,440 @@
...
+
+`init` accepts an object or `Map` containing the canonical uppercase Cedarling
+bootstrap properties. The policy store property is a **serialized JSON string**;
+the request passed to an authorization method is also a JSON string.
+
+```ts
+import { init } from "@janssenproject/cedarling_wasm";
+
+const policyStore = {
+  // A valid Cedarling policy-store document.
```
duttarnab, line 46: Please see [#service_cedarling > general chat @ 💬](https://chat.gluu.org/#narrow/channel/541647-service_cedarling/topic//near/620353562)

I think this means Policy-store in JSON format is not acceptable. YAML can be used for the test suites.

jans-cedarling/bindings/cedarling_wasm/js/README.md
```diff
@@ -0,0 +1,440 @@
...
+  CEDARLING_LOG_TYPE: "memory",
+  CEDARLING_LOG_TTL: 120,
+  CEDARLING_JWT_SIG_VALIDATION: "disabled",
+  CEDARLING_JWT_STATUS_VALIDATION: "disabled",
+});
+
+try {
+  const result = await cedarling.authorizeUnsigned(JSON.stringify({
+    principal: {
+      cedar_entity_mapping: { entity_type: "Task::User", id: "alice" },
```
duttarnab, line 62: May be you should have converted to accept the attributes of `principal`, `resource` in camelCase. For e.g 

```
const result = await cedarling.authorizeUnsigned(JSON.stringify({
    principal: {
      cedarEntityMapping: { entityType: "Task::User", id: "alice" },
      role: "member",
    },
    action: 'Task::Action::"Read"',
    resource: {
      cedarEntityMapping: { entityType: "Task::Document", id: "document-1" },
      owner: "alice",
    },
    context: { tenant: "example" },
  }));
```

jans-cedarling/bindings/cedarling_wasm/js/README.md
```diff
@@ -0,0 +1,440 @@
...
+
+Use it to evaluate Cedar policies locally, retain decision logs, manage
+context data, and validate token-based authorization requests.
+
+## Install
+
+```sh
+npm install @janssenproject/cedarling_wasm
+```
+
```
duttarnab, line 16: After Install topic, please include the topic of `Build from source`. Under this topic, you can provide the link to the page covering this topic. If user is interested then he can go to the page to read this topic.

jans-cedarling/bindings/cedarling_wasm/js/docs/readme.md
```diff
@@ -0,0 +1,52 @@
+# JavaScript package maintenance
```
duttarnab, line 1: Rename the file so that we could know its purpose. `readme.md` does not tell us anything.

jans-cedarling/bindings/cedarling_wasm/js/docs/readme.md
```diff
@@ -0,0 +1,52 @@
...
+
+`js/` is the hand-written package. `../pkg/` is ignored output from:
```
duttarnab, line 3: There is no introduction to the content that will be covered under this page. 
`'js/' is the hand-written package. '../pkg/' is ignored output from:` is confusing and not the proper way to start.

`wasm-pack build --release --locked --target web --scope janssenproject` --- Please describe the purpose of this command.

jans-cedarling/bindings/cedarling_wasm/js/docs/readme.md
```diff
@@ -0,0 +1,52 @@
...
+
+```sh
```
duttarnab, line 5: Please make this tutoial so that it can be easily digested by human. Just putting the unordered information on page does not helps much. See how these docs have been written:

https://github.com/JanssenProject/jans/tree/main/docs/agama
https://github.com/JanssenProject/jans/tree/main/docs/casa

duttarnab (changes requested):
Please see inline comments on Readme.md

jans-cedarling/bindings/cedarling_wasm/js/README.md
```diff
@@ -0,0 +1,440 @@
+# Cedarling WebAssembly for JavaScript
+
+`@janssenproject/cedarling_wasm` runs [Cedarling](https://docs.jans.io/stable/cedarling/)
```
duttarnab, line 3: The documentation should be the standard way to communicate with humans. 

For. e.g. in below cases the sentences have not been started properly.
- @janssenproject/cedarling_wasm runs [Cedarling](https://docs.jans.io/stable/cedarling/) authorization in JavaScript and TypeScript.
- init accepts an object or Map containing the canonical uppercase Cedarling bootstrap properties. 
- decision: false is a valid policy decision, not an exception. Invalid configuration, malformed requests, failed policy loading, or failed token validation reject their promise and should be handled with try/catch.
- init and initFromArchiveBytes are the normal application APIs: each first ensures the packaged WebAssembly module is initialized. initWasm and initSync remain useful for applications that already manage their own WebAssembly loading.

jans-cedarling/bindings/cedarling_wasm/js/README.md
```diff
@@ -0,0 +1,440 @@
...
+authorization in JavaScript and TypeScript. It exposes the Cedarling WebAssembly
+API with automatic WebAssembly initialization for browsers and Node.js, one
+packaged WebAssembly asset, CommonJS support, and an edge-specific entry.
+
+Use it to evaluate Cedar policies locally, retain decision logs, manage
```
duttarnab, line 8: > Use it to evaluate Cedar policies locally, retain decision logs, manage context data, and validate token-based authorization requests.


What it mean by `Use it to evaluate Cedar policies locally` ? Can't it be used on server or any other platform. The sentence choice should be thoughtful and complete.

jans-cedarling/bindings/cedarling_wasm/js/README.md
```diff
@@ -0,0 +1,440 @@
...
+context data, and validate token-based authorization requests.
+
+## Install
```
duttarnab, line 11: Please see the documentation done on jans-cedarling project and bindings. Refer below examples: 

https://github.com/JanssenProject/jans/tree/main/docs/casa
https://github.com/JanssenProject/jans/tree/main/docs/agama

I think we can improve the documentation so that it is more consumable and ordered.

jans-cedarling/bindings/cedarling_wasm/js/scripts/build.mjs
```diff
@@ -0,0 +1,370 @@
...
+const edgeOutput = join(distribution, "edge");
+const manualOutput = join(distribution, "manual");
+const wasmOutput = join(distribution, "wasm");
+const typesOutput = join(distribution, "types");
+const declarations = join(typesOutput, "esm");
+const commonJsDeclarations = join(typesOutput, "cjs");
+const generatedDeclarationPath = resolve(root, "../pkg/cedarling_wasm.d.ts");
+const wasmPath = resolve(
+  root,
+  "../pkg/cedarling_wasm_bg.wasm",
```
duttarnab, line 30: I understand we are using the wasm binding generated at cedarling_wasm/pkg. Can we put this path in some configuration file so that in case of change in path we need not search and make changes in all the files.

jans-cedarling/bindings/cedarling_wasm/js/scripts/build.mjs
```diff
@@ -0,0 +1,370 @@
...
+}
+
+const publicDeclarations = ["generated.d.ts", "index.d.ts", "edge.d.ts", "manual.d.ts", "wasm.d.ts"];
+
+async function rewriteDeclaration(name) {
+  const path = join(declarations, name);
+  const source = await readFile(path, "utf8");
+  await writeFile(
+    path,
+    source.replaceAll("../../pkg/cedarling_wasm.js", "./generated.js"),
```
duttarnab, line 184: I understand we are using the wasm binding generated at cedarling_wasm/pkg. Can we put this path in some configuration file so that in case of change in path we need not search and make changes in all the files.

jans-cedarling/bindings/cedarling_wasm/js/src/cjs.ts
```diff
@@ -0,0 +1,40 @@
+import { initSync as initializeGeneratedModule } from "cedarling:generated-glue";
+import type {
+  InitInput,
+  InitOutput,
+} from "../../pkg/cedarling_wasm.js";
```
duttarnab, line 5: I understand we are using the wasm binding generated at cedarling_wasm/pkg. Can we put this path in some configuration file so that in case of change in path we need not search and make changes in all the files.

jans-cedarling/bindings/cedarling_wasm/js/scripts/verify-consumer.mjs
```diff
@@ -0,0 +1,516 @@
...
+  ], {
+    cwd: root,
+    env: {
+      ...process.env,
+      CEDARLING_BROWSER_ARCHIVE: archive,
+      CEDARLING_BROWSER_BUNDLE: javascript,
+      CEDARLING_BROWSER_WASM: wasm[0],
+      CEDARLING_BROWSER_EXPECT_WASM_REQUEST: "true",
+    },
+  }).catch(output);
```
duttarnab, line 255: Promise.prototype.catch ignores a non-callable handler, so the rejection still propagates — but the child process's stdout/stderr is never printed.

duttarnab (changes requested):
1. I would suggest adding example/integration projects such as:

```
examples/
├── react-vite/
├── webpack/
├── node-esm/
├── node-commonjs/
└── edge/
```

jans-cedarling/bindings/cedarling_wasm/js/tests/browser.test.mjs
```diff
@@ -0,0 +1,118 @@
...
+import { createServer } from "node:http";
+import test from "node:test";
+
+import { chromium, firefox, webkit } from "@playwright/test";
+
+const bundlePath = process.env.CEDARLING_BROWSER_BUNDLE;
+const archivePath = process.env.CEDARLING_BROWSER_ARCHIVE;
+const wasmPath = process.env.CEDARLING_BROWSER_WASM;
+const expectsWasmRequest = process.env.CEDARLING_BROWSER_EXPECT_WASM_REQUEST === "true";
+if (bundlePath === undefined || archivePath === undefined) {
```
duttarnab, line 12: Need to include test cases for  unsigned and multi-issuer authorization.

duttarnab (changes requested):
1. Remove `bindings/cedarling_js/`.
2. 
**Good to have:**
Rename to camelCase have been half done. 

Methods not renamed to camelCase:
- json_string() on AuthorizeResult, MultiIssuerAuthorizeResult, DataEntry, DataStoreStats
- is_ok, item_index, batch_id, request_id, data_type, created_at, expires_at, access_count, entry_count, capacity_usage_percent etc.

faisalsiddique4400:
Most of this is small, two items I think are worth blocking on.

+1 to @duttarnab on removing `bindings/cedarling_js`, on the half-done camelCase rename, and on `verify-consumer.mjs:255`. Three things I can add to those.

**`cedarling_js` may be worth treating as blocking rather than cleanup.** It depends on `file:../cedarling_wasm/pkg` and imports 19 symbols this PR renames, and commit `2076fd5c4ca` removes its only CI job, so as far as I can tell it stops compiling on this branch with nothing left to report it.

**The rename list might be worth splitting**, since the two halves look like different decisions:

- `json_string()` (lib.rs 54, 914, 1289, 1338), `item_index` (951), `is_ok` (981, 1019) and `batch_id` (1064, 1103) are **methods**. Every sibling method was renamed, so these read as an oversight and seem worth fixing here.
- `request_id`, `data_type`, `created_at`, `expires_at`, `access_count`, `entry_count` and `capacity_usage_percent` are **`pub` fields**. The PR deliberately preserves generated data names and `js/README.md` documents that, so those feel like a separate decision rather than a miss.

**On `verify-consumer.mjs:255`, the cause looks like a shadowed binding.** `function output` is at line 64, but line 190 declares `const output = join(consumer, "manual-browser-output")`, a string, which shadows it for all of `verifyManualBrowserConsumer`. Line 80 shadows the same way in `verifyEdgeConsumer`, though that one is latent since the function has no `.catch(output)`. Renaming both locals to something like `outputDirectory` would cover both.

## Two I did not see covered yet

### 1. The docs site still shows the old snake_case API

These pages may need a pass alongside the rename:

| Page | Stale lines |
| --- | --- |
| `docs/cedarling/tutorials/javascript.md` | 18 |
| `docs/cedarling/tutorials/cedarling-getting-started.md` | 11 |
| `docs/cedarling/reference/cedarling-policy-store.md` | 8 |

Two more name the WASM helper directly:

- `bindings/cedarling_uniffi/README.md:314`
- `docs/cedarling/developer/mobile-apps/cedarling-uniffi.md:53`

### 2. The rename would ship as a patch bump

- `build-packages.yml:487` computes **`0.0.338` -> `0.0.339`**
- Every public method renames, and **`main` disappears** from the manifest
- So there is **no semver signal** for either change
- Flex admin-ui pins `0.0.338` and calls `init_from_archive_bytes` and `authorize_multi_issuer` directly, so it would break on the next bump

Might be worth a **minor or major** version, a `BREAKING CHANGE:` footer, and a rename table in the release notes.

jans-cedarling/bindings/cedarling_wasm/js/README.md
Dahkenangnon: Yes, according to https://github.com/JanssenProject/jans/pull/14980#pullrequestreview-5142552560 , Policy-store in JSON format may be remove soon. So, your comment is accepted and I will use either YAML or cjar format (I prefer this) in the snippets.

jans-cedarling/bindings/cedarling_wasm/js/README.md
Dahkenangnon: Examples/integrations sample must be introduced in a followup PR.

Ok, for the doc improvement

jans-cedarling/bindings/cedarling_wasm/js/scripts/build.mjs
Dahkenangnon: I don't think that, having a centralized config is not necessary because doing so, might prevent typescript to work properly as it need to be able to resolve dynamically the import path.

So, I recomment keep the import as it.

jans-cedarling/bindings/cedarling_wasm/js/README.md
duttarnab: ok noted

jans-cedarling/bindings/cedarling_wasm/js/scripts/build.mjs
duttarnab: Hi @faisalsiddique4400 do you think we can handle it elegantly.

faisalsiddique4400: @duttarnab yes, we can put the path in one place. Same mechanism as the `@/*` and `Components/*` aliases in admin-ui, just one entry in `tsconfig.json`:

```json
"paths": { "@generated/*": ["../pkg/*"] }
```

then `src/` imports `@generated/cedarling_wasm.js` and renaming the folder is a one-line change in `tsconfig.json` instead of a search across files.

Every one of the `src/` references is type-only (`import type`, `typeof import(...)`, `export type *`), so they're erased before esbuild ever runs and the compiled output has no trace of the alias.

@Dahkenangnon it stays a static specifier, so I don't think it runs into the resolution problem you mentioned. One knock-on: the `replaceAll` in `rewriteDeclaration` currently looks for `../../pkg/cedarling_wasm.js`, so that needle would become `@generated/cedarling_wasm.js`.

The few real filesystem paths in `build.mjs` could just be consts in one file.

faisalsiddique4400:
One maintainability suggestion: a linter and formatter for `js/`. There's no eslint, prettier or editorconfig config in the package, and `npm run check` runs tests plus package verification only.

Concrete payoff, it would have caught a live issue here on its own. `verify-consumer.mjs:190` declares `const output`, which shadows the `output` handler defined at line 64, so the `.catch(output)` on line 255 receives a string and silently does nothing. Line 80 shadows the same way, latent for now. `no-shadow` flags both.

While on tooling, `build.mjs` already sets `metafile: true` on all five esbuild builds (216, 229, 242, 255, 268) but only reads one of them for the edge import check. Since the metafiles are being generated anyway, a bundle report is nearly free. Sonda has an esbuild entrypoint and would give size and treemap output across all five bundles, which seems worth having for a package that ships five entries plus the wasm.

We run eslint 9 flat config with prettier 3 and Sonda in admin-ui, so both setups carry over nearly as-is if that's a useful starting point.

CC: @duttarnab

Dahkenangnon:
Hi @duttarnab  & @faisalsiddique4400 , thanks for your comments.

**Accepted**: I improved both guides and source-build instructions, completed `jsonString()` naming, centralized generated paths, fixed subprocess error output, added unsigned/multi-issuer coverage, and integrated Prettier/EditorConfig.

**Rejected**: Renaming canonical data fields/getters, removing approved `js_name` annotations, the unsupported docstring threshold, and adding Sonda without a demonstrated need.

**Deferred**: ESLint, legacy-package removal, external documentation, consumer migrations, and maintained examples—to preserve scope and the line budget. Versioning, migration communication, and nightly fallback remain release-owner follow-ups, with compatibility decisions required before publication.

faisalsiddique4400:
Hi @Dahkenangnon that covers a lot of ground.

Two points are still open from my side.

**ESLint.** The shadowed `output` binding has been fixed here, but `no-shadow` is the rule which catches this whole class of issue. I ran it on the pre-fix file with a 6-line config:

```
 80:9  error  'output' is already declared in the upper scope on line 64 column 10  no-shadow
190:9  error  'output' is already declared in the upper scope on line 64 column 10  no-shadow
```

Both were reported at write time, so that nobody needs to find them by reading 516 lines. The setup is a flat config plus `eslint`, `@eslint/js` and `typescript-eslint`. The code is new, so there is no backlog of violations to clear. Later this will not be the case.

**Rename list.** You have put versioning under release-owner follow-ups and I agree with that. The part which is harder to recover later is the mapping itself. Please add an old-name to new-name table in the PR description, so that whoever handles the release need not rebuild it from the `lib.rs` diff.

Tracking issues are fine for the rest. One point though: `cedarling_js` is already broken on this branch. `package.json:42` points to `file:../cedarling_wasm/pkg` and it uses 15 of the renamed symbols. 14 of these are string lookups in `engine/generated-wrapper.ts`, so the package type-checks cleanly and fails only at runtime. The `cedarling_js_tests` job which would have caught this is the one repurposed into `cedarling_wasm_test`.

CC: @duttarnab

Dahkenangnon:
I agree with adding EsLint, doing so will make the PR exceed it diff lines budget (5000 lines), my suggestion is to add it as a follow up PR. The same reason prevent me to remove the old cedarling_js. Do you think that's ok ?

I updated the PR about the api renaming.

jans-cedarling/bindings/cedarling_wasm/js/README.md
```diff
@@ -0,0 +1,494 @@
...
+
+| Environment                                 | Import                                              | WebAssembly setup           |
+| ------------------------------------------- | --------------------------------------------------- | --------------------------- |
+| Browser applications and supported bundlers | `@janssenproject/cedarling_wasm`                    | Automatic                   |
+| Node.js ESM                                 | `@janssenproject/cedarling_wasm`                    | Automatic                   |
+| Node.js CommonJS                            | `require("@janssenproject/cedarling_wasm")`         | Automatic                   |
+| Cloudflare Workers and Vercel Edge          | `@janssenproject/cedarling_wasm/edge`               | Automatic static module     |
+| Other ESM bundlers                          | `@janssenproject/cedarling_wasm/manual` and `/wasm` | Application emits the asset |
+
+The root export selects a browser, Node ESM, or Node CommonJS build. It is
```
duttarnab: The package root automatically selects the appropriate build for browsers, Node.js ESM, or Node.js CommonJS, so most applications should import Cedarling directly from the package root:

```
//Browser or Node.js with ESM
import { init } from "@janssenproject/cedarling_wasm"; 

//Node.js with CommonJS
const { init } = require("@janssenproject/cedarling_wasm"); 

```

For edge environments such as Cloudflare Workers, Vercel Edge, or other ESM-based bundlers, use the public /edge, /manual, and /wasm entry points when needed. For example:

```
import initWasm from "@janssenproject/cedarling_wasm/manual";
import wasmUrl from "@janssenproject/cedarling_wasm/wasm";
````

File: jans-cedarling/bindings/cedarling_wasm/js/README.md [hunk 21]
```diff
@@ -0,0 +1,494 @@
...
+qualified as an installed browser bundle with esbuild and as installed Node
+ESM/CommonJS consumers, so most applications should use it directly. Do not
+import files below `dist/`: only the package root, `./edge`, `./manual`, and
+`./wasm` are public entry points.
+
+## Key concepts
```
duttarnab: This section **(Key concepts)** is not required. This already have been covered in detail in Cedarling docs.

jans-cedarling/bindings/cedarling_wasm/js/README.md
```diff
@@ -0,0 +1,494 @@
...
+[Cedarling authorization methods](https://docs.jans.io/stable/cedarling/reference/cedarling-authz/),
+[policy stores and trusted issuers](https://docs.jans.io/stable/cedarling/reference/cedarling-policy-store/),
+and the [Context Data API](https://docs.jans.io/stable/cedarling/reference/cedarling-interfaces/#context-data-api).
+
+</details>
+
+## Quick start: unsigned authorization
+
+To create a Cedarling instance, call `init()` with an object or `Map` containing
+canonical uppercase Cedarling bootstrap properties. Use a Cedar archive URL for
```
duttarnab: > Use a Cedar archive URL for the policy store.

Better remove this line as there are other ways to define policy store.

jans-cedarling/bindings/cedarling_wasm/js/README.md
```diff
@@ -0,0 +1,494 @@
...
+| Multi-issuer authorization               | Policy evaluation using one or more tokens, potentially from different issuers, with validation governed by configuration. Token entities are available through `context.tokens`; this method does not use a principal entity. |
+
+References: [Cedar's principal, action, resource, and context model](https://docs.cedarpolicy.com/auth/authorization.html),
+[Cedarling authorization methods](https://docs.jans.io/stable/cedarling/reference/cedarling-authz/),
+[policy stores and trusted issuers](https://docs.jans.io/stable/cedarling/reference/cedarling-policy-store/),
+and the [Context Data API](https://docs.jans.io/stable/cedarling/reference/cedarling-interfaces/#context-data-api).
+
+</details>
+
+## Quick start: unsigned authorization
```
duttarnab: Can we keep the rest of the docs in below format. Cover the necessary instruction related to the browsers, Node.js ESM, or Node.js CommonJS ,  edge environments or other ESM-based bundlers

https://docs.jans.io/stable/cedarling/tutorials/javascript/#getting-started-with-cedarling-in-a-javascript-app

jans-cedarling/bindings/cedarling_wasm/js/README.md
```diff
@@ -0,0 +1,494 @@
...
+configuration and trusted-issuer settings required by its security model.
+
+## Configuration
+
+The package passes raw bootstrap properties to Cedarling without a
+JavaScript-specific configuration layer. Refer to the
+[Cedarling property reference](https://docs.jans.io/stable/cedarling/reference/cedarling-properties/)
+for the full, authoritative property contract.
+
+<details>
```
duttarnab: This bootstrap properties detail is not required in README. It has been already explained in Cedarling docs.

jans-cedarling/bindings/cedarling_wasm/js/docs/maintainer.md
```diff
@@ -0,0 +1,72 @@
...
+## Generated input and package output
+
+Run this command from `bindings/cedarling_wasm` after changing an exported Rust
+binding:
+
+```sh
+wasm-pack build --release --locked --target web --scope janssenproject
+```
+
+It compiles the Rust WebAssembly binding and emits the ignored `../pkg/` input:
```
duttarnab: **Replace with below content.**

This generates the ../pkg/ directory, which contains the JavaScript glue code, TypeScript declarations, and WebAssembly binary required to build the package.

The handwritten package source is maintained under js/. The package build consumes the generated files from ../pkg/ and produces the publishable distribution under dist/:


```
dist/
├── browser/
├── esm/
├── cjs/
├── edge/
├── manual/
├── types/
│   ├── esm/
│   └── cjs/
└── wasm/
    └── cedarling_wasm_bg.wasm
```

The published package contains a single WebAssembly binary. The browser, Node.js, CommonJS, edge, and manual entry points use this same binary through their respective loading mechanisms.

**Important:**
Do not edit files in pkg/ or dist/ directly. These directories contain generated build artifacts and must be regenerated when the source changes.
Treat js/ as the source of truth for handwritten package code, build configuration, and documentation.
Consumers should use only the package's public export paths: the package root, ./edge, ./manual, and ./wasm.
Files under dist/ are implementation details and are not intended to be imported directly.

jans-cedarling/bindings/cedarling_wasm/js/docs/maintainer.md
```diff
@@ -0,0 +1,72 @@
...
+the generated JavaScript glue, declarations, and WebAssembly binary. `js/`
+contains the handwritten package assembly. Its build reads that input and emits
+the ignored, publishable distribution:
```
duttarnab: Please go through the contents of the doc and see how it can be more consumable for a novice. Also cover the instructions to run test cases.

jans-cedarling/bindings/cedarling_wasm/js/tests/fixtures/multi-issuer.mjs
```diff
@@ -0,0 +1,143 @@
+// Follows the shared multi-issuer fixtures: token tags drive policy decisions.
+const issuers = ["Acme", "Dolphin"];
+const issuerUrl = (name) => `https://${name.toLowerCase()}.example.invalid`;
+const policyStore = {
```
duttarnab: Can we read policy-store from a separate `.yml` file for test cases.

jans-cedarling/bindings/cedarling_wasm/js/tests/browser.test.mjs
```diff
@@ -0,0 +1,142 @@
+import assert from "node:assert/strict";
```
duttarnab, line 2: Add comments at the starting of the file describing what test-cases it covers in what environment.

jans-cedarling/bindings/cedarling_wasm/js/tests/multi-issuer.test.mjs
```diff
@@ -0,0 +1,76 @@
+import {
```
duttarnab, line 2: Add comments at the starting of the file describing what test-cases it covers.

jans-cedarling/bindings/cedarling_wasm/js/tests/runtime.test.mjs
```diff
@@ -0,0 +1,248 @@
+import assert from "node:assert/strict";
```
duttarnab, line 2: Add comments at the starting of the file describing what test-cases it covers in what environment.

duttarnab (changes requested):
Please see the inline comments.

jans-cedarling/bindings/cedarling_wasm/js/README.md
Dahkenangnon: Only publicly callable api are renamed to camelCase. Their args and output data shape are deliberately kept as it including canonical bootstrap-property keys, request JSON, serialized JSON, and generated data/getter properties like request_id, data_type, is_ok, item_index, and batch_id—remain.

jans-cedarling/bindings/cedarling_wasm/js/README.md
Dahkenangnon: Agree with this; Already fixed.

jans-cedarling/bindings/cedarling_wasm/js/docs/readme.md
Dahkenangnon: The file is now renamed to `maintainer.md` as it target reader are maintainer of the package.

jans-cedarling/bindings/cedarling_wasm/js/docs/readme.md
Dahkenangnon: I ack with this doc improvement, the section is now properly introduced. The `wasm-pack build` command whose purpose is compile the rust WebAssembly binding and emits the ignored `pkg/` is also described in the document.

jans-cedarling/bindings/cedarling_wasm/js/docs/readme.md
Dahkenangnon: Thanks for the reference

jans-cedarling/bindings/cedarling_wasm/js/README.md
Dahkenangnon: `locally` was referred to the environment where Cedarling is integrated (server, browser, ...). Thanks to mention this,  it can effectively be confused. It is now fixed

jans-cedarling/bindings/cedarling_wasm/js/README.md
Dahkenangnon: Nice, thanks for the reference.

jans-cedarling/bindings/cedarling_wasm/js/scripts/build.mjs
Dahkenangnon: thanks @faisalsiddique4400 , your suggestion  is worthy and already applied.

jans-cedarling/bindings/cedarling_wasm/js/scripts/build.mjs
Dahkenangnon: Duplicated, already covered above.

jans-cedarling/bindings/cedarling_wasm/js/src/cjs.ts
Dahkenangnon: Duplicated, already covered above.

jans-cedarling/bindings/cedarling_wasm/js/scripts/verify-consumer.mjs
Dahkenangnon: noted. already fixed

jans-cedarling/bindings/cedarling_wasm/js/tests/browser.test.mjs
Dahkenangnon: Both unsigned and multi-issuer authz are tested, please see [verify-consumer.mjs](https://github.com/JanssenProject/jans/blob/jans-cedarling-14899/jans-cedarling/bindings/cedarling_wasm/js/scripts/verify-consumer.mjs)


On removing `cedarling_js` in this current PR, while I fully ack that it is now broken, I suggest do that cleanup in a separate follow up PR to keep thing clean and avoid exceed the PR diff budget.

jans-cedarling/bindings/cedarling_wasm/js/README.md
Dahkenangnon: Is the above a section replacement suggestion ? Please update your comments or confirm  my understanding.

jans-cedarling/bindings/cedarling_wasm/js/README.md
Dahkenangnon: The reason I added this section to avoid new user, the need to search and read outside doc before understanding the most important thing they need to know before starting cedarling integration. It's really worth for npm user which land on the package readme. Also, the section is not a duplication of the official docs, it's just an  summary of must-to-know terms, used in the readme itself.

That section can  also be useful for ai assisted integration, as the readme file is shipped with the pkg.

I would like to read your feedback on this ?

jans-cedarling/bindings/cedarling_wasm/js/README.md
Dahkenangnon: These details are added because you suggested earlier in previous PR (please see https://github.com/JanssenProject/jans/pull/14635#pullrequestreview-4805785395 & https://github.com/JanssenProject/jans/pull/14635#issuecomment-5115886079)

Therefore, if you find that, for this pr, we shouldn't provide to user full details of what props init or other api expected in the readme, please justify why; Because I found your mentioned comments insightful as it let user read quickly the package readme to know what it need to pass to a given method or function.

But I can ack the duplication issue, but for me it's justified, because the content of this README is shipped directly to user

jans-cedarling/bindings/cedarling_wasm/js/docs/maintainer.md
Dahkenangnon: Thanks for the suggestion.

jans-cedarling/bindings/cedarling_wasm/js/docs/maintainer.md
Dahkenangnon: The ``## Build and qualify`` section already cover how to run test. Can you confirm it's ok for you ?

jans-cedarling/bindings/cedarling_wasm/js/tests/browser.test.mjs
Dahkenangnon: Thanks, I'll apply it.

jans-cedarling/bindings/cedarling_wasm/js/tests/multi-issuer.test.mjs
Dahkenangnon: Thanks, I'll apply it.

jans-cedarling/bindings/cedarling_wasm/js/tests/runtime.test.mjs
Dahkenangnon: Thanks, I'll apply it.

jans-cedarling/bindings/cedarling_wasm/js/tests/fixtures/multi-issuer.mjs
Dahkenangnon: I would avoid introducing a YAML parser dependency solely for this fixture: it adds tooling and lockfile changes within our tight PR diff budget. The WASM binding supports JSON and .cjar inputs; YAML file loading is not supported through its [filesystem configuration.](https://docs.jans.io/nightly/cedarling/reference/cedarling-properties/?h=#loading-the-policy-store)

Could we use a separate .cjar fixture instead, keeping its readable policy sources alongside it? This separates policy data from test logic and reuses the archive-loading approach already exercised by other tests.

jans-cedarling/bindings/cedarling_wasm/js/README.md
duttarnab: The below content under https://github.com/JanssenProject/jans/tree/jans-cedarling-14899/jans-cedarling/bindings/cedarling_wasm/js#choose-an-integration can be replaced by above suggested content.

---------

The root export selects a browser, Node ESM, or Node CommonJS build. It is qualified as an installed browser bundle with esbuild and as installed Node ESM/CommonJS consumers, so most applications should use it directly. Do not import files below dist/: only the package root, ./edge, ./manual, and ./wasm are public entry points.

jans-cedarling/bindings/cedarling_wasm/js/README.md
duttarnab: We cannot include everything covered in the Cedarling documentation in the binding's README. The README should focus only on the information relevant to the binding (see other binding docs). This content repetition should be avoided. If you think any additional information is essential for the user's understanding, you can provide a reference link to the relevant Cedarling documentation .

jans-cedarling/bindings/cedarling_wasm/js/README.md
duttarnab: Already replied here: https://github.com/JanssenProject/jans/pull/14900#discussion_r3979241818

jans-cedarling/bindings/cedarling_wasm/js/tests/fixtures/multi-issuer.mjs
duttarnab: The .cjar fixture in the test-case will not allow to quickly reveal its content. Better use .yaml. I guess Json poly-store support is going to be removed soon. cc: @olehbozhok

jans-cedarling/bindings/cedarling_wasm/js/README.md
Dahkenangnon: Ok, thank for your clarification

jans-cedarling/bindings/cedarling_wasm/js/README.md
Dahkenangnon: Ok, let's go with that direction.

jans-cedarling/bindings/cedarling_wasm/js/tests/fixtures/multi-issuer.mjs
Dahkenangnon: Ok, I'll use pre-parsed the yml into a cjar which will be passed to the init or initfrom bytes function.

Dahkenangnon:
@coderabbit review or approve accordingly

faisalsiddique4400:
@Dahkenangnon Please check the findings: 
25 of the 30 renamed methods are not asserted anywhere. This PR is the one doing the rename, so a typo in a `js_name` attribute would ship silently, as the Rust tests call the Rust names.

`verify-consumer.mjs:486` already declares `const cedarling: Cedarling` but only checks `free()`. Adding `void cedarling.popLogs;` and so on covers the rest in about 25 lines, compile-time only. Better here than deferred, since this is where the rename happens.
CC: @duttarnab

faisalsiddique4400:
@Dahkenangnon Please check the findings: 
25 of the 30 renamed methods are not asserted anywhere. This PR is the one doing the rename, so a typo in a `js_name` attribute would ship silently, as the Rust tests call the Rust names.

`verify-consumer.mjs:486` already declares `const cedarling: Cedarling` but only checks `free()`. Adding `void cedarling.popLogs;` and so on covers the rest in about 25 lines, compile-time only. Better here than deferred, since this is where the rename happens.

Dahkenangnon:
> @Dahkenangnon Please check the findings: 25 of the 30 renamed methods are not asserted anywhere. This PR is the one doing the rename, so a typo in a `js_name` attribute would ship silently, as the Rust tests call the Rust names.
> 
> `verify-consumer.mjs:486` already declares `const cedarling: Cedarling` but only checks `free()`. Adding `void cedarling.popLogs;` and so on covers the rest in about 25 lines, compile-time only. Better here than deferred, since this is where the rename happens.

You are right. Let's me cover them

jans-cedarling/bindings/cedarling_wasm/js/README.md
```diff
@@ -0,0 +1,410 @@
...
+
+For policy and schema authoring, use the
+[Agama Lab Policy Designer](https://cloud.gluu.org/agama-lab/dashboard/policy-designer).
+For packaging your policy store, see the
+[Cedar Archive (.cjar) format](https://docs.jans.io/nightly/cedarling/reference/cedarling-policy-store/#cedar-archive-cjar-format).
+
+```ts
+import { init } from "@janssenproject/cedarling_wasm";
+
+const cedarling = await init({
```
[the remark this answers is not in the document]
Dahkenangnon: Resolved in [01284cf](https://github.com/JanssenProject/jans/pull/14900/commits/01284cf3337b3b7d886f6d9879db24bc59aacbfe)

.github/workflows/test-cedarling.yml
```diff
@@ -260,9 +260,12 @@ jobs:
       working-directory: jans-cedarling
       run: cargo +stable clippy -p cedarling_wasm --target wasm32-unknown-unknown -- -D warnings
 
-  cedarling_js_tests:
+  cedarling_wasm_test:
```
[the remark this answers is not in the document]
Dahkenangnon: Won't restore cedarling_js, we decide to remove it entirely from the tree as when this pr will be merged, it will become dead and broken.  See [01284cf](https://github.com/JanssenProject/jans/pull/14900/commits/01284cf3337b3b7d886f6d9879db24bc59aacbfe)

Dahkenangnon:
@coderabbitai review, approve if ok

Dahkenangnon:
@coderabbitai , I ack that finding, it's acceptable and is not introduced by this PR. It's inherited from the main branch. So, it must not block the approval if everything other are ok

Dahkenangnon:
@coderabbitai recheck, all job are now finished

faisalsiddique4400:
@Dahkenangnon The only item left from my side is the documentation outside this package. A few pages under `docs/cedarling/` still show the old snake_case API, mainly `tutorials/javascript.md`, `reference/cedarling-policy-store.md` and `developer/mobile-apps/cedarling-uniffi.md`. A ticket for that is enough, it does not need to be in this PR.
Please create that ticket and share the URL here in the PR.

faisalsiddique4400 (dismissed):

duttarnab (dismissed):
Following will be covered in next PR (GH issue):
1. Documentation
2. Adding example/integration projects such as: 

```
examples/
├── react-vite/
├── webpack/
├── node-esm/
├── node-commonjs/
└── edge/
```

moabu (approved):

duttarnab (approved):

PR #14941, opened 2026-09-01
fix(jans-cedarling): reject entity type bound to conflicting token keys

### LINKED ISSUE #14747: fix(jans-cedarling): resolve token metadata key from issuer, not global scan
## Summary

In multi-issuer setups where two or more `TrustedIssuer`s share the same Cedar entity type (`mapping`, e.g. both map their access token to `Jans::Access_Token`), `required_claims` validation can be silently skipped for some issuers. The token key used to look up claim requirements is resolved by scanning **all** configured trusted issuers, not the specific issuer the token actually came from — so it can return a key that doesn't exist in that issuer's own metadata, and the check is skipped with no error and no log.

## Environment

- Component: `jans-cedarling` (`cedarling/src/jwt/mod.rs`, `cedarling/src/jwt/issuer_index.rs`)
- Applies to: multi-issuer JWT authorization (`authorize_multi_issuer`)

## Steps to Reproduce

1. Configure two trusted issuers in the policy store, both declaring `entity_type_name: "Jans::Access_Token"` for their access token, but under **different** internal `token_metadata` keys, e.g.:
   - Issuer A: `token_metadata["access_token"] = { entity_type_name: "Jans::Access_Token", required_claims: [...] }`
   - Issuer B: `token_metadata["at"] = { entity_type_name: "Jans::Access_Token", required_claims: [] }`
2. Submit a multi-issuer authorization request with a valid token from whichever issuer's key is *not* picked by the global lookup below (order depends on `HashMap` iteration, so this can vary between runs/reloads).
3. Observe the outcome.

## Expected Behavior

`required_claims` configured for the resolved issuer's token metadata should always be validated against the token, regardless of which other issuers share the same `mapping`.

## Actual Behavior

`required_claims` validation is silently skipped for the token — no error, no warning log — whenever the globally-resolved token key doesn't happen to match a key present in the actual issuer's own `token_metadata`.

## Root Cause

`JwtService::validate_single_token` resolves the token type key via:

```rust
// cedarling/src/jwt/mod.rs
fn find_token_metadata_key<'a>(&'a self, entity_type_name: &'a str) -> Cow<'a, str> {
    if let Some(token_key) = self.issuer_configs.find_token_metadata_key(entity_type_name) {
        return Cow::Owned(token_key);
    }
    Cow::Borrowed(entity_type_name)
}
```

which delegates to:

```rust
// cedarling/src/jwt/issuer_index.rs:62-75
pub(super) fn find_token_metadata_key(&self, entity_type_name: &str) -> Option<String> {
    // Look through all trusted issuers to find the matching entity type name
    let index = self.iss_index.read().expect(MUTEX_POISONED_ERR);
    for issuer_config in index.values() {
        for (token_key, token_metadata) in &issuer_config.policy.token_metadata {
            if token_metadata.entity_type_name == entity_type_name {
                return Some(token_key.clone());
            }
        }
    }
    None
}
```

This scans **every** configured trusted issuer and returns the first `token_key` whose `entity_type_name` matches — before the token's actual issuer is even considered. `HashMap` iteration order is unspecified, so the returned key is effectively arbitrary among issuers sharing the mapping.

That key is then looked up against the metadata of the **specific** trusted issuer the token's `iss` claim resolved to:

```rust
// cedarling/src/jwt/mod.rs, inside validate_single_token
if let Some(trusted_iss) = &trusted_iss {
    let token_type: Option<&str> = /* derived from token_kind, built from find_token_metadata_key */;
    if let Some(token_type) = token_type {
        if let Some(token_metadata) = trusted_iss.token_metadata.get(token_type) {
            validate_required_claims(&validated_jwt.claims, token_type, token_metadata)?;
        }
        // else: silently skipped — no error, no log
    }
}
```

If the globally-resolved key isn't present under that specific issuer's `token_metadata` (because it uses a different internal key name for the same `entity_type_name`), `.get(token_type)` returns `None` and the whole `required_claims` check is skipped without any signal.

Two different trusted issuers legitimately sharing a `mapping`, each with independent internal key names, is a supported configuration — nothing requires the internal `token_metadata` key to match across issuers, only the `entity_type_name` needs to match the request's `mapping`.

## Impact

- `required_claims` enforcement becomes non-deterministic (depends on `HashMap` iteration order, which can change across process restarts / policy-store reloads) whenever ≥2 trusted issuers share a mapping with different internal token keys.
- The failure mode is silent: no error, no warning log — a security-relevant check can quietly stop applying to a subset of issuers with no observable signal.
- Scoped to the multi-issuer / shared-mapping case; a single issuer per mapping is unaffected, since that issuer's own `token_metadata.get(token_type)` always finds its own key.

## Suggested Fix

Resolve the token type key from the **already-resolved** `trusted_iss` (the issuer the token's `iss` claim matched), instead of searching all issuers globally beforehand. E.g. search `trusted_iss.token_metadata` directly for the entry whose `entity_type_name == mapping`, rather than calling `find_token_metadata_key(mapping)` before the issuer is known.

dagregi:
### Prepare

-------------------

#### Target issue
  
  
  
  
closes #14747

#### Implementation Details
  

-------------------

## Summary by CodeRabbit

* **Bug Fixes**
  * Trusted issuer configuration validation rejects duplicate Cedar entity-type mappings within or across issuers.
  * Duplicate OIDC endpoints continue to be rejected.
  * Validation errors identify whether the conflict involves an OIDC endpoint or entity-type mapping, including relevant details.
  * Schema validation now detects mismatches between declared and expected issuer types.

* **Tests**
  * Added coverage for valid distinct mappings, duplicate configurations, and issuer-type mismatches.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/d322f505-66e4-40bb-bf54-1c99a0960c0e?refs=b7c134f9d15fdd1bd10f386f351bdda90f3206a6&source=prChecksComment) |

olehbozhok:
## Why the shared-`entity_type_name` pattern cannot work (and should be rejected)

### How a multi-issuer token entity is linked to its issuer

A validated multi-issuer token becomes a Cedar entity whose two relevant parts come from
**different configuration fields**:

| piece | source | example |
| --- | --- | --- |
| token entity type | `token_metadata.{key}.entity_type_name` (the request `mapping`) | `Jans::Access_Token` |
| `iss` attribute entity type | the issuer's `name` field | `Alpha::TrustedIssuer` / `Beta::TrustedIssuer` |

The `entity_type_name` field is what selects the Cedar entity type — see
[`docs/cedarling/reference/cedarling-policy-store.md`](docs/cedarling/reference/cedarling-policy-store.md)
("Token Metadata Schema"). The type is resolved by `determine_token_entity_type` (returns
`entity_type_name`), while the `iss` attribute is set by `add_iss_claim`:

```rust
// jans-cedarling/cedarling/src/entity_builder/build_multi_issuer_entity.rs (add_iss_claim, Jwt branch)
RestrictedExpression::new_entity_uid(
    EntityBuilder::trusted_issuer_cedar_uid(&token_iss.name, &issuer)?
)
```

`trusted_issuer_cedar_uid` builds the reference as `{name}::TrustedIssuer::{iss}`, i.e. it uses the
**issuer's own namespace**, not the token type's namespace.

### What the schema demands

A token entity that wants an issuer link declares `iss?: TrustedIssuer`. Inside `namespace Jans`,
`TrustedIssuer` resolves to `Jans::TrustedIssuer`. The schema-driven attribute builder
(`AttrSrc::EntityRef` in `build_entity_attrs.rs`) resolves `iss` against the already-built
entities using that schema-declared type, and errors with `MissingEntityRefs` when it is absent.

The canonical multi-issuer schema puts **each issuer in its own namespace** — see
[`docs/cedarling/reference/cedarling-multi-issuer.md`](docs/cedarling/reference/cedarling-multi-issuer.md)
("Cedar Schema for Multi-Issuer Tokens"), e.g. `namespace Acme { entity Access_token { iss?: TrustedIssuer } }`.

### The two concrete failure modes

For two issuers `Alpha` and `Beta` both mapped to the shared type `Jans::Access_Token`, the
schema can only declare **one** type for `iss`, so:

1. **`iss: TrustedIssuer`** (== `Jans::TrustedIssuer`): no issuer produces that type — Alpha/Beta
   emit `Alpha::TrustedIssuer` / `Beta::TrustedIssuer`. Building a token entity therefore fails for
   *both* issuers:

   ```
   MissingEntityRefs(["Jans::TrustedIssuer"])
   ```

2. **`iss: Alpha::TrustedIssuer`**: the schema path finds `Alpha::TrustedIssuer` (it exists), so the
   build *succeeds* — but `add_reserved_claims` then overwrites `iss` with the real issuer namespace,
   producing `Beta::TrustedIssuer` for Beta's token. That **silently** violates the schema (there is
   no post-build validation in `build_single_token_entity`).

> Note: a cross-namespace `iss: Acme::TrustedIssuer` **does** appear in the docs, but only on a
> `Workload` principal in the classic single-issuer `authorize` path, where a principal belongs to a
> single fixed issuer — see
> [`docs/cedarling/reference/cedarling-authz.md`](docs/cedarling/reference/cedarling-authz.md)
> (`namespace MyApp { entity Workload { iss: Acme::TrustedIssuer } }`). It is not a shared
> multi-issuer token type.

In both cases a single shared type cannot express a correct `iss` link for more than one issuer
namespace.

### Conclusion

A shared `entity_type_name` (`Jans::Access_Token` for Alpha and Beta) is buildable only when the
schema **omits `iss`** on that shared type — which means the token loses its issuer linkage
altogether. As soon as `iss` is declared, the config either fails loudly or silently mis-types the
reference.

The correct, supported model is therefore **one entity type per issuer namespace**
(`Acme::Access_token` with `name: "Acme"`, `Google::Id_Token` with `name: "Google"`, …), so that
`entity_type_name` and `iss` live in the same namespace.

For that reason the shared-mapping configuration should not be treated as a supported pattern, and
rejecting it at startup (this PR) converts what would otherwise be a silent, hard-to-diagnose
mis-configuration into an explicit, actionable error.

The unit test `test_two_issuers_sharing_entity_type_with_iss_in_schema_fails_to_build` pins the
loud-failure path (`MissingEntityRefs("Jans::TrustedIssuer")`).

olehbozhok:
It means that we should have only one entity per one (any) token.

dagregi:
> It means that we should have only one entity per one (any) token.

Changed it in https://github.com/JanssenProject/jans/pull/14941/commits/e99e9f31f7a1e8706bb7771c904d76483e2801d0

olehbozhok:
## Review — `enforce global uniqueness of Cedar entity types`

The pivot from "reject conflicting token keys" to `DuplicateEntityType` (reject **any** shared
`entity_type_name`, regardless of the key) is the right call. It matches the root cause: a shared
entity type cannot build a correct `iss` reference across different issuer namespaces. Removing
the integration test that asserted shared mapping "works" is also correct — that test was pinning
broken behavior.

### Gap the current check does not cover

`DuplicateEntityType` catches *sharing*, but the underlying invariant is narrower and is not yet
enforced: the token's `iss` attribute is always built as `{name}::TrustedIssuer`
(`trusted_issuer_cedar_uid(&token_iss.name, ...)` in `add_iss_claim`). So a **single** issuer with
a mismatched namespace — e.g. `name: "Alpha"` with `entity_type_name: "Jans::Access_token"` —
still passes validation and then either:

- fails at entity-build time with `MissingEntityRefs("Jans::TrustedIssuer")` (when the schema
  declares `iss`), or
- silently mis-builds when `iss` is not declared / no schema is present.

### Suggested follow-up (two checks)

**1. `iss` type check — hard error (deterministic)**

For each trusted issuer's token metadata, inspect the schema shape of `entity_type_name`
(`MappingSchema::get_entity_shape`, `entity_builder/schema.rs`). If it declares `iss` as an entity
reference (`AttrSrc::EntityRef`), the resolved type must equal `{name}::TrustedIssuer`. Any other
type is a deterministic build failure and can be rejected at startup.

**2. namespace heuristic — warning (non-fatal)**

When there is no schema or `iss` is not declared, log a warning if the namespace of
`entity_type_name` differs from `name`:

> issuer 'Alpha' declares token entity type 'Jans::Access_token' in a namespace different from
> its name 'Alpha'; the token's iss will reference 'Alpha::TrustedIssuer' — ensure the schema
> declares a matching iss type.

### Where

`validate_trusted_issuers_config` stays pure (hard errors: duplicate entity type / duplicate OIDC
endpoint). The `iss`-type check needs schema access, so it fits in `build_authz` (has
`policy_store.schema` + `log`) or `EntityBuilder::new` (already builds the `MappingSchema`).

dagregi:
> ## Review — `enforce global uniqueness of Cedar entity types`
> 
> The pivot from "reject conflicting token keys" to `DuplicateEntityType` (reject **any** shared `entity_type_name`, regardless of the key) is the right call. It matches the root cause: a shared entity type cannot build a correct `iss` reference across different issuer namespaces. Removing the integration test that asserted shared mapping "works" is also correct — that test was pinning broken behavior.
> ### Gap the current check does not cover
> 
> `DuplicateEntityType` catches _sharing_, but the underlying invariant is narrower and is not yet enforced: the token's `iss` attribute is always built as `{name}::TrustedIssuer` (`trusted_issuer_cedar_uid(&token_iss.name, ...)` in `add_iss_claim`). So a **single** issuer with a mismatched namespace — e.g. `name: "Alpha"` with `entity_type_name: "Jans::Access_token"` — still passes validation and then either:
> 
>     * fails at entity-build time with `MissingEntityRefs("Jans::TrustedIssuer")` (when the schema
>       declares `iss`), or
> 
>     * silently mis-builds when `iss` is not declared / no schema is present.
> 
> 
> ### Suggested follow-up (two checks)
> 
> **1. `iss` type check — hard error (deterministic)**
> 
> For each trusted issuer's token metadata, inspect the schema shape of `entity_type_name` (`MappingSchema::get_entity_shape`, `entity_builder/schema.rs`). If it declares `iss` as an entity reference (`AttrSrc::EntityRef`), the resolved type must equal `{name}::TrustedIssuer`. Any other type is a deterministic build failure and can be rejected at startup.
> 
> **2. namespace heuristic — warning (non-fatal)**
> 
> When there is no schema or `iss` is not declared, log a warning if the namespace of `entity_type_name` differs from `name`:
> 
> > issuer 'Alpha' declares token entity type 'Jans::Access_token' in a namespace different from
> > its name 'Alpha'; the token's iss will reference 'Alpha::TrustedIssuer' — ensure the schema
> > declares a matching iss type.
> 
> ### Where
> 
> `validate_trusted_issuers_config` stays pure (hard errors: duplicate entity type / duplicate OIDC endpoint). The `iss`-type check needs schema access, so it fits in `build_authz` (has `policy_store.schema` + `log`) or `EntityBuilder::new` (already builds the `MappingSchema`).

Added the fix in https://github.com/JanssenProject/jans/pull/14941/commits/bb48408672d55078102715c74b349803cbb3e961

olehbozhok (approved):
## Review — `reject mismatched iss entity type at startup`

### The `iss`-type check looks correct

`EntityBuilder::validate_iss_types` is a precise implementation: for each issuer it computes
`expected = {name}::TrustedIssuer` (the same `trusted_issuer_typename` used by `add_iss_claim`,
so no drift) and compares it against the schema's declared `iss` type via
`AttrSrc::EntityRef`.

Notably, this is **more precise than a namespace heuristic** and has no false positives:

- It rejects only when the schema declares `iss` as an entity reference to a type other than
  `{name}::TrustedIssuer`.
- It correctly **allows** a cross-namespace token type when `iss` is explicitly declared to match —
  e.g. the Python example `policy-store-multi-issuer` has `entity_type_name: "Jans::Access_Token"`
  but `iss: JansTestIssuer::TrustedIssuer` with `name: "JansTestIssuer"`, which passes.
- It skips the no-schema / no-`iss` / non-entity-ref cases, so schema-less configs (like the JS
  multi-issuer fixture) are unaffected.

The error message carries the issuer name, schema type, and expected type, and there is a negative
unit test (`rejects_iss_type_from_foreign_namespace`); the positive path is covered by the existing
schema-based tests.

### Remaining: one stale JSON fixture (moot — JSON policy store is being removed)

`f6c046541ab` fixed `bindings/cedarling_uniffi/test_files/policy-store.json`, but a tracked copy
still has the old issuer name:

`bindings/cedarling_uniffi/javaApp/src/main/resources/config/policy-store.json`
- `"name": "TestIss"` with `entity_type_name: "Jans::Access_token"` / `"Jans::Id_token"` /
  `"Jans::Userinfo_token"`, and an embedded schema declaring `iss: TrustedIssuer`
  (→ `Jans::TrustedIssuer`).
- This fails with `IssTypeMismatch` (`TestIss::TrustedIssuer` vs `Jans::TrustedIssuer`).

Note on scope: this file is only loaded by the manual `Main.java` demo
(`CEDARLING_POLICY_STORE_LOCAL_FN: "./src/main/resources/config/policy-store.json"`), **not** by
CI — the `Benchmark` job runs `org.example.Benchmark`, which resolves policy stores from
`bindings/benchmarks/fixtures/scenarios.json` (→ `test_files/policy-store-multi-issuer-*.yaml`).
Those are consistent (`JansTestIssuer` with `iss: JansTestIssuer::TrustedIssuer`; `Acme`/`Dolphin`
per-namespace), so the new check does not break CI.

Since JSON policy store support is scheduled for removal in a follow-up issue, this `.json`
fixture (and the `Main.java` demo loading it) is going away anyway — no separate fix is required.

### Nitpicks (non-blocking)

- The code comment "a string/claim `iss` accepts the emitted value regardless" is slightly
  misleading: for JWT tokens a schema-declared `iss: String` is still overwritten with an
  `EntityUid` by `add_reserved_claims`, so it is a silent mismatch rather than "accepted".
- The optional namespace-mismatch `WARN` (for schema-less / no-`iss` configs) is still not
  implemented; fine to leave out given the deterministic check above.

dagregi:
@coderabbitai approve

olehbozhok (changes requested):
## Replace the O(n) token-metadata reverse lookup with a precomputed index

### Context

The multi-issuer path validates tokens through
`JwtService::validate_multi_issuer_tokens`. For every JWT token it needs to
resolve the token metadata key (`access_token`, `id_token`, …) from the request
`mapping`, which carries the Cedar entity type name (e.g. `Dolphin::Access_Token`).

That reverse mapping is currently done by `IssuerIndex::find_token_metadata_key`
(`cedarling/src/jwt/issuer_index.rs`), which linearly scans **every trusted
issuer and every one of its token metadata entries** until it finds a match:

```rust
for issuer_config in index.values() {
    for (token_key, token_metadata) in &issuer_config.policy.token_metadata {
        if token_metadata.entity_type_name == entity_type_name {
            return Some(token_key.clone());
        }
    }
}
```

This runs once per token on every authorization request.

### Why this can now be made O(1)

Two recent changes make the lookup unambiguous:

1. `validate_trusted_issuers_config` (`cedarling/src/common/policy_store.rs`)
   enforces global uniqueness of Cedar entity types — the same
   `entity_type_name` cannot be owned by two tokens/issuers.
2. `EntityBuilder::validate_iss_types` (`cedarling/src/entity_builder/mod.rs`)
   rejects token types whose schema `iss` references a foreign namespace.

As a result, `entity_type_name -> token_key` is now a **bijection**: at most one
issuer and one token key match any given entity type name. The "first match
wins" linear scan is therefore correct but unnecessary.

### Proposal

Precompute a reverse index at `JwtService` initialization:

```
entity_type_name -> (issuer_id, token_key)
```

and replace the scan in `find_token_metadata_key` with an O(1) lookup. The
existing `TODO` in `issuer_index.rs` already calls for this. A side benefit:
the `Cow` fallback in `JwtService::find_token_metadata_key` (which returns the
entity type name itself when no key is found) becomes unreachable for a valid
config and can be removed or turned into a defensive error.

### Out of scope

- The `seen_combinations` duplicate detection in `validate_multi_issuer_tokens`
  (same issuer + mapping appearing twice in one request) is unrelated and must
  stay.
- Custom (non-JWT) issuers resolve their metadata through `CustomIssuerIndex`,
  not through this reverse lookup; this issue is JWT-only.

olehbozhok (approved):
We will make an index change in the https://github.com/JanssenProject/jans/issues/15009

olehbozhok:
@coderabbitai approve

PR #14972, opened 2026-09-04
perf(ci): cut AIO integration build time

mo-auto:
Run [33770049400](https://github.com/JanssenProject/jans/actions/runs/33770049400) burned 98 of its 150 min on the build before the first test ran; both matrix legs were killed by `timeout-minutes: 150` mid-suite.

### Build time

| change | saves |
|---|---|
| `-Dmaven.gitcommitid.skip=true` — the plugin sat 8 min per module against the shallow checkout, writing only cosmetic `git.properties` | ~8m |
| skip `*-fips` reactors (`-pl '!server-fips'`) — packaging-only WARs; the service images fetch the plain `jans-<svc>-<ver>.war` | ~33m |
| build the five service images concurrently instead of in a serial loop | ~10m |
| `-T 1C` — the reactor was serial on an 8-core VM | build-bound |
| dedicated-CPU droplet (`g-8vcpu-32gb`/`c-8`), falling back to `s-8vcpu-16gb` | removes CPU steal |
| `actions/cache` for `~/.m2` (minus `io/jans`) + cargo registry, staged in the rsync'd checkout | cold-start downloads |
| unit-suite `timeout` 900s → 420s | six suites at 900s alone exceeded the budget |

Cache is deliberately bounded, since the 10 GB budget is shared with the other workflows:
- downloaded dependencies only — `jans-cedarling/target` would take the entry to ~8 GB
- `io/jans` never cached; a stale `0.0.0-nightly` from another commit would resolve silently
- saved only on non-PR runs and only from the first matrix leg — a PR's cache is scoped to its own ref, so it would evict entries still in use while being readable by nothing. PRs still restore via the base-branch fallback
- weekly key: entries are immutable, so a content hash would write a fresh ~3 GB entry per dependency change

### Droplet provisioning

Both legs of [33869974704](https://github.com/JanssenProject/jans/actions/runs/33869974704) failed ~7s after creating their SSH key with `422 ... are invalid key identifiers`. The key is now polled before first use, and that specific 422 retries the same region 5 times before aborting. Every other non-2xx still aborts on the first attempt, so a 5xx is never re-POSTed into a duplicate droplet.

### AIO nginx diagnostics

The CIBA `bc-authorize` negative tests fail with an HTML 400 that the AIO's own nginx generates — indistinguishable from an upstream 400, because nginx logs its rejection reasons at `info` and the image sets no `error_log`. Adds `CN_AIO_NGINX_LOG_LEVEL` plus four buffer vars, **emitted only when set**, so an unset var leaves nginx on its own defaults (which for the buffers are page-size dependent). Values are validated before rendering — a stray `;` would otherwise inject arbitrary directives. CI opts into `info` and wider buffers, the latter for the `upstream sent too big header` 502s in the same run.

Parallel builds interleave output, so both the reactor (`-T 1C`) and the image builds read less linearly; each image gets its own log under `aio-logs/`.

## Summary by CodeRabbit

* **Performance**
  * Improved integration build times through reusable build caches and parallelized build and service-image preparation.
  * Reduced unit-test timeout limits for faster feedback.

* **Reliability**
  * Added fallback provisioning across dedicated-CPU and Basic compute sizes and multiple regions.
  * Improved SSH setup retry handling and integration-test result collection.
  * Preserved build caches after integration runs where enabled.
  * Improved consistency across supported build and integration-test environments.

* **Configuration**
  * Added configurable Nginx logging and proxy buffer settings, with validation for supported values.

### DIFF
--- a/.github/workflows/build-ci-snapshot.yml
+++ b/.github/workflows/build-ci-snapshot.yml
@@ -1,0 +1,130 @@
+name: "Build: CI droplet snapshot"
+
+# Bakes the toolchain the integration run installs on every VM (JDK, maven, docker, the pinned rust
+# toolchain, protoc) into a DigitalOcean snapshot. test-integration.yml boots from the newest
+# jans-ci-base-* snapshot and falls back to stock Ubuntu when none exists, so this is optional.
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 2 * * 0"
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
+jobs:
+  snapshot:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Harden runner
+        uses: step-security/harden-runner@8d3c67de8e2fe68ef647c8db1e6a09f647780f40 # v2.19.0
+        with:
+          egress-policy: audit
+
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Build and snapshot
+        env:
+          DO_TOKEN: ${{ secrets.DO_TOKEN }}
+        run: |
+          set -euo pipefail
+          [ -n "${DO_TOKEN:-}" ] || { echo "::error::DO_TOKEN secret is not set"; exit 1; }
+          api() { curl -sS --connect-timeout 10 --max-time 60 -H "Authorization: Bearer ${DO_TOKEN}" \
+                    -H "Content-Type: application/json" "$@"; }
+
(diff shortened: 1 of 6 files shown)


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/83ec2f56-d188-4b61-88b6-ea74f6403e17?refs=73ed6786b2f1b897b899244e86685e11d16c2a61&source=prChecksComment) |

mo-auto:
Error: Hi @mo-auto, You did not reference an open issue in your [PR](#14972). I attempted to create an [issue](#14973) for you.
Please update that [issues'](#14973) title and body and make sure I correctly referenced it in the above PRs body.

moabu:
@coderabbitai approve

moabu (approved):

PR #14975, opened 2026-09-04
fix(jans-fido2): decode RSA and EC2 key variants that SignatureVerifier already supports

imran-ishaq:
### Prepare

- [x] Read [PR guidelines](https://github.com/JanssenProject/jans/blob/main/docs/CONTRIBUTING.md#prs)
- [x] Read [license information](https://github.com/JanssenProject/jans/blob/main/LICENSE)

-------------------

#### Target issue

closes #14932

Sub-issue of #14925, under #14891. Follows #14950 (#14929), #14958 (#14931) and #14970 (#14924), all merged.

#### Implementation Details

The key decoder was narrower than the signature verifier, so our advertised capability overstated what
actually worked. In `CoseService.createUncompressedPointFromCOSEPublicKey()` the RSA branch accepted only
`RS256` and `RS65535` while `RS384`, `RS512`, `PS256`, `PS384` and `PS512` all threw, and the EC2 branch
accepted only `ES256` while `ES384` and `ES512` threw. `SignatureVerifier.getSignatureChecker()` implements
every one of those, so the verifier could handle a credential the decoder refused to build a key for — the
same mismatch that caused the EdDSA failure in #14929, in a different branch.

Both branches now check membership of an explicit capability set instead of a hand-maintained `switch`:

```java
DECODABLE_RSA_ALGORITHMS = EnumSet.of(RS256, RS384, RS512, RS65535, PS256, PS384, PS512);
DECODABLE_EC2_ALGORITHMS = EnumSet.of(ES256, ES384, ES512);
```

No new decode paths were needed. Every RSA key reconstructs from the same `(n, e)` pair and every EC2 key
from the same `(crv, x, y)` triple, so the algorithm only ever selected which reconstruction to run — the
narrowness was the list, not the code. Net 19 insertions, 11 deletions.

This only works end to end because the two changes it sits on are already in: #14958 added P-384 and P-521
to the curve mapping, so an `ES384` credential's curve now resolves, and #14970 corrected `ES384` to `-35`
and `ES512` to `-36`, so a configured `ES384` now advertises the code point it claims to.

`server-fips/` needs no mirroring: the module is `pom.xml` and `target/` only, with no `src/` tree — its
antrun step copies `${server.webapp.dir}` from the `server` module.

-------------------
### Test and Document the changes
- [x] Static code analysis has been run locally and issues have been fixed
- [x] Relevant unit and integration tests have been added/updated
[...]

PR #14974, opened 2026-09-04
fix(jans-fido2): read and enforce the crossOrigin member of client data

imran-ishaq:
### Prepare

- [x] Read [PR guidelines](https://github.com/JanssenProject/jans/blob/main/docs/CONTRIBUTING.md#prs)
- [x] Read [license information](https://github.com/JanssenProject/jans/blob/main/LICENSE)

-------------------

#### Target issue

closes #14939

Sub-issue of #14938. Unrelated to the #14891 algorithm-table tree — this is client-data verification, and
none of #14891's acceptance criteria touch it.

#### Implementation Details

WebAuthn Level 3 requires the Relying Party to inspect the `crossOrigin` member of `CollectedClientData`
during both registration and authentication verification. We never read it:

```
grep -rn "crossOrigin\|topOrigin" --include=*.java   # across the entire monorepo
-> zero hits
```

`CommonVerifiers.verifyClientJSON()` required `challenge`, `origin` and `type` to be present, validated the
challenge as base64url, optionally validated `tokenBinding`, and checked `origin` was a non-empty string.
That was the whole of it, so a ceremony performed inside a cross-origin iframe was accepted exactly as if it
had been performed at the top level. The `origin` check we do run compares against the RP's configured
origins, so this was not an open door — but it is precisely the check L3 adds for the framed case.

This adds `verifyCrossOrigin(...)` to `verifyClientJSON()`:

- an absent `crossOrigin` member is treated as `false`, per spec
- `crossOrigin = true` is rejected with a logged error and a distinct error type
- a non-boolean `crossOrigin` is rejected as `invalid_request`, so a string is never silently read as false

One guard covers both ceremonies: `verifyClientJSON` is called from `AttestationService:263` (registration)
and `AssertionService:366` (assertion), which is exactly the pair L3 names.

**New public error value, worth a reviewer's eye.** #14939 asks for "a distinct, logged error rather than a
generic invalid-request" but leaves the mechanism open. I added
`CommonErrorResponseType.CROSS_ORIGIN_NOT_ALLOWED` (`cross_origin_not_allowed`) rather than reusing
`invalid_request`, so a client can tell a framed-ceremony rejection from a malformed challenge. It went in
`CommonErrorResponseType` because `verifyClientJSON` serves both ceremonies — a ceremony-specific enum would
have forced the verifier to branch on which ceremony called it. This value is visible to clients in the
`{status, errorMessage}` envelope, so it is effectively API surface.

[...]

PR #14970, opened 2026-09-04
fix(jans-fido2): correct COSE algorithm code points and guard unknown-algorithm lookups

imran-ishaq:
### Prepare

- [x] Read [PR guidelines](https://github.com/JanssenProject/jans/blob/main/docs/CONTRIBUTING.md#prs)
- [x] Read [license information](https://github.com/JanssenProject/jans/blob/main/LICENSE)

-------------------

#### Target issue

closes #14924

Sub-issue of #14891. Independent of the other children; #14932 and #14936 both depend on it.

#### Implementation Details

Five constants held code points belonging to unrelated entries in the
[IANA COSE Algorithms registry](https://www.iana.org/assignments/cose/cose.xhtml#algorithms), and two of
them silently advertised a different algorithm than the admin configured.

`AttestationService.resolveEc2NumericValue()` resolves a configured name through
`CoseEC2Algorithm.valueOf(...).getNumericValue()` straight into `pubKeyCredParams`, so these values go out
on the wire:

| Constant | Was | Now | −260 / −261 / −262 actually are |
|---|---|---|---|
| `CoseEC2Algorithm.ES384` | −36 | **−35** | |
| `CoseEC2Algorithm.ES512` | −37 | **−36** | |
| `CoseEC2Algorithm.ED256` | −260 | *removed* | WalnutDSA |
| `CoseEC2Algorithm.ED512` | −261 | *removed* | TurboSHAKE128 |
| `CoseRSAAlgorithm.RS1` | −262 | *removed* | TurboSHAKE256 |

Configuring `ES384` advertised the code point for ES512; configuring `ES512` advertised `-37`, which is
`PS256` — an RSA algorithm. Our own two enums assigned `-37` to two different algorithms, so this needed no
external source to confirm. `RS1` is a duplicate rather than a wrong value: its real code point is `-65535`,
already present as `RS65535`, so the constant is deleted rather than corrected. `PS256/384/512` and
`RS256/384/512` were already right.

`CoseEdDSAAlgorithm.Ed25519(-8)` is renamed to `EdDSA`. `-8` is EdDSA; the v2.3 table's separate `Ed25519`
entry is `-19` and is added under #14936. Leaving the old name would put two different things called
`Ed25519` in the same codebase.

Second half of the issue: `fromNumericValue` returns `null` for any code point with no constant, and call
sites fed that straight into a `switch`, so an unrecognised algorithm produced an unhandled
`NullPointerException` before any input was read. Guarded at the three sites that dereference it —
[...]

PR #14958, opened 2026-09-03
fix(jans-fido2): support P-384 and P-521 curves in COSE key decoding

imran-ishaq:
### Prepare

- [x] Read [PR guidelines](https://github.com/JanssenProject/jans/blob/main/docs/CONTRIBUTING.md#prs)
- [x] Read [license information](https://github.com/JanssenProject/jans/blob/main/LICENSE)

-------------------

#### Target issue

closes #14931

Sub-issue of #14925, under #14891. Follows #14950 (#14929), which landed the OKP branch in the same method.

#### Implementation Details

`CoseService.convertCoseCurveToSunCurveName()` mapped exactly one curve:

```java
if (curve == COSE_CURVE_P256) {
    return SECP256R1_CURVE_NAME;
}
throw new Fido2RuntimeException("Unsupported curve");
```

So an authenticator on any curve other than P-256 failed registration regardless of which algorithm it
negotiated — confirmed at runtime with an EC2 key on P-384, which produced
`Fido2RuntimeException: Unsupported curve`. P-384 is **Recommended** in the FIDO Server Requirements v2.3
table and P-521 is Optional.

This maps curve `2` to `secp384r1` and curve `3` to `secp521r1`, and makes the `default` branch report the
curve identifier it actually received rather than a bare "Unsupported curve" — the old message gave an
operator nothing to act on.

No change was needed in `convertUncompressedPointToECKey`: it already derives the field size from the curve
parameters, so the coordinate handling works for all three curves once the name resolves.

Note this is the curve mapping only. The EC2 branch still accepts just `ES256`, so a P-384 credential
negotiating `ES384` remains unsupported — that is #14932, and the wrong `ES384` / `ES512` code points are
#14924.

`server-fips/` needs no mirroring: it is a pom-only module that repackages the `server` artifact and has no
`src/` tree of its own. All three curve names resolve through `AlgorithmParameters.getInstance("EC")`, which
is JDK-provided rather than provider-specific.

-------------------
### Test and Document the changes
- [x] Static code analysis has been run locally and issues have been fixed
- [x] Relevant unit and integration tests have been added/updated
- [ ] Relevant documentation has been updated if any (i.e. user guides, installation and configuration guides, technical design docs etc)

New `CoseServiceCurveTest` — round-trip decode for a generated keypair on each of P-256, P-384 and P-521,
asserting the rebuilt key matches the original encoding, plus a case asserting an unknown curve identifier
appears in the error message.

```
[...]

PR #15031, opened 2026-09-09
refactor(jans-fido2): derive advertised algorithms from decoder and verifier capability

### LINKED ISSUE #14933: refactor(jans-fido2): derive advertised algorithms from decoder and verifier capability

The set of algorithms we advertise, the set we can decode keys for, and the set we can verify signatures with
are three independently maintained lists that do not agree:

- advertised: hardcoded in `AttestationService.preparePublicKeyCredentialSelection():504-508`, plus whatever
  `enabledFidoAlgorithms` resolves to via `resolveRsaNumericValue` / `resolveEc2NumericValue` /
  `resolveEdDsaNumericValue` (`:542-560`)
- decodable: the `switch` in `CoseService.createUncompressedPointFromCOSEPublicKey()`
- verifiable: the `switch` in `SignatureVerifier.getSignatureChecker()`

Every defect in this tree is an instance of that disagreement. EdDSA is advertised and verifiable but not
decodable (B1). `ES384` is verifiable but neither correctly advertised (A) nor decodable (B3). An admin can
configure an algorithm that resolves to a code point no part of the pipeline can complete.

Fixing B1–B3 closes today's gaps but leaves the structure that produced them: the next algorithm someone adds
to one list and not the others reintroduces the same class of bug, silently, and it will not surface until an
authenticator in the field picks that algorithm.

Make an algorithm advertisable **only if** both the decoder and the verifier can handle it — enforced by
construction rather than by convention. Concretely: a single registry of supported algorithms that both
`CoseService` and `SignatureVerifier` are driven from (or validated against at startup), with
`preparePublicKeyCredentialSelection()` reading from it instead of a literal list.

An admin-configured algorithm that the registry does not support should fail loudly at configuration time,
not silently produce a registration failure later.

- **A unit test asserting the three lists agree.** Much cheaper and catches the drift at CI time rather than
  preventing it. A reasonable substitute if the refactor is judged too invasive — most of the value for a
  fraction of the work.
- **Leave the lists separate and document the invariant.** Rejected: that is effectively the current state.

**Additional context**
Sub-issue of #14925. Wants #14929–14932landed first, otherwise it encodes today's narrow capability as the contract. This is the item that stops the audit's findings recurring — worth keeping even if it is scheduled well after the correctness fixes.

imran-ishaq:
### Prepare

-------------------

#### Target issue

closes #14933

Sub-issue of #14925, under #14891. Follows #14950 (#14929), #14958 (#14931), #14970 (#14924) and
#14975 (#14932), all merged. This is the item that stops the rest of the tree's findings recurring.

#### Implementation Details

Three lists decided what this server does with an algorithm, and they were maintained independently:

- **advertised** — a literal list in `AttestationService.preparePublicKeyCredentialSelection()`, plus
  whatever `enabledFidoAlgorithms` resolved to
- **decodable** — `CoseService.createUncompressedPointFromCOSEPublicKey()`
- **verifiable** — `SignatureVerifier.getSignatureChecker()`

Every defect in this tree was an instance of those disagreeing. #14975 tied the last two together with a
drift guard; the advertised list was still free to name anything. An admin could configure an algorithm no
part of the pipeline could complete and get a registration failure in the field rather than an error in the
log.

An algorithm is now advertisable only when this server can complete a registration with it end to end:

```java
private boolean isFullySupported(int algorithm) {
    return coseService.isDecodable(algorithm) && signatureVerifier.isSupported(algorithm);
}
```

Both the unconfigured defaults and each configured name pass through it. A recognised name that fails the
check is logged at `ERROR` and left out of `pubKeyCredParams` instead of being advertised.

No new registry class was needed. `CoseService` and `SignatureVerifier` already *were* the two capability
sources — they just answer the question now:

- `CoseService.isDecodable(int)` reads the same `DECODABLE_RSA_ALGORITHMS` / `DECODABLE_EC2_ALGORITHMS`
  sets the decoder branches on, so there is no second list to drift from.
- `SignatureVerifier.isSupported(int)` calls `getSignatureChecker` and reports whether it succeeded. It
  asks the provider rather than a list, which is the part that matters below.

**Why this is the unblocker for #14937, not a BouncyCastle upgrade.** The #14935 spike found that
`bcprov-jdk18on 1.79` fully supports ML-DSA-44/65/87 while no released `bc-fips` line does — not our pin,
not 2.0.0, not 2.1.1. `SecurityProviderUtility.getBCProvider()` returns BCFIPS whenever that provider is on
the classpath, so shipping ML-DSA against a hardcoded advertised list would make a FIPS deployment
advertise it and then fail at `Signature.getInstance` — exactly the advertise-then-fail shape as the EdDSA
defect in #14929. Because `isSupported` asks the running provider, the FIPS build now reports its own,
narrower capability and simply advertises less. That is the precondition #14935 identified.

**Empty result.** If no configured algorithm survives the check, the server logs an error and falls back to
the defaults it does support. An empty `pubKeyCredParams` is not a safe failure — the client falls back to
its own defaults — so silently honouring an unusable configuration would be worse than loudly ignoring it.

**Behaviour change.** A deployment that today configures an algorithm this server cannot complete gets it
advertised; after this it does not, and the reason appears at `ERROR` on each options call. No supported
configuration changes.

antrun step copies `${server.webapp.dir}` from the `server` module.

-------------------

`AttestationServiceAlgorithmSelectionTest` covers the derivation: the unconfigured defaults, a default the
provider cannot verify being dropped, configured names being honoured, an unsupported name being skipped so
the next in its family is used, the empty-result fallback, and an unknown name being ignored.
`preparePublicKeyCredentialSelection()` was made package-private so this can be tested without standing up
`options()`.

`CoseServiceDecoderParityTest` now also asserts that `isDecodable && isSupported` agrees with what the
decoder and verifier actually just did, for every algorithm it walks. A predicate that drifted from the
behaviour would put an algorithm back into `pubKeyCredParams` that the ceremony then fails on, and the
existing guard would not have caught it.

I checked that assertion is not vacuous — inverting the return of `SignatureVerifier.isSupported` makes it
fail on both ES256 and RS256, and I reverted after confirming it.

```
AttestationServiceAlgorithmSelectionTest    6 tests, 0 failures
CoseServiceDecoderParityTest                4 tests, 0 failures
jans-fido2 model + server                 504 tests, 0 failures
```

Docs: `docs/janssen-server/fido/fido2-server-properties-config.md` gains an "Advertised algorithms" section
covering the derivation, what it means for the FIPS build, and the empty-result fallback; the
`enabledFidoAlgorithms` row now records that an unusable name is logged and dropped. Carried in a separate
`docs:` commit. The section is placed after the configuration table rather than inside it, following the
review feedback on #14974.

## Summary by CodeRabbit

* **New Features**
  * FIDO2 registration now advertises only credential algorithms fully supported by the server, including key decoding and signature verification.
  * Unsupported configured algorithms are excluded automatically, with supported alternatives selected when available.
  * If no configured algorithms are usable, supported defaults are advertised.

* **Bug Fixes**
  * Corrected RSA-PSS signature handling for SHA-384 and SHA-512 algorithms.

* **Documentation**
  * Added guidance on algorithm advertisement, FIPS differences, fallback behavior, and related logging.

### DIFF
--- a/docs/janssen-server/fido/fido2-server-properties-config.md
+++ b/docs/janssen-server/fido/fido2-server-properties-config.md
@@ -50,7 +50,7 @@ This nested block defines WebAuthn and FIDO2 attestation and assertion policy be
 | `unfinishedRequestExpiration` | Integer | `120` | Expiration time in seconds for incomplete registration/authentication requests. |
 | `metadataRefreshInterval` | Integer | `1296000` | Expiration time in seconds (e.g., 15 days) before checking and reloading the FIDO Alliance MDS TOC. |
 | <span id="servermetadatafolder">`serverMetadataFolder`</span> | String | `"/etc/jans/conf/fido2/server_metadata"` | Folder where local vendor metadata statement JSON files are placed manually. |
-| `enabledFidoAlgorithms` | Array of Strings | `["RS256", "ES256"]` | Enabled cryptographic signing algorithms allowed for credentials. Accepted names: `RS256`, `RS384`, `RS512`, `RS65535`, `PS256`, `PS384`, `PS512`, `ES256`, `ES384`, `ES512`, `EdDSA` — the algorithms the server can both advertise and complete a registration with. When unset, the server advertises `RS256`, `ES256` and `EdDSA`. An unrecognised name is ignored. |
+| `enabledFidoAlgorithms` | Array of Strings | `["RS256", "ES256"]` | Enabled cryptographic signing algorithms allowed for credentials. Accepted names: `RS256`, `RS384`, `RS512`, `RS65535`, `PS256`, `PS384`, `PS512`, `ES256`, `ES384`, `ES512`, `EdDSA` — the algorithms the server can both advertise and complete a registration with. When unset, the server advertises `RS256`, `ES256` and `EdDSA`. An unrecognised name is ignored. A recognised name the deployment cannot actually complete a registration with is logged at `ERROR` and left out of `pubKeyCredParams` — see [Advertised algorithms](#advertised-algorithms). |
 | `rp` | Array of Objects | `[ { "id": "https://jans.io", "origins": ["jans.io"] } ]` | Relying Party (RP) configuration mapping expected IDs to valid origins. |
 | `metadataServers` | Array of Objects | `[ { "url": "https://mds.fidoalliance.org/" } ]` | External FIDO Metadata Service endpoints to download statement catalogs. |
 | `disableMetadataService` | Boolean | `false` | If set to `true`, the FIDO2 server skips validating authenticators against the MDS3 service. |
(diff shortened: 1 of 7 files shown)


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/36fd62c5-37ed-427c-96ca-17bb814dbb5c?refs=1531d942c8646789fec3ffeac9e2382104f9689a&source=prChecksComment) |

yurem (approved):

PR #15032, opened 2026-09-09
feat(jans-fido2): add ESP256 and ESP384 signature algorithms

imran-ishaq:
### Prepare

- [x] Read [PR guidelines](https://github.com/JanssenProject/jans/blob/main/docs/CONTRIBUTING.md#prs)
- [x] Read [license information](https://github.com/JanssenProject/jans/blob/main/LICENSE)

-------------------

#### Target issue

closes #14934

Sub-issue of #14926, under #14891. Depends on #14958 (#14931), which is merged — `ESP384` signs over
P-384, and the decoder rejected that curve before it landed.

#### Implementation Details

`ESP256` and `ESP384` are Recommended in the FIDO Server Requirements v2.3 algorithm table and we supported
neither. `SignatureVerifier.getSignatureChecker()` throws `"Unknown mapping"` for any code point outside its
switch, so an authenticator offering either could not register.

These are the *fully-specified* ECDSA algorithms: the code point fixes the curve rather than the key
carrying it, so the cryptography is identical to `ES256` and `ES384`, which are already implemented. The
work is the new COSE code points and the plumbing to reach existing crypto — no new cryptography.

- `CoseEC2Algorithm` gains `ESP256(-9)` and `ESP384(-51)`, per the
  [IANA COSE Algorithms registry](https://www.iana.org/assignments/cose/cose.xhtml#algorithms).
- `SignatureVerifier.getSignatureChecker()` maps `-9` to the existing `SHA256withECDSA` instance and `-51`
  to `SHA384withECDSA`.
- `CoseService.DECODABLE_EC2_ALGORITHMS` gains both. No new decode path was needed — every EC2 key
  reconstructs from the same `(crv, x, y)` triple, and `ESP384`'s P-384 curve already resolves since
  #14958.

`getDigest()` needs no entry: that switch serves only the legacy U2F, `-8` and `-257` paths, and `ES256` and
`ES384` are absent from it too.

They are configurable via `enabledFidoAlgorithms` immediately, since `resolveEc2NumericValue` resolves
through `CoseEC2Algorithm.valueOf`. Once #15031 (#14933) merges they also become advertisable with no
further change here — that PR derives the advertised set from whatever the decoder and verifier both
handle, which now includes these.

`ESP512 (-52)` is deliberately left out. It is Optional in v2.3 where these two are Recommended, and this
issue names only the Recommended pair. It is three lines and the same pattern whenever it is wanted.

[...]

PR #14950, opened 2026-09-02
fix(jans-fido2): decode OKP public keys so advertised EdDSA credentials can register

imran-ishaq:
Prepare

- [x] Read PR guidelines (https://github.com/JanssenProject/jans/blob/main/docs/CONTRIBUTING.md#prs)
- [x] Read license information (https://github.com/JanssenProject/jans/blob/main/LICENSE)

---

Description

Target issue

closes #14929

Sub-issue of #14925, under #14891.

Implementation Details

AttestationService.preparePublicKeyCredentialSelection() advertises CoseEdDSAAlgorithm.Ed25519(-8) in pubKeyCredParams whenever enabledFidoAlgorithms is unset — the default configuration. An Ed25519 credential returns a COSE key with kty = OKP, and CoseService.createUncompressedPointFromCOSEPublicKey() was literally case OKP: { throw new Fido2RuntimeException(...); }. EdDSA and curve Ed25519 are both Required in FIDO Server Requirements v2.3, so an authenticator picking EdDSA off our own advertised list could not register.

SignatureVerifier.getSignatureChecker():72 handles -8 correctly, which is why this stayed quiet — the request failed in the decoder and never reached the verifier.

This adds the OKP branch:

- Resolves the algorithm via CoseEdDSAAlgorithm.fromNumericValue and raises a handled Fido2RuntimeException when it is unrecognised, rather than passing null into the switch and throwing an unhandled NPE.
- Reads the COSE crv (-1) and x (-2) parameters and rebuilds the key by wrapping the raw 32 bytes in an RFC 8410 SubjectPublicKeyInfo (1.3.101.112), then loading it through KeyFactory.
- Uses SecurityProviderUtility.getBCProvider() and the algorithm name Ed25519, matching what SignatureVerifier already does.

Every attestation processor (PackedAttestationProcessor, AppleAttestationProcessor, AndroidKeyAttestationProcessor, U2FAttestationProcessor) and AssertionVerifier route through this one method, so no other call site needed changing.

Curves other than Ed25519 and the remaining OKP algorithms are deliberately out of scope — Ed448 is tracked in #14936, and the algorithm-name rename (Ed25519(-8) → EdDSA) in #14924.

server-fips/ parity was verified rather than assumed. The module has no sources of its own — it repackages the server webapp and swaps bcprov-jdk18on for bc-fips — so the question is provider capability. Checked directly against bc-fips 1.0.2.4:

provider=BCFIPS v1.000204
[...]

Issue #14935, opened 2026-09-01
chore(jans-fido2): assess BouncyCastle and FIPS support for ML-DSA post-quantum algorithms


imran-ishaq:
**Is your feature request related to a problem? Please describe.**

The three post-quantum algorithms `ML-DSA-44`, `ML-DSA-65` and `ML-DSA-87` are Recommended in v2.3, and we
cannot currently size the work to add them because we do not know whether our crypto providers support them —
in either build variant.

- Standard build: `bc.version = 1.79` (`jans-bom/pom.xml:69`, `bcprov-jdk18on`)
- FIPS build: `bc-fips (,1.0.2.4]` and `bcpkix-fips 1.0.8` (`jans-bom/pom.xml:315-323`) — a considerably older
  line, and the likely blocker for `server-fips/` parity

Committing to an implementation before answering this risks discovering mid-change that the FIPS variant
cannot follow, which would either break `server-fips/` parity or strand the work.

**Describe the solution you'd like**

A written assessment, posted on this issue, answering:

1. Does `bcprov-jdk18on 1.79` support ML-DSA key parsing and signature verification? If not, which version does,
   and what does upgrading `bc.version` in `jans-bom` affect across the Java modules?
2. Does any released `bc-fips` line support ML-DSA? If not, is there a roadmap, and what is the FIPS
   certification status of the implementation when it lands?
3. What is the resulting recommendation — implement now, implement standard-only with an explicit parity
   exception, or defer with a documented reason?

The deliverable is the writeup, not code. "Defer, here is the evidence" is a perfectly good outcome and closes
this issue successfully.

**Describe alternatives you've considered**

- **Skip the review and start implementing.** Rejected — that is precisely the risk described above.
- **Ask BouncyCastle upstream rather than assessing locally.** Complementary, not an alternative; worth doing
  if the local answer is unclear.

**Additional context**
Sub-issue of #14926. Independent of everything in #14924 and #14925, so it can run in parallel with the correctness work and is a good first thing to pick up.

imran-ishaq:
ML-DSA BouncyCastle / FIPS capability review

This closes out the assessment asked for in this issue. Measured by loading each provider jar directly from the local Maven repository and instantiating the provider class by reflection, so the results reflect the artifacts the build actually resolves rather than a classpath-dependent guess. Run twice, a week apart, with identical results.

Question 1 — does bcprov-jdk18on 1.79 support ML-DSA?

[...]

Issue #14929, opened 2026-09-01
fix(jans-fido2): decode OKP public keys so advertised EdDSA credentials can register


imran-ishaq:
**Is your feature request related to a problem? Please describe.**

`AttestationService.preparePublicKeyCredentialSelection()`
(`server/src/main/java/io/jans/fido2/service/operation/AttestationService.java:504-508`) advertises
`CoseEdDSAAlgorithm.Ed25519(-8)` in `pubKeyCredParams` whenever `enabledFidoAlgorithms` is unset — the default
configuration. An Ed25519 credential comes back as a COSE key with `kty = OKP (1)`, and the decoder cannot
handle it (`server/src/main/java/io/jans/fido2/service/CoseService.java:123-125`):

```java
case OKP: {
    throw new Fido2RuntimeException("Don't know what to do with this key" + keyType);
}
```

Confirmed at runtime — a synthetic OKP COSE key (`kty=1, alg=-8, crv=6, x=32 bytes`) passed to
`createUncompressedPointFromCOSEPublicKey`:

```
io.jans.fido2.exception.Fido2RuntimeException: Don't know what to do with this keyOKP
```

`grep -rn "OKP" --include=*.java` across jans-fido2 returns three hits total — two in `CoseKeyType`, plus this
throw — so nothing filters OKP out before it reaches here. Every attestation processor
(`PackedAttestationProcessor:140`, `AppleAttestationProcessor:178`, `AndroidKeyAttestationProcessor:139`,
`U2FAttestationProcessor:217`) and `AssertionVerifier:112` route through this one method. There is no
alternative path.

`SignatureVerifier.getSignatureChecker():72` handles `-8` correctly via `Signature.getInstance("Ed25519")`,
which is why this has gone unnoticed: the verifier is right, the request never reaches it.

**EdDSA and curve Ed25519 are both Required in the v2.3 table.** An authenticator that selects EdDSA from our
own advertised list cannot register.

**Describe the solution you'd like**

Add an OKP branch to `createUncompressedPointFromCOSEPublicKey` that reconstructs an Ed25519 public key from
the COSE `crv` (`-1`) and `x` (`-2`) parameters, and returns a `PublicKey` the existing
`SignatureVerifier` path accepts unchanged. Cover it with a unit test that registers and verifies an Ed25519
credential end to end.

**Describe alternatives you've considered**

- **Drop EdDSA from the default advertised set.** One line, removes the failure immediately. Rejected as the
[...]

PR #15034, opened 2026-09-09
feat(jans-fido2): add Ed25519 and Ed448 signature algorithms

### LINKED ISSUE #14936: feat(jans-fido2): add Ed25519 and Ed448 signature algorithms

The v2.3 table lists `Ed25519 (-19)` and `Ed448 (-53)` as Recommended algorithms, at code points distinct
from the existing `EdDSA (-8)`. We support neither.

There is a naming trap here worth stating explicitly: our `CoseEdDSAAlgorithm.Ed25519` constant holds `-8`,
which is **EdDSA**, not the table's separate `Ed25519` entry. Adding the real one without renaming the
existing constant first would leave two different things called `Ed25519` in the same codebase.

Rename the existing `CoseEdDSAAlgorithm.Ed25519(-8)` to `EdDSA` (this is part of sub-issue A), then add
`Ed25519 (-19)` and `Ed448 (-53)`. `Ed25519` maps to the same `Signature.getInstance("Ed25519")`
we already instantiate for `-8`; `Ed448` needs a provider capability check. Both are OKP key types, so both
need the decoder support from B1.

- **`Ed25519` only, defer `Ed448`.** Reasonable if the provider check for Ed448 turns up a gap — they are
  separable, and Ed25519 is the one with real-world authenticator support.
- **Add the new constants without the rename.** Rejected: two constants named `Ed25519` holding different
  values is a defect waiting to happen.

**Additional context**
Sub-issue of #14926. Depends on #14929 (OKP decoding) and on the rename in #14924. Code points per IANA: Ed25519 = -19, Ed448 = -53, versus EdDSA = -8 for the existing constant.

imran-ishaq:
### Prepare

-------------------

#### Target issue

closes #14936

Sub-issue of #14926, under #14891. Depends on #14950 (#14929) for OKP decoding and on the `Ed25519` →
`EdDSA` rename in #14970 (#14924), both merged.

#### Implementation Details

The v2.3 table lists `Ed25519 (-19)` and `Ed448 (-53)` as Recommended, at code points distinct from the
existing `EdDSA (-8)`. We supported neither, so an authenticator offering either could not register —
`SignatureVerifier.getSignatureChecker()` throws `"Unknown mapping"` outside its switch, and `CoseService`
rejected any OKP curve but Ed25519.

**Provider support, measured rather than assumed.** #14936 flags that `Ed448` needs a capability check, so I
loaded each provider jar directly and exercised keygen, `KeyFactory`, sign and verify:

```
bcprov-jdk18on 1.79    Ed25519 OK (spki 44, sig 64)   Ed448 OK (spki 69, sig 114)
bc-fips 1.0.2.4        Ed25519 OK (spki 44, sig 64)   Ed448 OK (spki 69, sig 114)
```

Both build variants support both curves at the versions already pinned in `jans-bom`. So unlike the ML-DSA
answer in #14935, there is no dependency work, no FIPS gap, and no reason to split `Ed448` out — the
issue's "Ed25519 only, defer Ed448" alternative does not apply.

The change itself:

- `CoseEdDSAAlgorithm` gains `Ed25519(-19)` and `Ed448(-53)` per the
  [IANA COSE Algorithms registry](https://www.iana.org/assignments/cose/cose.xhtml#algorithms). The naming
  trap the issue warns about is already gone: #14970 renamed the old `Ed25519(-8)` constant to `EdDSA`, so
  there is no point at which two different things are called `Ed25519`.
- `SignatureVerifier.getSignatureChecker()` maps `-19` to the `Ed25519` instance it already builds for `-8`,
  and `-53` to `Ed448`.
- `CoseService.convertRawKeyToEdDSAKey()` handles COSE curve 7 alongside 6, selecting the algorithm name,
  the RFC 8410 SubjectPublicKeyInfo prefix (OID `1.3.101.113`) and the raw key length (57 bytes rather than
  32). I read the Ed448 DER prefix off a generated key rather than hand-computing it.

**Curve pinning.** `Ed25519` and `Ed448` are fully specified: the code point names the curve, so the curve
carried in the key is not free to disagree. A `REQUIRED_OKP_CURVES` map rejects a mismatched pair before key
construction. This is the same class of gap review found on #15032 for `ESP256`/`ESP384`, applied here up
front. It has to live in the decoder — `SignatureVerifier` resolves these to a plain `Signature` instance
that never sees the COSE curve, so the decoder is the only place the pairing is visible.

`EdDSA (-8)` is deliberately exempt: it is not fully specified, it legitimately carries either Edwards
curve, and constraining it would invalidate every credential already registered under it. A test pins that
exemption.

**Behaviour change.** `EdDSA (-8)` credentials are unaffected in every respect, and gain Ed448 keys as a
side effect of the widened decoder. Nothing that worked before stops working.

antrun step copies `${server.webapp.dir}` from the `server` module. The FIPS provider probe above is what
makes that safe here rather than merely structural.

-------------------

`CoseServiceFullySpecifiedEdDSATest` covers the registration and assertion pair for each algorithm: build
the COSE key an authenticator would send, decode it, assert it reconstructs byte-for-byte, then verify a
real signature made with the matching private key. It also covers the wrong-curve rejection in both
directions, `EdDSA (-8)` still accepting either curve, and a truncated key for each length.

One existing test needed repointing rather than extending.
`CoseServiceEdDSATest.createUncompressedPointFromCOSEPublicKey_withEd448Curve_throwsUnsupportedCurve`
asserted that Ed448 was unsupported — a true statement when #14950 landed and a false one now. It is
renamed to `..._withNonSignatureCurve_...` and uses COSE curve 4 (X25519, key agreement rather than
signing), so the unsupported-curve path is still covered by a curve that is genuinely unsupported.

```
CoseServiceFullySpecifiedEdDSATest    8 tests, 0 failures
CoseServiceEdDSATest                  7 tests, 0 failures
jans-fido2 model + server           506 tests, 0 failures
```

Docs: `docs/janssen-server/fido/fido2-server-properties-config.md` adds `Ed25519` and `Ed448` to the
accepted `enabledFidoAlgorithms` names and gains a "Fully-specified EdDSA algorithms" section explaining
the distinction from `EdDSA` and recording that existing credentials are unaffected. Carried in a separate
`docs:` commit.

## Summary by CodeRabbit

* **New Features**
  * Added support for Ed25519 and Ed448 FIDO2/COSE algorithms.
  * Added signature verification for fully specified Ed25519 and Ed448 algorithms.
  * Added validation to reject mismatched curves and invalid key lengths.
  * Continued support for generic EdDSA credentials using either supported curve.

* **Documentation**
  * Updated configuration documentation with accepted algorithm names, curve behavior, and registration validation details.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/20a0d06d-86ba-41e7-b997-bc918d5afcb0?refs=1ad5adec94976f69f79f9816500d2f0ac2fc5347&source=prChecksComment) |

yurem (dismissed):

yurem (dismissed):

yurem (approved):

Issue #14939, opened 2026-09-01
fix(jans-fido2): read and enforce the crossOrigin member of client data


imran-ishaq:
**Is your feature request related to a problem? Please describe.**
WebAuthn L3 requires the Relying Party to inspect the crossOrigin member of CollectedClientData during both registration and authentication verification. We never read it.

CommonVerifiers.verifyClientJSON() (server/src/main/java/io/jans/fido2/service/verifier/CommonVerifiers.java:314-361) requires challenge, origin and type to be present, validates the challenge as base64url, optionally validates tokenBinding, and checks origin is a non-empty string. That is the whole of it:

long count = Arrays.asList(clientJsonNode.hasNonNull(CHALLENGE), clientJsonNode.hasNonNull("origin"),
        clientJsonNode.hasNonNull("type")).parallelStream().filter(f -> !f).count();

grep -rn "crossOrigin\|topOrigin" --include=*.java   # across the entire monorepo
-> zero hits

A ceremony performed inside a cross-origin iframe is therefore accepted exactly as if it had been performed at the top level. The origin check we do run compares against the RP's configured origins, so this is not an open door — but it is precisely the check L3 adds for the framed case, and we are silently skipping it.

**Describe the solution you'd like**
- Read crossOrigin in verifyClientJSON(). When the member is absent, treat it as false, per spec.
- Reject a ceremony with crossOrigin = true by default, with a distinct, logged error rather than a generic invalid-request.
- Unit-test all three cases over the client-data verifier: absent, false, and true.
- Verify server-fips/ parity.

**Describe alternatives you've considered**
- Log the value without acting on it. Half a fix — a rejected ceremony is the point of the check.
- Do nothing until the cross-origin iframe feature is scheduled. Rejected: this is a correctness gap against the current spec on its own, whether or not that feature ever ships.
- Accept crossOrigin = true and validate only topOrigin. That is #14940, and it needs this member read first; rejecting by default is the conservative starting posture.

**Additional context**
Sub-issue of #14938. Surfaced by the WebAuthn L3 audit that produced #14891, but unrelated to the algorithm table. Independent of everything in the v2.3 tree — can be picked up at any time.

Reference: WebAuthn L3 (https://www.w3.org/TR/2026/REC-webauthn-3-20260825/)

Issue #14938, opened 2026-09-01
fix(jans-fido2): validate crossOrigin and topOrigin in client data as WebAuthn L3 requires


imran-ishaq:
**Is your feature request related to a problem? Please describe.**
WebAuthn L3 requires the Relying Party to inspect the crossOrigin member of CollectedClientData during both registration and authentication verification, and to take topOrigin into account when it is true. We do neither.

CommonVerifiers.verifyClientJSON() (server/src/main/java/io/jans/fido2/service/verifier/CommonVerifiers.java:314-361) requires exactly challenge, origin and type to be present (:333), validates the challenge as base64url, optionally validates tokenBinding, and checks that origin is a non-empty string (:355). crossOrigin and topOrigin are never read.

grep -rn "crossOrigin\|topOrigin" --include=*.java   # across the entire monorepo
-> zero hits

So a ceremony performed inside a cross-origin iframe is accepted exactly as if it had been performed at the top level, and the RP has no record of which origin actually framed it. The origin check we do run compares against the RP's configured origins, so this is not an open door — but it is the specific check L3 adds to make the framed case safe, and we are silently not doing it.

**Describe the solution you'd like**
- Read crossOrigin from the client data in verifyClientJSON(). When absent, treat it as false per spec.
- When crossOrigin is true, validate topOrigin against the RP's configured origin policy rather than accepting the ceremony unconditionally, and make the outcome explicit in the logs.
- Decide and document the default posture for cross-origin ceremonies — reject unless configured otherwise is the conservative reading — and expose it as configuration rather than hardcoding it.
- Cover both branches with unit tests over the client-data verifier, and update docs/janssen-server/fido/ for whatever configuration property lands.
- Verify server-fips/ parity.

**Describe alternatives you've considered**
- Read crossOrigin and reject any true outright, ignoring topOrigin. Simplest and safe, but forecloses the cross-origin iframe passkey work tracked separately, which needs the topOrigin path to exist.
- Do nothing until the iframe feature is scheduled. Rejected: the validation gap is a correctness issue against the current spec on its own, independent of whether we ever ship the iframe feature.
- Log the value without acting on it. Half a fix; a rejected ceremony is the point.

**Additional context**
Surfaced by the same WebAuthn L3 audit that produced #14891, but unrelated to the algorithm table — filed separately on purpose, since this is client-data verification in CommonVerifiers and none of #14891's acceptance criteria touch it.

[...]

PR #15037, opened 2026-09-09
feat(jans-fido2): validate topOrigin against a configurable cross-origin policy

### LINKED ISSUE #14940: feat(jans-fido2): validate topOrigin against a configurable cross-origin policy
Once crossOrigin is read (sub-issue D1), rejecting every framed ceremony outright is the safe default but not the end state: topOrigin is what makes the cross-origin case safe rather than merely blocked. WebAuthn L3 expects the RP to take topOrigin into account when crossOrigin is true, and we have no mechanism to express which framing origins an RP will accept.

Without it, deployments that legitimately need a framed passkey ceremony have no option but to accept all cross-origin ceremonies or none.

- When crossOrigin is true, validate topOrigin against the RP's configured origin policy instead of rejecting unconditionally.
- Add a configuration property for the permitted framing origins, defaulting to empty — deny — so that upgrading changes no behaviour. Reject with an explicit, logged error when topOrigin is missing or unlisted.
- Document the property under docs/janssen-server/fido/ alongside the existing FIDO2 properties.
- Test the accept and reject paths, plus crossOrigin = true with topOrigin absent.
- Verify server-fips/ parity.

- Reuse the existing origin allow-list for topOrigin too. Simpler, but conflates two different questions — which origin served the ceremony versus which origin framed it — and would silently widen the framing policy for every existing deployment.
- Hardcode the policy rather than exposing configuration. Rejected: the right answer is deployment-specific.
- Skip this and leave D1's blanket rejection as the permanent behaviour. Defensible until someone needs a framed ceremony; this issue is the thing to pick up when they do.

**Additional context**
Sub-issue of #14938. Depends on #14939. This is also the validation prerequisite for the separately tracked cross-origin iframe create() work — it is not that feature, but that feature cannot be done safely without it.

imran-ishaq:
### Prepare

-------------------

#### Target issue

closes #14940

Sub-issue of #14938. Depends on #14939, delivered by #14974 and merged — this is the follow-up that PR
explicitly deferred to.

#### Implementation Details

#14974 made the server read `crossOrigin` and reject every framed ceremony. That is the safe default but not
the end state: `topOrigin` is what makes the cross-origin case *safe* rather than merely *blocked*. WebAuthn
Level 3 expects the RP to take `topOrigin` into account once `crossOrigin` is `true`, and until now a
deployment that legitimately needs a framed passkey ceremony had no option at all.

`CommonVerifiers.verifyCrossOrigin` now delegates to `verifyTopOrigin` when `crossOrigin` is `true` instead
of rejecting unconditionally. The ceremony is rejected with `cross_origin_not_allowed`, each case carrying
its own log line, when:

- `allowedTopOrigins` is empty — the default
- `topOrigin` is absent, `null`, not a string, or blank
- `topOrigin` is not listed

Everything else about the member is unchanged: an absent `crossOrigin` still means `false`, and a
non-boolean or explicitly `null` `crossOrigin` still fails as `invalid_request`. One guard continues to cover
both ceremonies, since `verifyClientJSON` serves attestation and assertion alike.

**No behaviour change on upgrade.** `allowedTopOrigins` defaults to empty, and empty means deny, so a
deployment that upgrades without configuring anything keeps exactly the blanket rejection #14974 introduced.
Framed ceremonies become possible only once an administrator opts in by naming the framing origins.

**A separate list, not the `rp` origins.** This is the alternative #14940 raises and rejects, and the
rejection is right. The `origins` under `rp` say which origin may *serve* a ceremony; `allowedTopOrigins`
says which may *frame* one. Reusing the former would silently widen the framing policy of every existing
deployment the moment this merged — precisely the kind of quiet widening the cross-origin work exists to
prevent.

**A flat property rather than per-`RequestedParty`.** Worth a reviewer's opinion. #14940's wording ("the
RP's configured origin policy") can be read either way. `verifyClientJSON` parses client data and has no RP
context, so scoping the list per RP would mean threading the resolved RP through both ceremonies — a much
larger change than the policy itself. If per-RP scoping is wanted, the structure to hang it on already
exists and it is a clean follow-up.

**Matching.** Comparison is against the whole origin, ignoring case and surrounding whitespace. It is
deliberately *not* host-only like `verifyRpDomain`: a different scheme or port is a different origin, so
`https://portal.example.com` must not be satisfied by `http://portal.example.com`.

antrun step copies `${server.webapp.dir}` from the `server` module.

-------------------

Six cases added to `CommonVerifiersTest`: an allowed `topOrigin`; one differing only in case and surrounding
whitespace; one not on the list; one differing only in scheme; `topOrigin` absent while the list is
configured; and `topOrigin` present but not a string.

One existing test is renamed rather than changed. `verifyClientJSON_whenCrossOriginTrue_rejected` asserted
that `crossOrigin = true` is rejected, which is now true only because no origins are configured — it becomes
`verifyClientJSON_whenCrossOriginTrueAndNoTopOriginsConfigured_rejected`, same assertion, accurate name, and
it is what pins the no-behaviour-change-on-upgrade claim above.

I checked the allow-list comparison is load-bearing rather than incidentally satisfied — short-circuiting it
to always allow fails exactly the unlisted and scheme-mismatch cases and nothing else, and I reverted after
confirming.

```
CommonVerifiersTest         74 tests, 0 failures
jans-fido2 model + server  504 tests, 0 failures
```

`FetchMdsProviderServiceTest` errors on my machine because it makes live calls to `mds3.fido.tools` and local
DNS is down; it is excluded from that count and is unrelated to this change.

Docs: `docs/janssen-server/fido/fido2-server-properties-config.md` gains an `allowedTopOrigins` row, and the
"Cross-origin ceremonies" section is rewritten — it previously stated that no such configuration existed,
which this change makes false. The new text lists each rejection case, the matching rules, and why the
property is separate from `rp.origins`. Carried in a separate `docs:` commit.

`jans-linux-setup/jans_setup/templates/jans-fido2/dynamic-conf.json` is deliberately untouched: it is already
a partial mirror of `Fido2Configuration` (around ten of its properties), and the empty default is the one a
fresh install should have. Happy to add it explicitly if reviewers prefer the installer to surface the knob.

## Summary by CodeRabbit

* **New Features**
  * Cross-origin FIDO2 ceremonies can now be permitted using a configurable list of allowed top-level origins.
  * Origins are validated against the configured allow-list, including scheme and formatting requirements.

* **Bug Fixes**
  * Cross-origin ceremonies now return a clear rejection when the top origin is missing, invalid, or not allow-listed.

* **Documentation**
  * Updated FIDO2 configuration guidance to describe allowed top origins and distinguish them from relying-party origins.

### DIFF
--- a/docs/janssen-server/fido/fido2-server-properties-config.md
+++ b/docs/janssen-server/fido/fido2-server-properties-config.md
@@ -59,15 +59,25 @@ This nested block defines WebAuthn and FIDO2 attestation and assertion policy be
 | `hints` | Array of Strings | `["security-key", "client-device", "hybrid"]` | Preferred authenticator type hints presented to the Relying Party. |
 | `enterpriseAttestation` | Boolean | `false` | Enables support for enterprise-specific hardware attestation profiles. |
 | `attestationMode` | String | `"monitor"` | Options are: `disabled` (skip attestation checks), `monitor` (log/validate but allow credentials if attestation is absent/unknown), and `enforced` (fail credential creation if attestation check fails). |
+| `allowedTopOrigins` | Array of Strings | `[]` | Full origins permitted to frame a cross-origin ceremony, each written as scheme, host and optional port (for example `https://portal.example.com`). Empty — the default — denies every framed ceremony. See [Cross-origin ceremonies](#cross-origin-ceremonies). |
 
 ### Cross-origin ceremonies
 
-Registration and authentication ceremonies performed inside a cross-origin iframe are rejected. As WebAuthn
-Level 3 requires, the server reads the `crossOrigin` member of `CollectedClientData`: an absent member is
-treated as `false`, a value of `true` fails the request with `cross_origin_not_allowed`, and both a
-non-boolean value and an explicit `null` fail with `invalid_request`.
+As WebAuthn Level 3 requires, the server reads the `crossOrigin` member of `CollectedClientData`. An absent
+member is treated as `false`; both a non-boolean value and an explicit `null` fail with `invalid_request`.
 
-There is no configuration to permit a framed ceremony against a chosen set of framing origins yet, so a
-deployment that embeds the ceremony in a cross-origin iframe will stop working after this change.
+When `crossOrigin` is `true`, the ceremony is allowed only if its `topOrigin` — the origin of the page that
+framed it — appears in `allowedTopOrigins`. The request fails with `cross_origin_not_allowed` when:
+
+- `allowedTopOrigins` is empty, which is the default and denies every framed ceremony
+- `topOrigin` is absent, `null`, not a string, or blank
+- `topOrigin` is not listed
+
+Entries are compared against the whole origin, ignoring case and surrounding whitespace. A different scheme
+or port is a different origin, so `https://portal.example.com` does not permit `http://portal.example.com`.
+
+`allowedTopOrigins` is deliberately separate from the `origins` under `rp`. Those say which origin may
+*serve* a ceremony; this says which origin may *frame* one. Reusing the former would silently widen the
+framing policy of every existing deployment.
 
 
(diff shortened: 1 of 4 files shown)


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/3d19bfd8-5626-466f-a642-b20b419a6f09?refs=5c4859e697bb3b0ecee7759e5de187d28d095a7b&source=prChecksComment) |

yurem (approved):

Issue #14926, opened 2026-09-01
feat(jans-fido2): add Recommended v2.3 algorithms (ESP256, ESP384, Ed25519, Ed448, ML-DSA)


imran-ishaq:
**Is your feature request related to a problem? Please describe.**

The FIDO Server Requirements v2.3 algorithm table lists seven Recommended algorithms we do not support —
`ESP256`, `ESP384`, `Ed25519`, `Ed448`, and the three post-quantum `ML-DSA-44` / `-65` / `-87`.
`SignatureVerifier.getSignatureChecker()` throws `"Unknown mapping"` (`:113`) for anything outside its switch,
so an authenticator offering one of these cannot register at all.

Nothing is broken for existing users — no shipping authenticator negotiates these today. But "Recommended,
post-quantum" is the kind of line item that turns into a customer or certification question on a short horizon,
and it is better to know the dependency answer before we are asked.

**Describe the solution you'd like**

Two tranches, because the cost profile is very different.

*Cheap — new code points, not new crypto:*

- `ESP256` and `ESP384` are the fully-specified ECDSA variants, cryptographically identical to
  `SHA256withECDSA` / `SHA384withECDSA`.
- `Ed25519` maps to the same Ed25519 signature we already instantiate for `-8`.
- `Ed448` needs a provider check but is otherwise routine.

*Expensive — gated on a dependency review:*

ML-DSA-44/65/87 require a BouncyCastle capability review across both build variants **before** any commitment:

- Standard: `bc.version = 1.79` (`jans-bom/pom.xml:69`, `bcprov-jdk18on`)
- FIPS: `bc-fips (,1.0.2.4]`, `bcpkix-fips 1.0.8` (`jans-bom/pom.xml:315-323`) — older line, likely blocker
  for `server-fips/` parity.

The first deliverable here should be that review written up on the issue, not code. A documented "deferred,
here is why" is an acceptable outcome for the ML-DSA tranche.

**Describe alternatives you've considered**

- **Ship all seven together.** Rejected: holds four cheap additions hostage to a dependency review that may
  conclude "not yet possible on the FIPS line."
- **Do ML-DSA in the standard build only, skip FIPS.** Rejected as a default — it breaks `server-fips/` parity,
  which is a standing expectation for this repo. Would need an explicit decision.
- **Wait for a customer request.** Genuinely defensible for ML-DSA specifically. The four cheap ones are not
  worth deferring.

**Additional context**

[...]

Issue #14891, opened 2026-08-28
feat(jans-fido2): conform to FIDO Server Requirements v2.3 COSE algorithm table


imran-ishaq:
**Is your feature request related to a problem? Please describe.**

[FIDO Server Requirements v2.3](https://fidoalliance.org/specs/fidoserver/fido-server-v2.3-rd-20260226.html)
(Review Draft, 2026-02-26) is the document a FIDO2 server is certified against, and it ships an updated
algorithm table alongside [WebAuthn L3](https://www.w3.org/TR/2026/REC-webauthn-3-20260825/) and CTAP 2.3. Auditing jans-fido2 against that table turned up three problems, two of which are pre-existing defects rather than new spec surface:

1. **We advertise a Required algorithm we cannot process.** `AttestationService.preparePublicKeyCredentialSelection()`
   (`server/src/main/java/io/jans/fido2/service/operation/AttestationService.java:504-508`) offers
   `RS256(-257)`, `ES256(-7)` and `CoseEdDSAAlgorithm.Ed25519(-8)` whenever `enabledFidoAlgorithms` is unset —
   the default configuration. An Ed25519 credential returns a COSE key with `kty = OKP`, and
   `CoseService.createUncompressedPointFromCOSEPublicKey()` (`server/src/main/java/io/jans/fido2/service/CoseService.java:121-
   123`) is literally `case OKP: { throw new Fido2RuntimeException(...); }`. EdDSA and curve Ed25519 are both
   **Required** in the v2.3 table, so this is a conformance failure. `SignatureVerifier.getSignatureChecker():72`
   handles `-8` correctly, which is likely why it went unnoticed — the request never reaches the verifier.
   Every attestation processor (`PackedAttestationProcessor:140`, `AppleAttestationProcessor:178`,
   `AndroidKeyAttestationProcessor:139`, `U2FAttestationProcessor:217`) reaches it via
   `CoseService.getPublicKeyFromUncompressedECPoint():183`, and `AssertionVerifier:112` calls it directly.
   There is no alternative path.

2. **Some COSE code points in our enums are wrong, and two silently substitute a different algorithm.**
   `AttestationService.resolveEc2NumericValue()` (`:549-555`) passes an admin's configured algorithm name
   straight through `CoseEC2Algorithm.valueOf(...).getNumericValue()` into `pubKeyCredParams`. Configuring
   `ES384` advertises `-36`; configuring `ES512` advertises `-37`, which is an RSA algorithm. No error is
[...]

PR #15047, opened 2026-09-10
feat(jans-fido2): add ML-DSA-44, ML-DSA-65 and ML-DSA-87 post-quantum algorithms

### LINKED ISSUE #14937: feat(jans-fido2): add ML-DSA-44, ML-DSA-65 and ML-DSA-87 post-quantum algorithms

`ML-DSA-44`, `ML-DSA-65` and `ML-DSA-87` are Recommended in the FIDO Server Requirements v2.3 algorithm table
and unsupported here. `SignatureVerifier.getSignatureChecker()` throws `"Unknown mapping"` (`:114`) for
anything outside its switch, and `CoseService` has no branch for their key type.

Nothing is broken for users today — no shipping authenticator negotiates these. The reason to do it is
forward-looking: "Recommended, post-quantum" is the kind of line that becomes a customer or certification
question on a short horizon, and the dependency work has a long lead time.

Implement all three (`ML-DSA-44 = -48`, `ML-DSA-65 = -49`, `ML-DSA-87 = -50`): COSE constants, key decoding in `CoseService`, signature
verification in `SignatureVerifier`, and registration + assertion test coverage per parameter set. Verify
`server-fips/` parity explicitly rather than assuming it.

- **Standard build only, skip FIPS.** Breaks `server-fips/` parity, which is a standing expectation in this
  repo. Should not be the default; would need an explicit, recorded decision.
- **Wait for a customer request.** Genuinely defensible for ML-DSA specifically — more so than for any other
  item in this tree. If C3 concludes the provider support is not there, this is where it lands.
- **One parameter set (ML-DSA-65) first.** A reasonable de-risking step if the full three prove heavy.

**Additional context**
Sub-issue of #14926. Blocked on 14936— do not start before that review concludes. Also blocked on #14925: adding these to the verifier while the decoder cannot handle their key type would reproduce the exact advertise-then-fail bug this audit was opened for. Code points per IANA: ML-DSA-44 = -48, ML-DSA-65 = -49, ML-DSA-87 = -50.

imran-ishaq:
### Prepare

-------------------

#### Target issue

closes #14937

Sub-issue of #14926, under #14891. This is the last unblocked item in the v2.3 algorithm tree.

#### Implementation Details

**Why this is unblocked now.** The #14935 spike recommended deferring ML-DSA, and recorded exactly one
condition that would change the answer: land #14933 first. That merged as [#15031](https://github.com/JanssenProject/jans/pull/15031/). Before it, the advertised
algorithm list was hardcoded, so shipping ML-DSA standard-only would have made a FIPS deployment advertise
an algorithm it then failed to verify — the same advertise-then-fail shape as the EdDSA defect in #14929.
Now the advertised set is derived from real provider capability, so a FIPS deployment reports ML-DSA
unsupported and simply never offers it. The deferral reason is gone; the dependency situation itself has not
changed.

I re-ran the provider probe rather than trusting the earlier writeup:

```
bcprov-jdk18on 1.79   ML-DSA-44/65/87  keygen + KeyFactory + Signature OK, verify=true
bc-fips 1.0.2.4       ML-DSA-44/65/87  NoSuchAlgorithmException
```

`jans-fido2/server-fips/pom.xml` excludes `bcprov-jdk18on-*.jar` from the WAR and ships `bc-fips`, and
`SecurityProviderUtility.checkFipsMode()` selects the FIPS provider purely on that class being present. So
the FIPS build genuinely cannot do ML-DSA, and genuinely will not advertise it.

**The COSE values are verified against primary sources**, since a wrong code point is exactly the defect
class this tree was opened for. Each was confirmed twice — the IANA registry CSVs and RFC 9964:

| Item | Value | Source |
| --- | --- | --- |
| ML-DSA-44 / 65 / 87 | `-48` / `-49` / `-50` | IANA COSE Algorithms; RFC 9964 |
| Key type `AKP` | `7` | IANA COSE Key Types; RFC 9964 Figure 2 |
| `pub` parameter | `-1` | IANA COSE Key Type Parameters; RFC 9964 Figure 2 |
| Public key sizes | 1312 / 1952 / 2592 | RFC 9964 Table 1; matches the probe |

The change:

- `CoseKeyType` gains `AKP(7)`.
- `CoseMLDSAAlgorithm` carries the three code points plus their JCA names, because `ML-DSA-44` is not a
  legal Java identifier. `fromName` accepts both `ML-DSA-44` and `ML_DSA_44`, so an administrator can use
  the registry spelling.
- `CoseService` gains an AKP decode branch. **This is a genuinely new decode path, not a widened list** —
  AKP carries the entire public key in one `pub` parameter rather than the `(n, e)` pair or `(crv, x, y)`
  triple the other branches reassemble. The raw key is wrapped in a per-parameter-set SubjectPublicKeyInfo
  prefix; I read those 22-byte prefixes off provider-generated keys rather than hand-assembling the DER,
  since the embedded length fields differ per parameter set.
- `SignatureVerifier` maps `-48`/`-49`/`-50` to the provider. On the FIPS build this throws,
  `isSupported` returns false, and the algorithm is dropped from `pubKeyCredParams`.
- `AttestationService` gains a resolver so the names work in `enabledFidoAlgorithms`.

**Behaviour change.** None for existing deployments. Nothing is advertised that was not advertised before
unless an administrator opts in by naming an ML-DSA algorithm, and on FIPS even that is refused with a
logged error rather than honoured.

`server-fips/` needs no mirroring: the module is `pom.xml` and `target/` only, with no `src/` tree. Unlike
the earlier PRs in this tree, that is not the whole story here — the FIPS provider genuinely lacks these
algorithms, and the probe plus the derived-capability mechanism from #15031 are what make that safe rather
than merely structural.

-------------------

`CoseServiceMLDSATest` covers, per parameter set: the registration and assertion pair end-to-end (build the
COSE key an authenticator would send, decode it, assert it reconstructs byte-for-byte, then verify a real
signature made with the matching private key); a key of the wrong length; and a missing `pub` label. It also
covers a valid ML-DSA-65 key offered under ML-DSA-44, which the distinct key lengths and OIDs must reject.

The test that matters most asserts `isSupported` tracks **the running provider** rather than a list. That is
the invariant the whole standard-only decision rests on: if it ever reported support the provider does not
have, a FIPS deployment would advertise ML-DSA and fail the ceremony.

I checked the suite is not vacuous — pointing the verifier at a different algorithm for these code points
fails three of its cases, and I reverted after confirming.

```
CoseServiceMLDSATest         11 tests, 0 failures
jans-fido2 model + server   464 tests, 0 failures
```

Docs: `docs/janssen-server/fido/fido2-server-properties-config.md` adds the three names to
`enabledFidoAlgorithms` and gains a "Post-quantum algorithms (ML-DSA)" section stating plainly that they are
standard-build-only, and why that needs no action from an administrator — the same configuration stays safe
to share across both build variants. Carried in a separate `docs:` commit.

## Summary by CodeRabbit

* **New Features**
  * Added support for ML-DSA post-quantum algorithms in FIDO2 credentials and signature verification.
  * Supports ML-DSA-44, ML-DSA-65, and ML-DSA-87 for key decoding, attestation, and configured algorithm advertisement.
  * Automatically filters algorithms based on available runtime capabilities and validates key compatibility.

* **Documentation**
  * Documented ML-DSA configuration options, supported build types, capability filtering, error logging, and accepted hyphenated or underscored algorithm names.

### DIFF
--- a/docs/janssen-server/fido/fido2-server-properties-config.md
+++ b/docs/janssen-server/fido/fido2-server-properties-config.md
@@ -50,7 +50,7 @@ This nested block defines WebAuthn and FIDO2 attestation and assertion policy be
 | `unfinishedRequestExpiration` | Integer | `120` | Expiration time in seconds for incomplete registration/authentication requests. |
 | `metadataRefreshInterval` | Integer | `1296000` | Expiration time in seconds (e.g., 15 days) before checking and reloading the FIDO Alliance MDS TOC. |
 | <span id="servermetadatafolder">`serverMetadataFolder`</span> | String | `"/etc/jans/conf/fido2/server_metadata"` | Folder where local vendor metadata statement JSON files are placed manually. |
-| `enabledFidoAlgorithms` | Array of Strings | `["RS256", "ES256"]` | Enabled cryptographic signing algorithms allowed for credentials. Accepted names: `RS256`, `RS384`, `RS512`, `RS65535`, `PS256`, `PS384`, `PS512`, `ES256`, `ES384`, `ES512`, `EdDSA` — the algorithms the server can both advertise and complete a registration with. When unset, the server advertises `RS256`, `ES256` and `EdDSA`. An unrecognised name is ignored. A recognised name the deployment cannot actually complete a registration with is logged at `ERROR` and left out of `pubKeyCredParams` — see [Advertised algorithms](#advertised-algorithms). |
+| `enabledFidoAlgorithms` | Array of Strings | `["RS256", "ES256"]` | Enabled cryptographic signing algorithms allowed for credentials. Accepted names: `RS256`, `RS384`, `RS512`, `RS65535`, `PS256`, `PS384`, `PS512`, `ES256`, `ES384`, `ES512`, `EdDSA`, `ML-DSA-44`, `ML-DSA-65`, `ML-DSA-87` — the algorithms the server can both advertise and complete a registration with. When unset, the server advertises `RS256`, `ES256` and `EdDSA`. An unrecognised name is ignored. A recognised name the deployment cannot actually complete a registration with is logged at `ERROR` and left out of `pubKeyCredParams` — see [Advertised algorithms](#advertised-algorithms). |
 | `rp` | Array of Objects | `[ { "id": "https://jans.io", "origins": ["jans.io"] } ]` | Relying Party (RP) configuration mapping expected IDs to valid origins. |
 | `metadataServers` | Array of Objects | `[ { "url": "https://mds.fidoalliance.org/" } ]` | External FIDO Metadata Service endpoints to download statement catalogs. |
 | `disableMetadataService` | Boolean | `false` | If set to `true`, the FIDO2 server skips validating authenticators against the MDS3 service. |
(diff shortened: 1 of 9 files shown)


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/909554a4-f1bb-442f-b057-9b9908a185ba?refs=9615617090daab7ac25dd395751b399b1bdf3894&source=prChecksComment) |

yurem (approved):

PR #14700, opened 2026-08-04
feat(jans-core): allow extend types and move metric to separate module

yurem:
### Prepare

- [x] Read [PR guidelines](https://github.com/JanssenProject/jans/blob/main/docs/CONTRIBUTING.md#prs)
- [x] Read [license information](https://github.com/JanssenProject/jans/blob/main/LICENSE)

-------------------

#### Target issue
#14699
  
  
  
closes #14699

#### Implementation Details
  

-------------------
### Test and Document the changes
- [x] Static code analysis has been run locally and issues have been fixed
- [x] Relevant unit and integration tests have been added/updated
- [x] Relevant documentation has been updated if any (i.e. user guides, installation and configuration guides, technical design docs etc)

Please check the below before submitting your PR. The PR will not be merged if there are no commits that start with `docs:` to indicate documentation changes or if the below checklist is not selected.
- [x] **I confirm that there is no impact on the docs due to the code changes in this PR.**

## Summary by CodeRabbit

- **New Features**
  - Added a unified metrics framework supporting counters, timers, custom metric types, subtypes, and metric reports.
  - Added metric subtype storage and indexing across directory and database schemas.
  - Added asynchronous processing support for annotated operations.

- **Bug Fixes**
  - Corrected node identifier naming across metric services.

- **Refactor**
  - Updated audit API schemas with clearer field names, status values, telemetry intervals, and refreshed examples.
  - Deprecated the legacy metric type definition in favor of an extensible model.
  - Improved metric persistence and subtype-based filtering.

PR #13851, opened 2026-04-17
feat(jans-fido) : add trusted proxy validation to prevent client IP s…

imran-ishaq:
Description

MetricService.java in jans-fido2 was unconditionally reading proxy headers (X-Forwarded-For, Proxy-Client-IP, WL-Proxy-Client-IP, etc.) from every incoming HTTP request without validating that the request actually originated from a trusted reverse proxy. Since HTTP headers are client-controlled strings, any attacker with direct access to the FIDO2 endpoint could inject an arbitrary IP into X-Forwarded-For and have that spoofed address recorded in metrics and audit logs.

This PR resolves the two TODO comments that explicitly flagged the gap at MetricService.java lines 567–568.

Changes:

AppConfiguration.java — added two new opt-in fields:

trustedProxyEnabled (Boolean, default null) — null preserves legacy behaviour with zero regression for existing deployments; false disables proxy header trust entirely; true enables conditional trust via CIDR validation
trustedProxyIpRanges (List<String>) — list of trusted reverse-proxy CIDR blocks (e.g. ["10.0.0.0/8", "192.168.1.0/24"])
MetricService.java — extractIpAddress() now validates remoteAddr against the configured CIDR ranges before reading any proxy header. Requests from IPs outside the trusted ranges fall back to remoteAddr directly. Added pure-Java helpers isFromTrustedProxy() and isIpInCidr() (IPv4 + IPv6, no new dependency). Simplified isValidIpAddress() to use InetAddress.getByName().

Target issue

Fixes: MetricService.java blindly trusts proxy headers (X-Forwarded-For etc.) without validating the source IP, allowing client IP spoofing in FIDO2 metrics and audit records. TODOs at lines 567–568 of MetricService.java flagged this as a known gap requiring a configuration option and trusted proxy IP range validation.
  
  
  
closes #13850

#### Implementation Details
  

-------------------
### Test and Document the changes
- [x] Static code analysis has been run locally and issues have been fixed
- [ ] Relevant unit and integration tests have been added/updated
- [ ] Relevant documentation has been updated if any (i.e. user guides, installation and configuration guides, technical design docs etc)

Please check the below before submitting your PR. The PR will not be merged if there are no commits that start with `docs:` to indicate documentation changes or if the below checklist is not selected.
- [ ] **I confirm that there is no impact on the docs due to the code changes in this PR.**

## Summary by CodeRabbit

* **New Features**
  * Configurable proxy-header trust for client IP extraction: legacy (unset), enabled, or disabled.
[...]

PR #15095, opened 2026-09-15
fix(jans-fido2): trust proxy headers only from configured reverse proxies

### LINKED ISSUE #13850: fix(jans-fido2): Blind proxy header trust in MetricService allows client IP spoofing
Bug Description

MetricService.java in jans-fido2 unconditionally reads proxy headers (X-Forwarded-For, Proxy-Client-IP, WL-Proxy-Client-IP, etc.) from every incoming HTTP request without verifying whether the request originated from a trusted reverse proxy.

Since HTTP headers are client-controlled values, an attacker who can directly access the FIDO2 endpoint can spoof these headers and inject arbitrary IP addresses into metrics and audit logs.

The code itself already acknowledged this issue via TODO comments:

Add configuration option to enable/disable proxy header trust
Add validation to ensure request came from trusted proxy IP range
Steps to Reproduce
Deploy jans-fido2 in an environment where the FIDO2 endpoint is directly accessible (no reverse proxy or proxy bypass possible).
Send a request with a forged proxy header, e.g. X-Forwarded-For set to a fake IP.
Check the stored metrics or audit logs.
Observed Behavior

The system records the spoofed IP address from the header instead of the real client IP address.

Expected Behavior

Proxy headers should only be trusted when the request originates from a known and trusted reverse proxy. Specifically:

A configuration flag (trustedProxyEnabled) should control whether proxy headers are trusted.
When enabled:
The direct connection address (remoteAddr) must be validated against configured CIDR ranges (trustedProxyIpRanges).
If remoteAddr is not within trusted ranges, proxy headers must be ignored.
When disabled:
Always use remoteAddr.
When not configured:
Maintain full backward compatibility (legacy behavior).

imran-ishaq:
### Prepare

-------------------

#### Target issue

closes #13850

Replaces #13851, which is being closed. Same design, re-implemented against current `main` — that branch
was 401 commits behind, and the code it changed has since been rewritten by #14653 and relocated by
#14700, so reconciling the diff was more work and more risk than porting the approach.

#### Implementation Details

`MetricService` derived the recorded client IP from proxy headers supplied by the caller, with no check on
who the caller was. Any client able to reach a FIDO2 endpoint could choose the address stored against its
own ceremony. The method's own javadoc already recorded the caveat, and the `TODO` comments named the fix.

`trustedProxyEnabled` now governs what is believed:

| Value | Behaviour |
| --- | --- |
| unset | Legacy — headers trusted unconditionally |
| `false` | Headers never read; socket address used |
| `true` | Headers read only when the socket address is inside `trustedProxyIpRanges` |

**The default is unset, so merging this changes no existing deployment's behaviour.** It ships the
mechanism; a deployment that wants a trustworthy recorded address has to opt in. That is deliberate —
silently starting to ignore headers would break every deployment that is correctly fronted by a proxy
today — but it does mean this PR alone does not close the exposure anywhere.

An empty `trustedProxyIpRanges` with the flag on trusts **nothing**, and logs a warning. #13851 was
self-contradictory here: its config javadoc said an empty list trusted all source IPs while its
implementation ignored headers. "Enable the protection, trust everyone" is the dangerous reading of the
two, so this fails closed and the docs and code agree.

Three details beyond what #13850 specifies:

- **`X-Forwarded-For` is read right to left**, skipping hops that are themselves trusted proxies. The
  leftmost entry is attacker-controlled — a client prepends before the real proxy appends — so taking it
  would leave the spoof open even with trust configured.
- **Both sides of a CIDR comparison must be IP literals.** `InetAddress.getByName` resolves hostnames, and
  this runs on the request path for every metrics write, so a mistyped range would otherwise become a
  blocking DNS lookup. `isIpLiteral` rejects and logs instead.
- **IPv4-mapped normalisation**, so `::ffff:10.1.2.3` matches an IPv4 range — a dual-stack JVM can report
  the connecting address in that form, and without this the proxy stops being recognised depending on how
  the JVM was started.

CIDR matching is pure `InetAddress`, no new dependency, IPv4 and IPv6.

**A limitation reviewers should weigh.** Where the reverse proxy runs on the same host, requests reach the
FIDO2 server from `127.0.0.1` — and so do any sent directly to it. Trusting loopback therefore does not
distinguish the proxy from a direct caller, so that topology cannot be fully secured by this mechanism
alone. This also interacts with #14801, which needs the Auth Server relay's `X-Forwarded-For` to be
believed: enabling the check with a narrow range would break it. The unset default means nothing breaks on
merge, but the two cannot both be satisfied by configuration in that topology. Flagging it rather than
leaving it to be discovered. @yurem — this is the same boundary question from the #14801 thread.

-------------------

13 cases added to `MetricServiceTest`, covering each of the three configuration states, the empty-ranges
fail-closed path, an untrusted caller being ignored, a trusted caller believed, a chain spoofed on the left
resolving to the correct hop, every hop being a trusted proxy, and CIDR behaviour for IPv4, IPv6,
IPv4-mapped, bare addresses, out-of-range prefixes and non-literals. `extractIpAddress` was made
package-private, matching how `isValidIpAddress` is already exercised by this suite.

I checked the guard is not vacuous — short-circuiting the trusted-proxy check fails
`extractIpAddress_whenCallerIsNotATrustedProxy_ignoresItsHeaders` and nothing else, and I reverted after
confirming.

One bug was caught by these tests during development: reusing the existing `isValidIpAddress` as the
literal guard rejects `::ffff:a.b.c.d`, because its IPv6 pattern forbids dots, and accepts `localhost`,
which is a name. Hence the separate `isIpLiteral`.

```
MetricServiceTest           32 tests, 0 failures
jans-fido2 model + server  501 tests, 0 failures
```

Docs: `docs/janssen-server/fido/fido2-server-properties-config.md` documents both properties in the
top-level table and adds a "Client IP in metrics" section covering the three states, the right-to-left
chain rule, the CIDR notation accepted, and the same-host caveat above. Carried in a separate `docs:`
commit.

## Summary by CodeRabbit

* **New Features**
  * Added configurable trusted-proxy settings for client IP detection in FIDO2 metrics.
  * Supports IPv4 and IPv6 CIDR ranges, trusted forwarding chains, and secure handling of forwarded headers.
  * Metrics fall back to the connecting address when forwarded information is unavailable or untrusted.

* **Documentation**
  * Documented proxy settings, supported headers, client IP resolution, and accepted address formats.

* **Tests**
  * Added coverage for proxy trust modes, CIDR matching, IPv4/IPv6 handling, invalid configurations, and fallback behavior.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/6af5588d-1879-474a-a7ba-53ed17351966?refs=e8637237415fa4f44f540311896560bb55c53c17&source=prChecksComment) |

yurem (approved):

PR #15101, opened 2026-09-16
fix(jans-fido2): close the two gaps left by the trusted-proxy validation

### LINKED ISSUE #15097: fix(jans-fido2): legacy proxy headers bypass trusted-proxy validation in MetricService
**Describe the bug**

Follow-up to [#13850](https://github.com/JanssenProject/jans/issues/13850) / [#15095](https://github.com/JanssenProject/jans/pull/15095/). With `trustedProxyEnabled=true`, a client can still choose the IP recorded
against its own ceremony by using a proxy header other than `X-Forwarded-For`.

`MetricService.extractFromTrustedProxy` (`MetricService.java:770`) confirms the connecting address is a
trusted proxy, then reads `X-Forwarded-For` right to left, skipping hops that are themselves trusted. That
part is correct. But when `X-Forwarded-For` yields nothing usable, it falls through at
`MetricService.java:796` to `firstUsableSingleValueHeader` (`:838`), which returns the first value that
parses as an IP from any of the eight alternative headers declared at `:720`:

Proxy-Client-IP, WL-Proxy-Client-IP, HTTP_X_FORWARDED_FOR, HTTP_X_FORWARDED,
HTTP_X_CLUSTER_CLIENT_IP, HTTP_CLIENT_IP, HTTP_FORWARDED_FOR, HTTP_FORWARDED

That value is never checked against `trustedProxyIpRanges`. A reverse proxy normally overwrites
`X-Forwarded-For` but passes other request headers through untouched, so a header the client set arrives
intact and is recorded verbatim.

The effect is that the protection added in [#15095](https://github.com/JanssenProject/jans/pull/15095/) covers one header and not the other eight. This is the
same defect [#13850](https://github.com/JanssenProject/jans/issues/13850) describes, reachable through a different header, in the mode that exists to prevent it.

**To Reproduce**

1. Configure `trustedProxyEnabled: true` and `trustedProxyIpRanges` containing the reverse proxy's address.
2. Through that proxy, send a request to a FIDO2 endpoint that records metrics, with **no**
   `X-Forwarded-For` header and `Proxy-Client-IP: 203.0.113.9`.
3. Read the resulting `jansFido2MetricsEntry`.

**Observed behavior**

`ipAddress` is `203.0.113.9`. The address is client-chosen, and neither the connecting address nor any
trusted-range check constrained it.

**Expected behavior**

With trust enabled, no address should be recorded that has not passed the same check `X-Forwarded-For`
values pass. Two defensible resolutions, and the second is probably right:

1. Apply the trusted-range check to the alternative headers as well, as is already done per hop for
   `X-Forwarded-For`.
2. Ignore the alternative headers entirely when `trustedProxyEnabled=true`, and fall back to the connecting
   address. They exist for legacy container behaviour; a deployment that has explicitly declared its
   proxies is stating that `X-Forwarded-For` is the contract. This is the smaller and more obviously
   correct change.

Either way, legacy mode (`trustedProxyEnabled` unset) must keep reading all nine headers exactly as it does
now — that path is relied on for backward compatibility and was itself the subject of a regression fix
during review of #15095.

**Additional context**

Raised by CodeRabbit on [#15095](https://github.com/JanssenProject/jans/pull/15095/) as *"Do not fall back to unsanitized legacy proxy headers in trusted mode"*
(`MetricService.java:796`). The comment was posted outside the diff range in the same review that approved
the PR, so it was not addressed before merge. Filed rather than reopening [#13850](https://github.com/JanssenProject/jans/issues/13850), since that issue's
mechanism did land — this is the gap left in it.

Introduced in [#15095](https://github.com/JanssenProject/jans/pull/15095/) (`cd7e42db4a9`). No released version is affected while `trustedProxyEnabled` remains
unset, which is the default, because legacy mode trusts every header by design.

### LINKED ISSUE #15098: fix(jans-fido2): IPv4-mapped CIDR ranges in trustedProxyIpRanges are silently rejected
**Describe the bug**

Follow-up to [#13850](https://github.com/JanssenProject/jans/issues/13850) /[ #15095](https://github.com/JanssenProject/jans/pull/15095/). An IPv4-mapped IPv6 range written with an IPv6-scale prefix, such as
`::ffff:10.0.0.0/104`, is rejected and dropped instead of being honoured. The administrator believes a
proxy is trusted; it is not.

`MetricService.isIpInCidr` (`MetricService.java:884`) normalises both operands through `normalizeAddress`
(`:961`), which converts an IPv4-mapped IPv6 address to its 4-byte IPv4 form so that it compares correctly
against IPv4 ranges on a dual-stack JVM. That normalisation happens *before* the prefix length is
validated at `:901`:

```java
int maxBits = cidrBytes.length * Byte.SIZE;
int prefixLength = (parts.length == 2) ? Integer.parseInt(parts[1].trim()) : maxBits;
if ((prefixLength < 0) || (prefixLength > maxBits)) {
    // rejected
}
```

Once ::ffff:10.0.0.0 has been narrowed to 4 bytes, maxBits is 32, so a prefix of 104 — the correct
IPv6 expression of an IPv4 /8 — fails the bounds check and the range is discarded.

The prefix is interpreted against the normalised address while the administrator wrote it against the
address as typed. The two disagree by the 96 bits the mapping prefix occupies.

To Reproduce

1. Configure trustedProxyEnabled: true and trustedProxyIpRanges: ["::ffff:10.0.0.0/104"].
2. Send a request through a reverse proxy whose address is 10.1.1.1, carrying
   X-Forwarded-For: 203.0.113.9.
3. Read the resulting jansFido2MetricsEntry, and the server log.

Observed behavior

The range is skipped with Ignoring trusted proxy range '::ffff:10.0.0.0/104': prefix length 104 is out of range, the proxy is not recognised as trusted, and the connecting address is recorded instead of the
forwarded one. With that as the only configured range, the deployment behaves as though
trustedProxyIpRanges were empty.

Expected behavior

A mapped range should be honored with the prefix the administrator wrote. When the address is narrowed
from 16 bytes to 4, a prefix expressed against the 16-byte form should be reduced by the 96 bits of the
mapping prefix rather than rejected, so ::ffff:10.0.0.0/104 and 10.0.0.0/8 select the same set. A
prefix below 96 on a mapped address is genuinely out of range and should still be rejected.

Plain IPv4 (10.0.0.0/8) and plain IPv6 (2001:db8::/32) ranges already behave correctly and must be
unaffected.

Additional context

Raised by CodeRabbit on [#15095](https://github.com/JanssenProject/jans/pull/15095/) as "Normalize IPv4-mapped CIDR prefixes before validation" (MetricService.java:902). Posted outside the diff range in the same review that approved the PR, so it
was not addressed before merge.

Lower impact than the sibling issue, because it fails closed and is logged — the range is ignored rather
than wrongly matched, so no untrusted proxy is ever believed as a result. The cost is a silently
ineffective configuration.

Introduced in [#15095](https://github.com/JanssenProject/jans/pull/15095/) (cd7e42db4a9). Only reachable with trustedProxyEnabled=true, which is not the default.

imran-ishaq:
### Prepare

-------------------

#### Target issue

closes #15097
closes #15098

Both are follow-ups to #15095, raised by CodeRabbit in the review that approved it — the comments landed
outside the diff range and were not addressed before merge. They sit in the same method family, so
separating them would mean two PRs editing the same code in conflict.

#### Implementation Details

**#15097 — alternative proxy headers bypassed the trusted-range check.** With `trustedProxyEnabled=true`,
`extractFromTrustedProxy` verified the connecting address and walked `X-Forwarded-For` correctly, then fell
back to eight other proxy headers whose values were never checked against `trustedProxyIpRanges`. A reverse
proxy overwrites `X-Forwarded-For` but passes other request headers through as the client sent them, so a
client could set `Proxy-Client-IP` and have it recorded verbatim — the original spoofing defect, reachable
through a different header, in the mode that exists to prevent it.

Trusted mode now consults `X-Forwarded-For` only, and otherwise records the connecting address. This is
option 2 from #15097: a deployment that has declared its proxies is stating that `X-Forwarded-For` is the
contract, and it is a smaller change than trust-checking eight more headers.
`firstUsableSingleValueHeader` had exactly one caller and is removed. Legacy mode still reads all nine,
unchanged.

`SINGLE_VALUE_PROXY_HEADERS` is renamed to `LEGACY_PROXY_HEADERS`. The old name was wrong twice over: the
array is now legacy-only, and legacy mode splits those headers on commas, so they were never single-value.

**#15098 — IPv4-mapped ranges written with an IPv6 prefix were dropped.** `normalizeAddress` narrows a
mapped range to four bytes before the prefix is bounds-checked, so `::ffff:10.0.0.0/104` — the correct IPv6
expression of `10.0.0.0/8` — failed `104 > 32` and was discarded with a warning. The administrator believed
a proxy was trusted; it was not. The prefix is now re-expressed against the normalised address.

Two details worth a reviewer's attention:

- **The mapped form is detected from the text, not the parsed type.**
  `InetAddress.getByName("::ffff:10.0.0.0")` returns an `Inet4Address` — Java collapses the literal at parse
  time, so the parsed object no longer records how it was written, while the prefix the administrator chose
  does. An `instanceof Inet6Address` check silently never fires; the first version of this fix used one and
  the tests caught it.
- **A prefix of 32 or less is deliberately left alone.** Because of that same collapsing,
  `::ffff:10.0.0.0/8` is accepted on `main` today and behaves as IPv4 `/8` — a natural thing to configure
  after copying an address out of a log on a dual-stack JVM. Re-reading every mapped prefix on the IPv6
  scale would have silently stopped honouring a range already in use, so only a prefix above 32, which
  cannot be an IPv4 mask, is treated as IPv6-scale. Between 33 and 95 a mapped prefix covers part of the
  mapping itself and is meaningless either way; those remain rejected.

Net effect on mapped ranges, relative to `main`:

| prefix | before | after |
| --- | --- | --- |
| ≤ 32 (`/8`) | honoured as IPv4 | unchanged |
| 96–128 (`/104`) | rejected | honoured |
| 33–95 | rejected | unchanged |

**No existing deployment is affected.** `trustedProxyEnabled` is unreleased — no tag contains
`cd7e42db4a9` and the current release is v2.1.0 — so nothing can yet depend on the behaviour #15097
changes, and the #15098 change is a strict superset of what `main` accepts.

-------------------

11 cases added to `MetricServiceTest` (46 in the class): four alternative headers ignored in trusted mode
and still read in legacy mode, mapped ranges at `/104` and `/120`, a bare mapped address, an IPv4-scale
prefix on a mapped range left alone, and 33/95 rejected.

Two checks beyond the suite passing:

- **Sabotage.** Reinstating the trusted-mode fallback and dropping the prefix adjustment fails 7 tests,
  exactly the new ones and nothing else.
- **Backward compatibility, measured rather than argued.** Running this branch's tests against `main`'s
  `MetricService` fails exactly 5 — the four `ignoresTheAlternativeProxyHeaders` cases and
  `honoursIpv4MappedRangesWrittenWithAnIpv6Prefix`. Every other case passes against both, including
  `leavesAnIpv4ScalePrefixOnAMappedRangeAlone`, which is what demonstrates the `/8` form keeps working.

```
MetricServiceTest           46 tests, 0 failures
jans-fido2 model + server  515 tests, 0 failures
```

Docs: `docs/janssen-server/fido/fido2-server-properties-config.md` records that trusted mode consults only
`X-Forwarded-For`, and how a mapped range's prefix is read. Carried in a separate `docs:` commit.

## Summary by CodeRabbit

- **Bug Fixes**
  - Trusted-proxy mode now uses only `X-Forwarded-For` for client IP detection and falls back to the direct connection address when no usable value is available.
  - Legacy mode continues to support alternative proxy headers.
  - IPv4-mapped IPv6 CIDR ranges now match consistently, including mapped IPv4-scale prefixes, while unsupported prefix lengths are rejected.

- **Documentation**
  - Clarified proxy-header handling, fallback behavior, and IPv4-mapped IPv6 CIDR configuration rules.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/b43e95b6-d038-4ac8-b4a4-b0deca264238?refs=6c8f014db64f46d5fa15f18d836834ee445ca2bf&source=prChecksComment) |

yurem (approved):

Issue #15112, opened 2026-09-17
fix(jans-cli-tui): preserve the per-RP FIDO2 policy when editing a relying party


imran-ishaq:
**Describe the bug**

Sub-issue of #14516, and filed as a bug rather than a feature because of what the save path does.

`jans-cli-tui/cli_tui/plugins/020_fido/main.py:294` rebuilds each relying party from scratch on save, from
only the two columns the table holds:

```python
fido2_static['rp'] = []
for name, domains in self.requested_parties_container.data:
    fido2_static['rp'].append({'id': name, 'origins': domains.splitlines()})

The table itself is built from the same two fields (:153, headers ['id', 'origins'] at :157). So once
RequestedParty carries a policy, an administrator who opens the FIDO2 screen and saves — even without
touching the relying parties — silently discards the policy of every RP. Nothing warns, and the loss is
only visible later as attestation behaviour quietly reverting to the global mode.

This is a data-loss path, not a missing feature. It is worth landing close behind the model change rather
than being left as follow-up polish.

To Reproduce

1. Configure an RP with a policy (for example attestationMode: enforced), by Config-API or directly.
2. Open jans-cli-tui, go to the FIDO2 plugin, and save without editing anything.
3. Read the configuration back.

Observed behavior

The RP's policy is gone. Its attestationMode silently falls back to the global value.

Expected behavior

Editing or saving preserves any part of a relying party the screen does not manage, and the policy is
viewable and editable alongside the existing id and origins.

Describe the solution you'd like

Two parts, and the first matters more:

1. Stop discarding unknown fields. Carry the original RP objects and update them, rather than
   reconstructing {'id': ..., 'origins': ...}. That also makes the screen robust against the next field
   added to the model — of which #14516 lists several.
2. Add the policy to the RP dialog so it can be edited. The dialog already reads its field help from the
   published schema —
   get_schema_from_reference('Fido2', '#/components/schemas/RequestedParty') at :52, used via
   get_help_from_schema(...) at :55 — so the swagger from the Config-API sub-issue supplies the help
   text for free, provided that lands first.

Additional context

Depends on the model sub-issue, and on the Config-API sub-issue for the schema-driven help. Part 1 is worth
doing even if part 2 is deferred, since it is what prevents the data loss.

Issue #15111, opened 2026-09-17
docs(jans-config-api): add the per-RP FIDO2 policy to the fido2-plugin swagger


imran-ishaq:
**Is your feature request related to a problem? Please describe.**

Sub-issue of #14516, which lists the Config-API plugin and its swagger among the surfaces that must carry
the per-RP policy.

Checking what the plugin actually does narrows this considerably. `Fido2ConfigResource` does not map fields
individually — `getFido2Configuration()` returns the whole `AppConfiguration` and
`updateFido2Configuration()` takes the whole object and merges it:

```java
public Response getFido2Configuration() {
    AppConfiguration appConfiguration = this.fido2Service.find();
    return Response.ok(appConfiguration).build();
}

So once RequestedParty carries a policy field, the REST API serialises and accepts it through Jackson
with no resource or service change. The endpoint is already there: @GET and @PUT on
Constants.FIDO2_CONFIG.

What does not follow automatically is the swagger, which is hand-maintained. Today
jans-config-api/plugins/docs/fido2-plugin-swagger.yaml:1732 declares:

    RequestedParty:
      type: object
      properties:
        id:
          type: string
        origins:
          type: array
          items:
            type: string

Left alone it silently drifts: the API accepts and returns a field its own published contract does not
document, so a client generated from the spec drops it.

Describe the solution you'd like

- Add the policy object to the RequestedParty schema in fido2-plugin-swagger.yaml, mirroring the
  model, with the same "unset falls back to the global value" wording used on the model field.
- A round-trip test: PUT a configuration whose RP carries a policy, GET it back, and assert the policy
  survives — that is the thing that would otherwise regress unnoticed, since nothing else exercises it.

Describe alternatives you've considered

- Rely on Jackson and skip the swagger. Rejected: the spec is the published contract, generated clients are
  built from it, and it is also what drives the CLI-TUI help text (see the CLI-TUI sub-issue), so the drift
  propagates.

Additional context

Depends on the model sub-issue. An RP with no policy must serialise exactly as it does today, so existing
API clients are unaffected. Smaller than it first appears — schema plus a test, not plugin code.

Issue #14516, opened 2026-07-08
feat(jans-fido2): Per-RP assurance policy on RequestedParty (move from global config)


imran-ishaq:
**Is your feature request related to a problem? Please describe.**
Right now jans-fido2 makes every passkey/WebAuthn assurance decision globally, in jans fido2/model/src/main/java/io/jans/fido2/model/conf/Fido2Configuration.java (attestationMode, enabledFidoAlgorithms, enterpriseAttestation, hints, metadataServers,  ...). A relying party - jans-fido2/model/src/main/java/io/jans/fido2/model/conf/RequestedParty.java - carries only id and origins.

So an operator hosting multiple relying parties on one FIDO2 server cannot apply different security policies per RP. Every RP is forced to share one global configuration. For example, on the same server you can't have:

- RP-A: low-friction consumer passwordless login (defaults), and
- RP-B: a higher-assurance app that must require user verification, enforce attestation, restrict registration to an 
  allow-list of authenticator models, forbid synced/exportable credentials, and disallow weak account-recovery  
  paths.

As passkeys move into regulated and high-assurance deployments, this per-RP flexibility is increasingly a hard requirement, and today it's simply not expressible.

**Describe the solution you'd like**
Extend RequestedParty with a per-RP assurance/policy object (or fields) so assurance decisions can be set per relying party instead of only globally:

- requiredUserVerification - require the UV (user-verification) flag for this RP
- attestationMode - monitor/enforced, resolved per RP
- allowedAaguids / blockedAaguids + minCertificationLevel - authenticator allow/deny, checked against the FIDO 
   metadata service
- allowSyncedPasskeys - device-bound-only vs. synced/multi-device credentials
- allowedTransports - usb / nfc / ble / internal / hybrid (convenience hint)
- recoveryRequirements / fallbackRules - account-recovery assurance, plus the option to let users disable  
   password login once a passkey is registered
- (optional) an assuranceLevel preset that expands to sensible defaults for the above

Resolution rule (backward compatibility is mandatory): enforcement reads the per-RP policy first, falling back to the existing global Fido2Configuration values when a field is unset. A RequestedParty with no policy MUST behave exactly as it does today - zero regression for existing Casa/Agama flows. The resolved policy is what the attestation/assertion verifiers consult.

Surface the new fields end-to-end:
- Config-API plugin jans-config-api/plugins/fido2-plugin + its swagger jans-config-api/plugins/docs/fido2-
  plugin-swagger.yaml
[...]

PR #15115, opened 2026-09-17
feat(jans-fido2): per-RP attestationMode — policy model, resolver and enforcement

### LINKED ISSUE #15110: feat(jans-fido2): per-RP attestationMode — policy model, resolver and enforcement

Sub-issue of #14516. `attestationMode` is read in eight places today, each going directly to
`appConfiguration.getFido2Configuration().getAttestationMode()`, so every relying party on a server is
forced to share one attestation policy.

This sub-issue delivers the policy model, the resolution rule, and the enforcement wiring together. They
are kept in one change deliberately: the model on its own would put a configuration field into `main` that
nothing reads, which looks to an administrator like a policy in effect when it is not.

- `RequestedPartyPolicy` model, and a nullable `policy` field on `RequestedParty`. Null means "not set for
  this RP", not a default.
- A resolver returning the per-RP value when set and the global `Fido2Configuration` value otherwise. This
  makes #14516's backward-compatibility guarantee — "a RequestedParty with no policy MUST behave exactly as
  it does today" — one testable rule instead of a convention repeated at eight call sites. The RP is
  matched on origin exactly as `AttestationService.createRpDomain` matches it, so the policy that applies
  and the RP a ceremony is attributed to cannot disagree.
- Every live `attestationMode` read routed through it.

**All eight reads, or none.** Wiring only the easily reachable ones would let an administrator set
`enforced` on one RP and have it honoured in the attestation processors but silently ignored in the MDS
trust checks. Partial enforcement of a security policy is worse than uniform global enforcement.

The RP identity is already available: `Fido2RegistrationData` carries `origin` and `rpId`, is persisted
during `options()`, and is passed to every attestation processor. Reaching the deeper consumers means
adding an origin parameter to five internal methods along this chain, with no processor interface change:

process -> processAttestation -> processX5cAttestation -> getMetaDataFromCertificates
        -> getMetadataForU2fAuthenticator -> isAttestationEnforced

`attestationMode` is the only policy field added. Every other field from #14516's list arrives with the
change that enforces it, for the same reason given above.

- Ship the model first and the enforcement later. Rejected: leaves an unread field in a release.
- Resolve inline at each call site instead of a service. Rejected: the fallback rule would be restated
  eight times and could drift.
- Add all of #14516's policy fields now. Rejected: unenforced config is misleading config.

**Additional context**

`AttestationTrustPolicy` also reads the global value but is dead code — its only references are its own
file and its test, and nothing in the server calls it. It is left alone here rather than given a signature
no caller uses; it should be deleted or annotated separately, otherwise a future caller silently gets
global semantics while everything around it is per-RP.

Config-API and CLI-TUI surfacing are separate sub-issues.

Acceptance: an RP with `attestationMode` set has it honoured at every enforcement point; an RP without one,
or an origin matching no RP, behaves exactly as today.

imran-ishaq:
### Prepare

-------------------

#### Target issue

closes #15110

Sub-issue of #14516. The other two sub-issues — #15111 (Config-API swagger) and #15112 (CLI-TUI) — are
deliberately not in this PR; see "Known gap" below.

#### Implementation Details

`attestationMode` was read in eight places, each going straight to
`appConfiguration.getFido2Configuration().getAttestationMode()`, so every relying party on a server shared
one attestation policy.

**Model.** `RequestedPartyPolicy`, and a nullable `policy` on `RequestedParty`. Null means "not set for
this RP", never a default.

**Resolver.** `RpPolicyService.resolveAttestationMode(origin)` returns the per-RP value when set and the
global value otherwise. This is where #14516's mandatory guarantee lives — *"a RequestedParty with no
policy MUST behave exactly as it does today"* — as one testable rule rather than a convention restated at
eight call sites. The origin is matched exactly as `AttestationService.createRpDomain` matches it, so the
policy that applies and the RP a ceremony is attributed to cannot disagree.

**Enforcement: all eight reads, or none.** Wiring only the easily reachable ones would let an administrator
set `enforced` on one RP and have it honoured in the attestation processors but silently ignored in the MDS
trust checks. Partial enforcement of a security policy is worse than uniform global enforcement, so this
converts every live read.

The RP identity was already available — `Fido2RegistrationData` carries `origin` and `rpId`, is persisted
during `options()` at `AttestationService:207-209`, and is passed to every processor at `:293`. Reaching
the deeper consumers meant adding an `origin` parameter to five internal methods:

```
process -> processAttestation -> processX5cAttestation -> getMetaDataFromCertificates
        -> getMetadataForU2fAuthenticator -> isAttestationEnforced
```

No `AttestationFormatProcessor` interface change was required.

`attestationMode` is the only policy field added. Every other field #14516 lists arrives with the change
that enforces it — a field stored but never read looks to an administrator like a policy in effect when it
is not.

**Two things a reviewer should know.**

`AttestationTrustPolicy` also reads the global value and is deliberately left alone: it is dead code, with
no references outside its own file and its test. Giving it a signature no caller uses adds risk for nothing.
It should be deleted or annotated separately — otherwise a future caller silently gets global semantics
while everything around it is per-RP.

My first scope analysis of this change was wrong. A truncated search missed three consumers
(`U2FAttestationProcessor.isAttestationModeEnabled` and both `AttestationVerifier` sites); they surfaced
only because compilation kept failing. The list has since been verified exhaustively — one global read
remains, the dead one above.

**Behaviour change: none.** An RP with no policy, an RP whose policy leaves the field unset, and an origin
matching no RP all resolve to the global value, which is what happened before.

-------------------

`RpPolicyServiceTest` covers the resolution rule: per-RP override; no policy; policy present but field
unset; blank values not shadowing the global; an origin matching no RP; one RP's policy not leaking onto a
neighbour; case-insensitive matching; a null origin; a null `Fido2Configuration`; and the malformed shapes
(`null` RP list, empty list, RP with null origins, null entry in the list) resolving to the global value
rather than throwing on the ceremony path.

The wiring itself is pinned too — the existing processor tests now assert
`verify(rpPolicyService).resolveAttestationMode(...)` rather than a global read, so a future change that
bypassed the resolver would fail them.

I checked the tests are not vacuous: disabling the per-RP branch in the resolver fails exactly three of
them and nothing else, and I reverted after confirming.

```
RpPolicyServiceTest          11 tests, 0 failures
jans-fido2 model + server   526 tests, 0 failures
```

Docs: `docs/janssen-server/fido/fido2-server-properties-config.md` gains a "Per-relying-party policy"
section, placed under the `fido2Configuration` heading where `rp` is documented, stating the fallback and
that nothing needs changing on upgrade. Carried in a separate `docs:` commit.

#### Known gap, tracked

Once this merges, an RP policy can be configured — and jans-cli-tui will silently discard it.
`020_fido/main.py:294` rebuilds every relying party as `{'id': ..., 'origins': ...}` on save, so opening the
FIDO2 screen and saving, even without editing, drops `policy` from every RP. That is #15112, filed as a bug
and worth landing close behind this. Nothing can be lost before this PR merges, since the field does not
exist yet.

## Summary by CodeRabbit

- **New Features**
  - Added per-relying-party FIDO2 attestation policies that can override the global attestation setting.
  - Matching is based on the relying party’s origin, with case-insensitive support.
  - Relying parties without an override continue using the global configuration.

- **Documentation**
  - Added configuration guidance, fallback behavior, and JSON examples for per-relying-party policies.

- **Tests**
  - Added coverage for overrides, fallback behavior, origin matching, isolation, and malformed configurations.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/cf27a85b-462c-4508-bf8f-22cffb353b5a?refs=951a9c2eac3ecddca073ca6b499a82b607dfee51&source=prChecksComment) |

yurem (approved):

Issue #14605, opened 2026-07-23
feat(jans-fido2): structured native-app (iOS/Android) origin & RP-ID policy model


imran-ishaq:
**Is your feature request related to a problem? Please describe.**
The FIDO2 RP configuration is browser-only. `Fido2Configuration.requestedParties` is a flat list of `RequestedParty` objects, each just an `id` plus a `List<String> origins` (`model/src/main/java/io/jans/fido2/model/conf/RequestedParty.java`). Native-app origins (Android `android:apk-key-hash:…`, iOS RP-ID-scoped) get mixed into the same flat `origins[]` as web origins, and there is **no structured place** to record the platform-specific facts that native passkeys require: Android package name + SHA-256 signing fingerprint (upload key vs. Play App Signing key), or iOS Team ID + Bundle ID. Origin acceptance today is a plain string-equality check against `RequestedParty.getOrigins()` (`CommonVerifiers.verifyRpDomain`, `CommonVerifiers.java:104-107`) with a registrable-suffix rule in `DomainVerifier.verifyDomain` (`DomainVerifier.java:74-77`). When a native origin isn't configured correctly, the result is a generic `INVALID_ORIGIN` / `INVALID_DOMAIN` failure with no explanation of *why* — one of the most common and hardest-to-diagnose native deployment errors.

**Describe the solution you'd like**
Introduce an explicit, structured RP origin-policy model (as part of the RP refactoring) that separates web, Android, and iOS applications under a single RP ID, e.g.:

```jsonc
{
  "rpId": "example.com",
  "webOrigins": ["https://login.example.com"],
  "androidApps": [
    { "packageName": "com.example.app",
      "sha256CertFingerprints": ["AB:CD:…"],
      "distribution": "play-store" }      // play-store | self-signed
  ],
  "iosApps": [
    { "teamId": "T9A667JL6T", "bundleId": "com.example.app" }
  ]
}
```

- This single source of truth is what the `assetlinks.json` / AASA generation builds on, and what the readiness checker validates deployed files against.
- Add **origin-decision diagnostics** explaining why an origin was accepted or rejected, rather than only a generic error. Keep the public FIDO response envelope unchanged.

**Describe alternatives you've considered**
- *Keep the flat `origins[]` and just document native origin formats* — rejected: no place for fingerprints/Team IDs, and the file-generation and readiness-checker work have no source of truth to generate/validate from.
[...]

PR #15120, opened 2026-09-17
fix(jans-cli-tui): preserve unknown relying-party fields and edit the…

### LINKED ISSUE #15112: fix(jans-cli-tui): preserve the per-RP FIDO2 policy when editing a relying party

imran-ishaq:
### Prepare

-------------------

#### Target issue

closes #15112

Last of the three sub-issues of #14516, after #15115 (model and enforcement) and #15111 (swagger).

#### Implementation Details

**The bug.** `020_fido/main.py` rebuilt every relying party from scratch on save, from the only two fields
the table held:

```python
fido2_static['rp'] = []
for name, domains in self.requested_parties_container.data:
    fido2_static['rp'].append({'id': name, 'origins': domains.splitlines()})
```

Now that `RequestedParty` carries a `policy`, an administrator who opened the FIDO2 screen and saved —
without editing anything — silently discarded the policy of every relying party. The loss was invisible
until attestation behaviour quietly reverted to the global mode.

**The fix, in two parts.**

The relying parties are now kept as loaded, and the edited fields are written back onto them. Anything the
screen does not display survives. This is deliberately more general than the policy: #14516 lists several
more fields, and #14605 adds native-app origins, so a screen that reconstructs relying parties from its
own columns would lose each of them in turn.

Attestation mode is then surfaced: a third column, so an administrator can see at a glance which relying
parties override, and a dropdown in the edit dialog whose empty choice reads `-- inherit global --`. Empty
is not "no attestation" — it means the relying party states no preference and the global value applies,
which is how every RP behaved before per-RP policy existed. Clearing it removes the `policy` object
entirely rather than leaving `{"policy": {}}`, so such an RP serialises exactly as it did before the field
existed.

The dialog's help text comes from `#/components/schemas/RequestedPartyPolicy`, added in #15111. That
lookup is guarded: `get_schema_from_reference` indexes the spec directly, so a server whose spec predates
the schema raises rather than returning nothing, and help text is not worth crashing a dialog that works
today.

**One deliberate limitation.** A renamed relying party has no original to match and is treated as new, so
its policy does not follow the rename. The alternative is matching by row index, which breaks on add and
delete.

-------------------

**No automated test is added, and that box is left unticked deliberately.** The save logic lives inside a
plugin method that cannot be driven headlessly, and jans-cli-tui's existing tests (`test.py`,
`run_test.py`) are widget-level and require the whole application. A permanent guard would mean extracting
the transformation into a pure module-level function, which is a refactor beyond this fix — worth doing,
and I am happy to if a reviewer prefers it.

What I did instead: lifted the save loop **verbatim** into a harness and ran it over the cases that matter.
All nine pass, and the pre-fix loop fails four of them — including the reported symptom:

```
PASS  untouched save keeps the policy
PASS  unknown future field survives
PASS  setting a mode writes the policy
PASS  clearing the mode removes the policy
PASS  clearing one field keeps the rest of the policy
PASS  a renamed RP is treated as new
PASS  a new RP with a mode
PASS  editing origins keeps the policy
PASS  the loaded config is not mutated
```

**Regression review.** I traced every consumer of the row data, because widening the table from two
columns to three touches shared widget contracts:

- `JansVerticalNav` indexes `preferred_size[col]` for every column, so a three-column row with a
  two-entry list would raise — extended to `[30, 30, 20]`.
- `on_delete` passes the row, so `kwargs['selected'][1]` still resolves to origins.
- `on_enter` passes `selected` as an index, which is what `replace_item(item_index: int, ...)` expects.
- "Add Party" calls the dialog with no `passed`, so the default was widened to `['', '', '']`; a
  two-element default would raise on `cur_data[2]`.

**Not verified: appearance.** I cannot run a prompt_toolkit TUI headlessly, so the rendering and
interaction of the new column and dropdown are unverified — in particular how three columns lay out
against `max_width=65` now that `preferred_size` sums to 80. That needs someone to open the screen once.
Flagging it rather than implying otherwise.

No documentation change: nothing under `docs/` describes this screen or its fields (confirmed by search),
and the configuration property itself is already documented from #15115.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring FIDO attestation modes: disabled, monitor, enforced, or inherited from the global setting.
  * Added attestation mode visibility to requested-party rows and headers.

* **Bug Fixes**
  * FIDO policy editing now opens even when policy schemas are unavailable.
  * Edits preserve fields not displayed in the interface.
  * Renamed relying parties are correctly treated as new entries.
  * Empty policies are omitted when saving.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/3030e76d-49d3-4375-9bde-7a533f61a6b7?refs=d5b0197562c00e286b7e7d9487e72a585d704032&source=prChecksComment) |

yurem (approved):

yurem (approved):

yurem (approved):

yuremm (approved):

devrimyatar (approved):

PR #15137, opened 2026-09-21
fix(terraform-provider-jans): stop wiping undeclared config on update

### LINKED ISSUE #15128: terraform-provider-jans: fido2 and smtp configuration resources replace instead of patch, silently wiping undeclared fields
`jans_fido2_configuration` and `jans_smtp_configuration` update with a full PUT
built from Terraform state, while `jans_app_configuration` and
`jans_scim_app_configuration` use JSON Patch. Because config-api replaces the
stored configuration wholesale, any attribute the practitioner does not declare
is written back as a Go zero value and lost.

`provider/resource_fido2_configuration.go`:

```go
func resourceFido2ConfigurationUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
	c := meta.(*jans.Client)
	var fido2Config jans.JansFido2DynConfiguration
	if err := fromSchemaResource(d, &fido2Config); err != nil {
		return diag.FromErr(err)
	}
	if _, err := c.UpdateFido2Configuration(ctx, &fido2Config); err != nil {
		return diag.FromErr(err)
	}
	return resourceFido2ConfigurationRead(ctx, d, meta)
}
```

against `provider/resource_app_configuration.go`:

```go
patches, err := patchFromResourceData(d, &appConfig)
if _, err := c.PatchAppConfiguration(ctx, patches); err != nil {
```

Server side, `jans-config-api` fido2-plugin `Fido2Service.merge` replaces the
whole dynamic conf:

```java
public void merge(AppConfiguration fido2ConfigJson) {
    Conf conf = this.findConf();
    conf.setDynamicConf(fido2ConfigJson);
    mergeConf(conf);
}
```

## Impact

A minimal, reasonable-looking resource silently destroys the configuration:

```hcl
resource "jans_fido2_configuration" "global" {
  super_gluu_enabled = false
}
```

After one apply, the stored fido2 configuration on a working deployment became:

```json
{"issuer": null, "baseEndpoint": null, "cleanServiceInterval": 0,
 "useLocalCache": false, "loggingLevel": null, "metricReporterInterval": 0,
 "fido2Configuration": {"authenticatorCertsFolder": "", "mdsCertsFolder": "",
 "mdsTocsFolder": "", "enabledFidoAlgorithms": null, ...}}
```

FIDO2 enrolment stops working. It recurs on every pipeline run, so it reads as
an intermittent fault rather than a configuration bug.

## Suggested fix

Use `patchFromResourceData` / a patch endpoint for `jans_fido2_configuration`
and `jans_smtp_configuration`, matching the app and SCIM resources. Failing
that, document clearly that these two resources are authoritative and that
every attribute must be declared.

Version: terraform-provider-jans 2.4.0, Janssen 2.4.1, MySQL persistence.

mo-auto:
## Fixes

**#15128** — `jans_fido2_configuration` / `jans_smtp_configuration` built their PUT payload from an empty struct, so undeclared attributes were written back as Go zero values. They now GET the current config and overlay only the attributes present in the raw config, which distinguishes a declared zero value from an absent one. Nested blocks are overlaid the same way, so declaring one attribute of `fido2_configuration` no longer resets the rest of the block. `jans_logging_configuration` had the same bug and gets the same treatment.

**Nightly terraform failures** — `TestCustomUsers` and `TestAccResourceCustomUser_basic` both failed with a 500 on GET. Three separate server-side causes, all fixed:

- The provider sent `user-inum: <client_id>`. That is not a user inum, so `UserResource.validateUserPermission` failed the lookup. Header dropped.
- `RolePermissionMappingService.getUserByInum` treated a missing user as null, but the lookup raises `EntryPersistenceException`. Only a not-found (no cause) now maps to null; a backend failure is rethrown rather than reported as a missing user.
- `UserResource` caught `Exception` around its lookups, which also caught the `NotFoundException` that `checkResourceNotNull` raises, so a GET for a user that does not exist answered 500 instead of 404. `WebApplicationException` is rethrown ahead of the catch-all in the fetch, create, update and patch handlers. Terraform destroy refreshes before deleting and relies on that 404.

`AuthUtil.getJsonNodeKeyValue` also dereferenced `JsonNode.get` without a null check, so a token whose `authorizationDetails` lacks an `inum` threw an NPE.

**Machine clients and the user-mgt role check** — the check added in #15069 reads the caller's role from their token, so a `client_credentials` token cannot satisfy it and the provider is rejected while `userRolePermissionValidationEnabled` is on. New `userRolePermissionExcludedClients` config attribute lists client IDs that skip it:

- Empty by default; no behaviour change until an administrator enrols a client.
- The client ID comes from the introspection response, not a request header, so a client cannot exempt itself.
- Ignored for a token that carries a user, so the case the check was written for stays covered.
- Introspection is still only performed when a check needs it.
- Exposed as `user_role_permission_excluded_clients` on `jans_api_app_configuration`.

**Schema drift** with config-api and fido2:

- `jans_api_app_configuration`: adds `user_role_permission_validation_enabled`, `validate_user_inum_in_introspection_flag`, `fetch_user_role_in_introspection_flag` (from #15069), plus `service_name`, `protection_mode`, `return_client_secret_in_response`, `return_encrypted_client_secret_in_response`, `disable_external_logger_configuration`, `acr_exclusion_list` and the `cedarling_configuration` block.
- `jans_fido2_configuration`: adds the metrics and trusted-proxy settings, abandoned-assertion and MDS download settings, `hints`, `enterprise_attestation`, `attestation_mode`, `allowed_top_origins` and `metadata_servers`.
- `requested_parties` serialised `name`/`domains` while the server reads `id`/`origins`, so declaring the block stored relying parties with a null id and no origins. Per-RP `policy.attestation_mode` is now mappable.

**Dead attributes** — `super_gluu_enabled`, `old_u2f_migration_enabled` and `user_info_lifetime` have no counterpart in the fido2 configuration and were silently discarded. Removed. Configurations that set them must drop those lines.

**Docs** — resource docs updated for every schema change. The provider index gains a User Management Endpoints section with the exact errors returned when the allow list is unset and how to fix them, and states that updates overlay declared attributes onto the stored configuration. The user-mgt-plugin page documents the exemption and its limits.

## Tests

`go test ./provider/...`, `go vet ./...` and `TestProvider_InternalValidate` pass. New: `TestAttrConfigState`, `TestDecoderKeepsUndeclaredFields`, `TestPruneUndeclared`, `TestPruneUndeclaredKeepsDeclaredZeroValue`, `TestResourceFido2Config_RequestedPartyMapping`, `TestRequestedPartiesJSONNames`.

The acceptance suite went from 273 tests / 2 failures to 289 tests / 1 failure on both MySQL and PGSQL.

## Merged with one known failing test

`TestAccResourceCustomUser_basic` still fails on teardown with `did not get correct response code: 500`. That is the `NotFoundException` bug described above, fixed in this PR but not observable in CI: `automation/ci/run_aio_for_tf.sh` pulls `ghcr.io/janssenproject/jans/all-in-one:0.0.0-nightly`, built from `main`, so the job never runs the config-api code from this branch. The AIO logs for the failing run show the path exactly:

```
UserResource.java:161    User search by inum:441c350b-…
AuthUtil.java:753        Logged in userInum:null
UserMgmtService.java:286 Failed to load user entry
UserResource.java:186    Exception while fetching user is -
```

The next nightly image carries the fix and the test should pass on the following run. If it does not, the teardown 500 is a different cause and needs a fresh look.

Also not verified anywhere: the config-api changes have never been compiled — no JDK or Maven was available while writing them. Provider docs were hand-edited to match `tfplugindocs` output because the generator needs to download a Terraform binary; a regeneration on a networked machine should be a no-op.


mo-auto:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/janssenproject/pr-checks/a6704499-ab97-46e9-a1e8-1548498484e7?refs=196911bf9d550b42d4e63316c55b53f518dc4b77&source=prChecksComment) |

moabu:
@coderabbitai approve

moabu (approved):

moabu:
@coderabbitai approve

